PR #7, opened 2023-07-12
Add CMake install target

qdeslandes:
Add CMake install target to install locally bpfilter daemon, libbpfilter, and headers.
    
Add bpfilter.pc to use libbpfilter through pkg-config.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,11 @@ find_program(CLANG_FORMAT_BIN clang-format REQUIRED)
 
 include(GoogleTest)
 
+# Set the default install location to /usr/lib64 on 64-bit systems
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -22,6 +22,11 @@ find_program(CLANG_FORMAT_BIN clang-format REQUIRED)
 
 include(GoogleTest)
 
+# Set the default install location to /usr/lib64 on 64-bit systems
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+    set(CMAKE_INSTALL_LIBDIR lib64)
```
davide125, line 27: This won't work on multiarch distros like Debian. I think there's a cmake macro you can use to pick up the correct lib path

qdeslandes: Thanks for pointing it out! It should be fixed in #8 :)

PR #9, opened 2023-08-08
Add CI workflow

rphibel:


.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+env:
+  COVERAGE_THRESHOLD: 13
```
qdeslandes: Why 13 specifically?

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+  NC: \033[0m
+  RED: \033[0;31m
+
```
qdeslandes: Extra empty line

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
+name: Build and Test
```
qdeslandes: `CI` is better, as it does more than build and test. Same for the filename (`ci.yaml`).

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+run-name: Build and Test 🚀
```
qdeslandes: No need for a specific `run-name`.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+env:
+  COVERAGE_THRESHOLD: 13
+  NC: \033[0m
+  RED: \033[0;31m
+
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-22.04
```
qdeslandes: We must run the CI on Fedora too. Also, assuming default arch is x86, how can we run CI on arm64?

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+    permissions:
+      checks: write
+      pull-requests: write
+    strategy:
+      fail-fast: false
+      matrix:
+        buildtype:
+          - Debug
+          - Release
```
qdeslandes: Use lowercase build types.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Install packages
+        run: |
+          sudo apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get -y install cmake doxygen python3-sphinx libcriterion-dev lcov libbpf-dev
+          sudo wget https://apt.llvm.org/llvm-snapshot.gpg.key
+          sudo apt-key add llvm-snapshot.gpg.key
+          echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-16 main" | sudo tee -a /etc/apt/sources.list
+          echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-16 main" | sudo tee -a /etc/apt/sources.list
+          sudo apt-get update
+          sudo apt-get -y install clang-tidy-16 clang-format-16
```
qdeslandes: Isn't `clang-tidy` available in Ubuntu's repositories?

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Create build directory
+        run: |
+          mkdir build
+          cd build
+      - name: Configure build
+        run: cmake -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }} ${{ github.workspace }}
```
qdeslandes: `cmake -B build -S ${{ github.workspace}} -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }}` should be able to replace both those steps.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Lint
+        if: matrix.buildtype == 'Release'
+        run: make lint
+      - name: Check style
+        if: matrix.buildtype == 'Release'
+        run: make checkstyle
+      - name: Build
+        run: make
+      - name: Test
```
qdeslandes: More specifically those are the unit tests.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+        run: make test
+      - name: Publish Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v2
```
qdeslandes: Why using this action, compared to any other one (e.g. https://github.com/marketplace/actions/test-reporter)?

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+        if: always()
+        with:
+          files: |
+            tests/**/*.xml
```
qdeslandes: The report is always located in `tests/unit/test-report.xml`.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Coverage
+        if: matrix.buildtype == 'Debug'
+        run: |
+          make coverage
+          coverage=$(lcov --summary doc/lcov.out | grep -w lines | awk '{print $2}' | cut -d% -f 1 | cut -d. -f1); \
+          if (( coverage < ${{ env.COVERAGE_THRESHOLD }} ))
+          then
+            echo -e "${{ env.RED }}Coverage($coverage) lower that threshold($COVERAGE_THRESHOLD)${{ env.NC }}"
+            exit 1
+          fi
```
qdeslandes: It's better if the coverage progress is reported directly into the PR, instead of failing the job and forcing someone to dive in the logs.

Is it possible to trigger a coverage success/failure based on the changed lines, instead of the whole project?

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+          sleep 1
+          sudo ./tests/end_to_end/ipt
+          sudo ./tests/end_to_end/raw
+          sudo systemctl stop bpfilter
+          rc=$(systemctl show -PExecMainCode bpfilter)
+          if [[ "$rc" != 0 ]]
+          then
+            echo -e "${{ env.RED }}bpfilter returned with exit code $rc${{ env.NC }}"
+            exit 1
+          fi
```
qdeslandes: This can be removed for now, as there end-to-end tests need some work to be reliable.

CMakeLists.txt
```diff
@@ -19,8 +19,8 @@ pkg_check_modules(CRITERION REQUIRED criterion)
 find_program(SPHINX_BIN sphinx-build REQUIRED)
 find_program(LCOV_BIN lcov REQUIRED)
 find_program(GENHTML_BIN genhtml REQUIRED)
-find_program(CLANG_TIDY_BIN clang-tidy REQUIRED)
-find_program(CLANG_FORMAT_BIN clang-format REQUIRED)
+find_program(CLANG_TIDY_BIN clang-tidy-16 REQUIRED)
+find_program(CLANG_FORMAT_BIN clang-format-16 REQUIRED)
```
qdeslandes: This breaks Fedora 38 build.

qdeslandes (changes requested):
That's a good first step, I've got a few improvement suggestions :)

I was expecting this workflow to be picked-up by Github for this PR, so I need to find out what's wrong. It's might be in the project settings.

davide125:
> I was expecting this workflow to be picked-up by Github for this PR, so I need to find out what's wrong. It's might be in the project settings.

If you're adding a workflow for the first time it won't be run in the PR, you'll need to merge it blind. Future PRs should trigger it though.

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,85 @@ on:
...
     runs-on: ubuntu-22.04
-    container: fedora:38
     strategy:
       fail-fast: false
       matrix:
-        buildtype:
-          - debug
-          - release
+        buildtype: [ 'debug', 'release' ]
+        image: [ 'fedora:38', '' ]
```
qdeslandes: Is `''` referring to Ubuntu? In this case, it would be better to explicitely have `ubuntu` instead of empty quotes. This will also show up in the PR.

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,85 @@ on:
...
+            python3-pip \
+            python3-sphinx
+          pip install furo
       - name: Check out repository code
         uses: actions/checkout@v3
       - name: Configure build
         run: cmake -B $GITHUB_WORKSPACE/build -S $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }}
+      - name: Lint
+        if: matrix.buildtype == 'release'
+        run: make -C $GITHUB_WORKSPACE/build lint
```
qdeslandes: Let's skip linting for now, we have a lot of warnings already, they need to be fixed before we enable it.

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,85 @@ on:
...
+      - name: Check style
+        if: matrix.buildtype == 'release'
+        run: make -C $GITHUB_WORKSPACE/build checkstyle
+      - name: Doc
```
qdeslandes: `Doc` -> `Generate documentation`

qdeslandes: I would generate the documentation at the end of the workflow. Overall, I would go with:
- Build
- Lint
- Check style
- Run unit tests
- Create coverage report (and upload it to Codecov)
- Generate documentation

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,85 @@ on:
...
+        if: matrix.buildtype == 'release'
+        run: make -C $GITHUB_WORKSPACE/build doc
       - name: Build
         run: make -C $GITHUB_WORKSPACE/build
+      - name: Unit tests
```
qdeslandes: `Unit tests` -> `Run unit tests`

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,85 @@ on:
...
+        run: make -C $GITHUB_WORKSPACE/build test
+      - name: Publish Test Results
```
qdeslandes: `Publish Test Results` -> `Publish test results`

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,85 @@ on:
...
+        uses: EnricoMi/publish-unit-test-result-action@v2
+        if: always()
+        with:
+          files: |
+            build/tests/unit/test-report.xml
```
qdeslandes: Not sure what this is for, as no test result is published on the PR. On the other hand, do we want to publish the test results, or just run the unit tests and wait for the workflow to fail if tests fail?

If we publish the results of each unit test (Fedora and Ubuntu, debug and release) run, the PR will quickly become a mess. What do you think?

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,85 @@ on:
...
+      - name: Coverage
+        if: matrix.buildtype == 'Debug' && matrix.image == 'fedora:38'
+        run: |
+          make -C $GITHUB_WORKSPACE/build coverage
+      - name: Upload coverage reports to Codecov
```
qdeslandes: `reports` -> `report`?

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,85 @@ on:
...
+        if: matrix.buildtype == 'Debug' && matrix.image == 'fedora:38'
+        uses: codecov/codecov-action@v3
+        with:
+          files: |
+            build/doc/lcov.out
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
qdeslandes, line 77: I'm working on enabling Codecov support for the repo, so we wouldn't need a secret.

.clang-format
```diff
@@ -43,7 +43,6 @@ BraceWrapping:
   SplitEmptyNamespace: false
 BreakBeforeBinaryOperators: None
 BreakBeforeBraces: Custom
-BreakBeforeConceptDeclarations: Always
```
qdeslandes, line 46: There already are unsupported `clang-format` option in this file which are commented out (e.g. `InsertTrailingCommas`), to maintain consistency, either the options incompatible with Ubuntu should be commented out, or all the incompatible options should be deleted. Do as you want, I've no opinion, but it should be consistent.

qdeslandes (changes requested):

.github/workflows/ci.yaml
rphibel: Yes, it's for Ubuntu.

`image` is used to specify the container image. For Ubuntu, we don't need a container, we can run the CI directly on the Ubuntu VM. That's why we have `''` for the image.

.github/workflows/ci.yaml
rphibel: Just running the tests and check if they fail or not is probably enough. I thought it would be nice to have the list of test that failed quickly visible in the PR. But I agree that with the different build configurations, it will be a mess.

.github/workflows/ci.yaml
davide125: This definitely needs a clarifying comment then

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,76 @@ on:
...
         run: make -C $GITHUB_WORKSPACE/build
+      - name: Check style
+        if: matrix.buildtype == 'release'
+        run: make -C $GITHUB_WORKSPACE/build checkstyle
+      - name: Run unit tests
+        run: make -C $GITHUB_WORKSPACE/build test
+      - name: Coverage
+        if: matrix.buildtype == 'Debug' && matrix.image == 'fedora:38'
+        run: |
+          make -C $GITHUB_WORKSPACE/build coverage
```
qdeslandes: No need for a new line.

.github/workflows/ci.yaml
```diff
@@ -7,27 +7,76 @@ on:
...
+      - name: Upload coverage report to Codecov
+        if: matrix.buildtype == 'Debug' && matrix.image == 'fedora:38'
+        uses: codecov/codecov-action@v3
+        with:
+          files: |
+            build/doc/lcov.out
```
qdeslandes: No need for a new line.

.github/workflows/ci.yaml
qdeslandes: We should use a container for Ubuntu too for consistency purpose, so all the build would be isolated in a container. If my understanding is correct, right now we build for Ubuntu directly on the host, then start a Fedora container to build on Fedora. Also, for now the check summary on the PR is empty when running on Ubuntu. 

Screenshot 2023-09-05 at 11 33 29

rphibel: I didn't use a container for Ubuntu because I thought that we could not run the integrated tests inside them.

qdeslandes: Integration tests has not been run inside a container yet, so it might require more development to make them work, but it should be possible. Eventually, we'll be able to run integration tests in a container.

qdeslandes:
Looks all good to me! Thanks!

Screenshot 2023-09-06 at 12 27 23

PR #10, opened 2023-08-11
Add basic CI

rphibel:
This change adds a basic CI. This is meant to be a starting point, that's why we only build on Fedora.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,33 @@
+name: CI
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    container: fedora:38
+    strategy:
+      fail-fast: false
+      matrix:
+        buildtype:
+          - debug
+          - release
+    steps:
+      - name: Install packages
+        run: |
+          sudo dnf -y install 'dnf-command(copr)'
+          sudo dnf -y copr enable naccyde/criterion 
+          sudo dnf -y install cmake libbpf-devel \
+                              criterion-devel clang-tools-extra lcov libasan libubsan \
+                              doxygen python3-sphinx python3-breathe python3-furo
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Configure build
+        run: cmake -B $GITHUB_WORKSPACE/build -S $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }}
+      - name: Build
+        run: make -C $GITHUB_WORKSPACE/build


.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,38 @@
+name: Build and Test
```
qdeslandes: CI is better, as it does more than build and test. Same for the filename (ci.yaml).

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+run-name: Build and Test 🚀
```
qdeslandes: No need for `run-name`.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build-and-test:
```
qdeslandes: As this job only builds `bpfilter`, you can rename it to `build`.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-22.04
+    container: fedora:38
+    strategy:
+      fail-fast: false
+      matrix:
+        buildtype:
+          - Debug
+          - Release
```
qdeslandes: `bpfilter`'s `CMakeLists.txt` uses lowercase build types, you can do it here too.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+          sudo dnf -y copr enable naccyde/criterion 
+          sudo dnf -y install cmake libbpf-devel \
+                              criterion-devel clang-tools-extra lcov libasan libubsan \
+                              doxygen python3-sphinx python3-breathe python3-furo
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Create build directory
+        run: |
+          mkdir build
+          cd build
```
qdeslandes: Not needed, as CMake will do it.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Configure build
+        run: cmake -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }} $GITHUB_WORKSPACE
```
qdeslandes: `cmake -B build -S $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }}` will create a build directory (`build`), and run CMake with sources from `$GITHUB_WORKSPACE`.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Build
+        run: make -j
```
qdeslandes: Following previous comments: if CMake creates the build directory, then the current directory is not `build`, in which case we'll call `make` as `make -C $GITHUB_WORKSPACE`.

Let's also avoid using multiple threads (remove `-j`): I can't find if it's well supported by GitHub Runners, and the build failures will be clearer (build is fast, so it's fine not to build in parallel).

qdeslandes (changes requested):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,33 @@
...
+          sudo dnf -y copr enable naccyde/criterion 
+          sudo dnf -y install cmake libbpf-devel \
+                              criterion-devel clang-tools-extra lcov libasan libubsan \
+                              doxygen python3-sphinx python3-breathe python3-furo
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Configure build
+        run: cmake -B build -S $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }}
+      - name: Build
+        run: make -C $GITHUB_WORKSPACE/build
```
qdeslandes: Could you always use the same path for the build folder? (i.e. `cmake -B build` vs `make -C $GITHUB_WORKSPACE/build`). `build` is relative to the current directory, while `$GITHUB_WORKSPACE/build` is absolute. Those are the same path if run now, but in case that changes in the future, we'll be safer to use either `$GITHUB_WORKSPACE/build` or `build` for now.

qdeslandes (changes requested):

PR #17, opened 2023-09-19
Add support for Netfilter BPF programs

qdeslandes:
Add a new `NF` flavor to support generation of `BPF_PROG_TYPE_NETFILTER` programs using `bpfilter`. 

Until now, `bpfilter` would generate TC programs for `iptables` (IPT front). Meaning it had to perform a routing table lookup on TC ingress to check whether the `LOCAL_IN` or `FORWARD` filtering rules should be applied to the packet.

This PR add support for `BPF_PROG_TYPE_NETFILTER` programs, giving access to `LOCAL_IN`, `FORWARD, and `LOCAL_OUT` hooks to attach programs to.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -67,6 +67,8 @@ IncludeCategories:
     Priority: 1
   - Regex: '^<linux/'
     Priority: 2
+  - Regex: '^"external/.*'
+    Priority: 5
   - Regex: '^"'
     Priority: 4
   - Regex: '.*'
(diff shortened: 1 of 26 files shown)


qdeslandes:
While this PR coverage isn't high enough (7.94% instead of 10.70%), most of the unit tests would depend on mocking syscalls. Until a solution to do so it available, there is nothing much we can do.

PR #20, opened 2023-10-03
Add macos runner

rphibel:


.github/actions/ci-action/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM fedora:38
+RUN sudo dnf -y install 'dnf-command(copr)'
+RUN sudo dnf -y copr enable naccyde/criterion
```
davide125, line 3: Shouldn't need these anymore now that https://github.com/facebook/bpfilter/commit/eb168151bbfca2219e7a0397771c7542caaeabbf is in

PR #24, opened 2023-10-24
Various CMake fixes

DaanDeMeyer:


CMakeLists.txt
```diff
@@ -12,7 +12,15 @@ project(bpfilter
 include(GNUInstallDirs)
 
 find_package(Doxygen REQUIRED)
-find_package(cmocka REQUIRED)
+
+find_package(PkgConfig)
+pkg_check_modules(bpf REQUIRED IMPORTED_TARGET libbpf)
+pkg_check_modules(elf REQUIRED IMPORTED_TARGET libelf)
+pkg_check_modules(cmocka REQUIRED IMPORTED_TARGET cmocka)
```
qdeslandes:
```
find_package(cmocka REQUIRED) # We can keep this, and link `tests_unit` to `cmocka::cmocka`
find_package(PkgConfig REQUIRED)

pkg_check_modules(bpf REQUIRED IMPORTED_TARGET libbpf)
pkg_check_modules(elf REQUIRED IMPORTED_TARGET libelf)
```

.github/workflows/ci.yaml
```diff
@@ -36,7 +36,8 @@ jobs:
             libubsan \
             python3-breathe \
             python3-furo \
-            python3-sphinx
+            python3-sphinx \
+            pkgconf
```
qdeslandes, line 40: Add `pkgconf` to the requirements in `README.md` too.

CMakeLists.txt
```diff
@@ -21,7 +29,7 @@ find_program(CLANG_TIDY_BIN clang-tidy REQUIRED)
 find_program(CLANG_FORMAT_BIN clang-format REQUIRED)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-set(CMAKE_C_STANDARD 17)
+set(CMAKE_C_STANDARD 11)
```
qdeslandes: This can be removed in favour of adding `-std=gnu17` to `bpfilter_cflags` later in the file.

tests/unit/CMakeLists.txt
```diff
@@ -114,10 +114,12 @@ bf_test_configure_non_build_srcs(tests_unit
         ${bpfilter_shared_srcs}
 )
 
+# cmocka needs PIE disabled to be able to auto-discover tests.
```
qdeslandes: It's not exactly CMocka, but the test harness performing autodiscovery (https://github.com/facebook/bpfilter/blob/main/tests/unit/harness/elf.c).

CMakeLists.txt
```diff
@@ -12,7 +12,15 @@ project(bpfilter
...
-find_package(cmocka REQUIRED)
+
+find_package(PkgConfig)
+pkg_check_modules(bpf REQUIRED IMPORTED_TARGET libbpf)
+pkg_check_modules(elf REQUIRED IMPORTED_TARGET libelf)
+pkg_check_modules(cmocka REQUIRED IMPORTED_TARGET cmocka)
+
+# Required to get CMake to pass PIE flags to the compiler/linker
+include(CheckPIESupported)
+check_pie_supported()
```
qdeslandes, line 23: `POSITION_INDEPENDENT_CODE` property is not supported if this function is not called? Does it fails the generation if PIE is not supported?

qdeslandes (changes requested):

qdeslandes:
Please prefix your commits with `build` instead of `cmake`.

CMakeLists.txt
daandemeyer: Yes it only works if this is called. It does not fail if it is not supported, pie options will simply not be used in the emited compiler command lines

CMakeLists.txt
daandemeyer: @qdeslandes The `cmocka` CMake package script does not set up a target so we would have to use `${CMOCKA_LIBRARIES}` everywhere. I prefer using pkgconfig which does give us a target, but it's your call.

qdeslandes: It doesn't on yours? That's from `cmocka-config.cmake` on Fedora:
```
# Create imported target cmocka::cmocka
add_library(cmocka::cmocka SHARED IMPORTED)
```

CMakeLists.txt
qdeslandes: In this case, you should probably add a warning if `CMAKE_C_LINK_PIE_SUPPORTED` is false.

CMakeLists.txt
qdeslandes: After checking on Ubuntu, it doesn't. Ubuntu has `libcmocka-dev-1.1.5` vs `cmocka-devel-1.1.7` on Fedora. Let's go with your way then!

PR #59, opened 2024-04-25
Add custom target to build nftables and iptables from bpfilter's CMake

qdeslandes:
Use CMake's `ExternalProject_Add()` command to clone, patch, build, and install (within the build directory) `nftables` and `iptables` with `bpfilter` support. Update the documentation to explain how to use these targets.


tests/integration/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+# SPDX-License-Identifier: GPL-2.0
```
davide125: `GPL-2.0` is ambiguous -- you want either `GPL-2.0-or-later` or `GPL-2.0-only`.

tests/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@
 
 add_subdirectory(unit)
 add_subdirectory(end_to_end)
+add_subdirectory(integration)
```
davide125, line 6: Consider gating this behind a cmake option, so that it's easy to disable when packaging. Generally when packaging the build system doesn't have access to the Internet, so `ExternalProject` wouldn't work out of the box.

tests/integration/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+
+    PATCH_COMMAND git reset --hard ${BF_TOOLS_NFT_GIT_TAG}
+        COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/nftables/0001-libnftables-add-support-for-bpfilter.patch
+    CONFIGURE_COMMAND ./autogen.sh
+        COMMAND
+            ${CMAKE_COMMAND} -E env
+                LDFLAGS=-Wl,-rpath=${CMAKE_BINARY_DIR}/lib\ $<$<CONFIG:Debug>:-fsanitize=address\ -fsanitize=undefined>
+                CFLAGS=-I${CMAKE_SOURCE_DIR}/lib/include\ -I${CMAKE_SOURCE_DIR}/shared/include\ $<$<CONFIG:Debug>:-fsanitize=address\ -fsanitize=undefined>
+                BPFILTER_LIBS=-L${CMAKE_BINARY_DIR}/lib\ -lbpfilter
+                BPFILTER_CFLAGS=""
```
davide125: Is this clobbering the flags or just appending to them?

README.md
```diff
@@ -17,22 +19,30 @@ sudo dnf install \
...
 # Build bpfilter
-cmake -S $BPFILTER_SOURCES -B $BPFILTER_BUILD
-make -C $BPFILTER_BUILD
-make -C $BPFILTER_BUILD test
+cmake -S $SOURCES_DIR -B $BUILD_DIR
+make -C $BUILD_DIR
+make -C $BUILD_DIR test
+
+# Build a custom version of nftables and iptables to use with bpfilter
+make -C $BUILD_DIR nftables iptables
```
davide125, line 35: This is fine, but FYI one can also use `cmake -Gninja` (and environments do that by default); I have no idea if it works with this codebase, but it might be worth documenting either way.

tests/CMakeLists.txt
qdeslandes: The `ExternalProject` target are not part of the default build, meaning their will be no internet access unless `make nftables` or `make iptables` is called. I tried a local only build and there was no issue.

tests/integration/CMakeLists.txt
qdeslandes: Appending. I've modified the nftables patch to properly propagate `BPFILTER_*` flags, so `LDFLAGS` and `CFLAGS` only define `-fsanitize=*` when needed.

README.md
qdeslandes: I've added a one line fix to allow Ninja to be used, and updated the documenation. I'll kept the README.md as short as possible as this is now documented in the full instructions.

tests/integration/CMakeLists.txt
qdeslandes: This is a leftover from `bpfilter`'s kernel era. According to [this documentation](https://www.kernel.org/doc/html/v6.4/process/license-rules.html) the proper tag should then be `GPL-2.0-only`. I'll update this one in this PR and put up another PR to change it on the whole project.

PR #109, opened 2024-09-07
First part of sets support

qdeslandes:
Sets will allow fast and efficient filtering on a set of pre-defined data. This will be relevant to filter in/out a large pool of different IP addresses, for example. This PR introduces the new `bf_set` object and its interaction with `bf_chain`, as well as the `IP4_ADDR_SET` matcher, to filter on a set of IPv4 addresses.

A few other changes are added, related to the source tree.


qdeslandes:
Codecov support has been disabled for some time, so the coverage requirements are slightly off-base here. Let's merge this one to restore a valid target.

PR #146, opened 2024-10-21
Add CONTINUE verdict

ryantimwilson:
This PR adds a CONTINUE verdict for rules s.t. a rule can match and continue processing. This allows multiple rules to match and count a packet.

Note CONTINUE is not supported for chains as there is no nesting of chains and one cannot jump between BPF programs.

### DIFF
--- a/doc/usage/bfcli.rst
+++ b/doc/usage/bfcli.rst
@@ -56,7 +56,7 @@ With:
     - ``BF_HOOK_NF_POST_ROUTING``: similar to ``nftables`` and ``iptables`` postrouting hook.
     - ``BF_HOOK_TC_EGRESS``: egress TC hook.
 
-  - ``$POLICY``: action taken if no rule matches the packet, either ``ACCEPT`` forward the packet to the kernel, or ``DROP`` to discard it.
+  - ``$POLICY``: action taken if no rule matches the packet, either ``ACCEPT`` forward the packet to the kernel, or ``DROP`` to discard it. Note while ``CONTINUE`` is a valid verdict for rules, it is not supported for chain policy.
 
 ``$OPTIONS`` are hook-specific comma separated key value pairs:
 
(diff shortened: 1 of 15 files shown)


doc/usage/bfcli.rst
```diff
@@ -101,9 +101,14 @@ Rules are defined such as:
 With:
   - ``$MATCHER``: zero or more matchers. Matchers are defined later.
   - ``counter``: optional literal. If set, the filter will counter the number of packets and bytes matched by the rule.
-  - ``$VERDICT``: action taken by the rule if the packet is matched against **all** the criteria: either ``ACCEPT`` or ``DROP``.
+  - ``$VERDICT``: action taken by the rule if the packet is matched against **all** the criteria: either ``ACCEPT``, ``DROP`` or ``CONTINUE``.
+    - ``ACCEPT``: Forward the packet to the kernel
+    - ``DROP``: Discard the packet.
+    - ``CONTINUE``: Continue processing subsequent rules.
```
qdeslandes: Drop the capital letter after the colon for consistency.

doc/usage/bfcli.rst
```diff
@@ -101,9 +101,14 @@ Rules are defined such as:
...
 
-In a chain, as soon as a rule matches a packet, its verdict is applied, and the subsequent rules are not processed. Hence, the rules' order matters. If no rule matches the packet, the chain's policy is applied.
+In a chain, as soon as a rule matches a packet, its verdict is applied. If the verdict is ``ACCEPT`` or ``DROP``, the subsequent rules are not processed. Hence, the rules' order matters. If no rule matches the packet, the chain's policy is applied.
+
+Note ``CONTINUE`` means a packet can be counted more than once if multiple matchers specify ``CONTINUE`` and ``counter``.
```
qdeslandes: "if multiple matchers" -> "if multiple rules"

For reference, a rule contains 1+ matchers. The rule defines the verdict if all the matchers matches.

doc/usage/bfcli.rst
qdeslandes: It would be worth updating the "Chains" section of the doc to mention that even though `CONTINUE` exists, it's not valid as a chain policy.

src/bfcli/parser.y
```diff
@@ -133,7 +135,7 @@ chains          : chain
                 }
                 ;
 
-chain           : CHAIN hook raw_hook_opts POLICY verdict rules
+chain           : CHAIN hook raw_hook_opts POLICY policy rules
```
qdeslandes: Keep the `verdict` rule here, and if `policy == CONTINUE` then trigger `bf_parse_err()`.  This way we reuse the existing code, and we provide a contextual error (e.g. "CONTINUE not supported for chains").

src/core/verdict.h
```diff
@@ -7,16 +7,23 @@
...
     /** Drop the packet. */
     BF_VERDICT_DROP,
+    /** Non-terminal verdicts that allow further packet processing. */
+    /** Continue processing the next rule. */
+    BF_VERDICT_CONTINUE,
     _BF_VERDICT_MAX,
 };
 
+#define _BF_TERMINAL_VERDICT_MAX BF_VERDICT_CONTINUE
+
```
qdeslandes: Define it in the enum instead:
```c
...
    _BF_TERMINAL_VERDICT_MAX = BF_VERDICT_DROP,
...
```

src/core/chain.c
```diff
@@ -27,6 +27,8 @@ int bf_chain_new(struct bf_chain **chain, enum bf_hook hook,
 
     bf_assert(chain);
```
qdeslandes: No line break.

qdeslandes (changes requested):
A few nits, but overall, I like it :)

qdeslandes (approved):

qdeslandes:
LGTM, thanks for doing this!

PR #154, opened 2024-10-30
cgen: add an error counter

### LINKED ISSUE #105: Collect error metrics
Generated BPF programs call `kfunc`s and BPF helpers for various purposes, and those call might fail. In this situation, the program will return with a default verdict. There is currently no way to know if such an error occurred during the program's lifetime, except for reading `/sys/kernel/debug/tracing/trace_pipe` if `bpfilter` is running in debug mode (`--debug`).

An error/failure counter would bring more visibility to this situation and help to understand the program's behavior.

amscanne:
For kfunc call errors, add a dedicated counter that can be read by the daemon in the future. This counter is exposed implicitly as a named counter object ("errors") via `nft`.

Fixes #105

### DIFF
--- a/src/bpfilter/cgen/cgen.c
+++ b/src/bpfilter/cgen/cgen.c
@@ -185,15 +185,22 @@ void bf_cgen_dump(const struct bf_cgen *cgen, prefix_t *prefix)
     bf_dump_prefix_pop(prefix);
 }
 
-int bf_cgen_get_counter(const struct bf_cgen *cgen, uint32_t counter_idx,
+int bf_cgen_get_counter(const struct bf_cgen *cgen,
+                        enum bf_counter_type counter_idx,
                         struct bf_counter *counter)
 {
     bf_assert(cgen && counter);
 
-    /* There are 1 more counter than number of rules. The last counter is
-     * dedicated to the policy. */
-    if (counter_idx > bf_list_size(&cgen->chain->rules))
+    /* There are two more counter than rules. The special counters must
+     * be accessed via the specific values, to avoid confusion. */
+    enum bf_counter_type rule_count = bf_list_size(&cgen->chain->rules);
+    if (counter_idx == BF_COUNTER_POLICY) {
+        counter_idx = rule_count;
+    } else if (counter_idx == BF_COUNTER_ERRORS) {
+        counter_idx = rule_count + 1;
+    } else if (counter_idx < 0 || counter_idx >= rule_count) {
         return -EINVAL;
+    }
 
     return bf_program_get_counter(cgen->program, counter_idx, counter);
 }
--- a/src/bpfilter/cgen/cgen.h
+++ b/src/bpfilter/cgen/cgen.h
@@ -128,11 +128,21 @@ int bf_cgen_unload(struct bf_cgen *cgen);
 void bf_cgen_dump(const struct bf_cgen *cgen, prefix_t *prefix);
 
 /**
+ * @enum bf_counter_type
+ *
+ * Special counter types for @ref bf_cgen_get_counter .
+ */
+enum bf_counter_type
+{
+    BF_COUNTER_POLICY = -1,
+    BF_COUNTER_ERRORS = -2,
+};
+
+/**
  * Get packets and bytes counter at a specific index.
  *
- * Counters are referenced by their index in the counters map. There are 1 more
- * counter in the map than the number of rules. This last counter (the last in
- * the map) is dedicated to the policy.
+ * Counters are referenced by their index in the counters map or the enum
+ * values defined by @ref bf_counter_type .
  *
  * The counter from all the program generated from @p cgen are summarised
  * together.
@@ -142,7 +152,8 @@ void bf_cgen_dump(const struct bf_cgen *cgen, prefix_t *prefix);
  *        correspond to a valid index, -E2BIG is returned.
  * @param counter Counter structure to fill with the counter values. Can't be
  *        NULL.
- * @return 0 on success, or a negative errno  value on failure.
+ * @return 0 on success, or a negative errno value on failure.
  */
-int bf_cgen_get_counter(const struct bf_cgen *cgen, uint32_t counter_idx,
+int bf_cgen_get_counter(const struct bf_cgen *cgen,
+                        enum bf_counter_type counter_idx,
                         struct bf_counter *counter);
--- a/src/bpfilter/cgen/program.c
+++ b/src/bpfilter/cgen/program.c
@@ -891,6 +891,11 @@ int bf_program_generate(struct bf_program *program)
             bf_flavor_to_str(bf_hook_to_flavor(program->hook)),
             bf_front_to_str(program->front), bf_hook_to_str(program->hook));
 
+    /* Add 1 to the number of counters for the policy counter, and 1
+     * for the first reserved error slot. This must be done ahead of
+     * generation, as we will index into the error counters. */
+    program->num_counters = bf_list_size(&chain->rules) + 2;
+
     r = _bf_program_generate_runtime_init(program);
     if (r)
         return r;
@@ -935,9 +940,6 @@ int bf_program_generate(struct bf_program *program)
     r = _bf_program_fixup(program, BF_FIXUP_TYPE_FUNC_CALL);
     if (r)
         return bf_err_r(r, "failed to generate function call fixups");
-
-    // Add 1 to the number of counters for the policy counter.
-    program->num_counters = bf_list_size(&chain->rules) + 1;
 
     return 0;
 }
--- a/src/bpfilter/cgen/stub.c
+++ b/src/bpfilter/cgen/stub.c
@@ -19,6 +19,7 @@
 #include <endian.h>
 #include <stddef.h>
 
+#include "bpfilter/cgen/fixup.h"
 #include "bpfilter/cgen/jmp.h"
 #include "bpfilter/cgen/printer.h"
 #include "bpfilter/cgen/program.h"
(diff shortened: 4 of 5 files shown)


src/bpfilter/cgen/cgen.h
```diff
@@ -146,3 +149,29 @@ void bf_cgen_dump(const struct bf_cgen *cgen, prefix_t *prefix);
  */
 int bf_cgen_get_counter(const struct bf_cgen *cgen, uint32_t counter_idx,
                         struct bf_counter *counter);
+
+/**
+ * Get packets and bytes counter for the policy.
+ *
+ * See `bf_cgen_get_counter` for an explanation.
```
qdeslandes: Use `@ref` instead of backticks to reference functions defined in `bpfilter`.

src/bpfilter/cgen/program.c
```diff
@@ -936,8 +936,9 @@ int bf_program_generate(struct bf_program *program)
     if (r)
         return bf_err_r(r, "failed to generate function call fixups");
 
-    // Add 1 to the number of counters for the policy counter.
-    program->num_counters = bf_list_size(&chain->rules) + 1;
+    // Add 1 to the number of counters for the policy counter, and 1
+    // for the first reserved error slot.
```
qdeslandes: `/* */`

src/bpfilter/cgen/cgen.h
```diff
@@ -146,3 +149,29 @@ void bf_cgen_dump(const struct bf_cgen *cgen, prefix_t *prefix);
...
+ *
+ * @param cgen Codegen to get the counter for. Can't be NULL.
+ * @param counter Counter structure to fill with the counter values. Can't be
+ *        NULL.
+ * @return 0 on success, or a negative errno  value on failure.
```
qdeslandes: Two spaces between `errno` and `value`.

src/bpfilter/cgen/cgen.c
```diff
@@ -190,14 +190,37 @@ int bf_cgen_get_counter(const struct bf_cgen *cgen, uint32_t counter_idx,
...
+                              struct bf_counter *counter)
+{
+    bf_assert(cgen && counter);
+
+    // The error slot is always the last.
+    return bf_program_get_counter(cgen->program,
+                                  bf_list_size(&cgen->chain->rules)+1,
+                                  counter);
+}
+
```
qdeslandes: I think it would be easier to manage to do something like:
```
enum bf_counter_type
{
    BF_COUNTER_POLICY = -1,
    BF_COUNTER_ERRORS = -2,
};
```

Then call `bf_cgen_get_counter(bf_cgen *cgen, enum bf_counter_type counter_idx, struct bf_counter *counter)` with:
- `bf_cgen_get_counter(cgen, BF_COUNTER_POLICY, &counter)`
- `bf_cgen_get_counter(cgen, BF_COUNTER_ERRORS, &counter)`
- `bf_cgen_get_counter(cgen, rule_idx, &counter)`

`bf_cgen_get_counter()` would then request the proper index from `bf_program_get_counter()`.

qdeslandes (changes requested):
Thanks for doing this! 

Regarding `nft` support for counters, it's a bit of a pain at the moment as it diverged a lot from what `bfcli` can do. Also, if you send back the error counter, `nftables` will have no idea what to do with it, as that doesn't exist in `nft` world. 

To check if your counter is updated, the easiest solution would be to use `bpftool map dump id XXX` to dump the counters map (`bf_cmap_xxx`).

`bfcli` can't yet fetch the counters, but @ryantimwilson is working on it. In the meantime, I would suggest to go the `bpftool` way, as neither `iptables` nor `nftables` has any concept of error counter.

src/bpfilter/cgen/cgen.h
amscanne: Done

src/bpfilter/cgen/cgen.h
amscanne: Done

src/bpfilter/cgen/cgen.c
amscanne: There are some other consequences for this, and I don't *love* abusing the enum as a index also, but I don't have a strong preference. PTAL, and see if this makes sense with the ipt changes to support these semantics (since it will need to change rather than relying on the size of the rules being used to access the policy).

amscanne:
> Regarding nft support for counters, it's a bit of a pain at the moment as it diverged a lot from what bfcli can do. Also, if 
> you send back the error counter, nftables will have no idea what to do with it, as that doesn't exist in nft world.

I've gone this route and dropped the NFT code (since I think it opens a can of worms there). Do you know of a good way to force failures on these paths and test? Everything seems to be the way I expect it, but I'd like to test with real faults on these paths.

src/bpfilter/cgen/program.c
amscanne: Done

src/bpfilter/cgen/cgen.h
```diff
@@ -127,12 +127,22 @@ int bf_cgen_unload(struct bf_cgen *cgen);
 
 void bf_cgen_dump(const struct bf_cgen *cgen, prefix_t *prefix);
 
+/**
+ * @enum bf_counter_type
+ *
+ * Special counter types for @ref bf_cgen_get_counter.
```
qdeslandes: `@ref bf_cgen_get_counter.` -> `@ref bf_cgen_get_counter .` otherwise Doxygen assumes the symbol name is `bf_cgen_get_counter.`

src/bpfilter/cgen/cgen.h
```diff
@@ -127,12 +127,22 @@ int bf_cgen_unload(struct bf_cgen *cgen);
...
+};
+
 /**
  * Get packets and bytes counter at a specific index.
  *
- * Counters are referenced by their index in the counters map. There are 1 more
- * counter in the map than the number of rules. This last counter (the last in
- * the map) is dedicated to the policy.
+ * Counters are referenced by their index in the counters map or the enum
+ * values defined by @ref bf_counter_type.
```
qdeslandes: Same

src/bpfilter/cgen/cgen.c
qdeslandes: It's not ideal, I agree, but I feel the intent is clearer. For `iptables`, the semantics used by `bpfilter` are different, so the bridging between `iptables` and `bpfilter` is expected (same for `nftables`).

qdeslandes (changes requested):

qdeslandes:
> I've gone this route and dropped the NFT code (since I think it opens a can of worms there). Do you know of a good way to force failures on these paths and test? Everything seems to be the way I expect it, but I'd like to test with real faults on these paths.

Unfortunately, no, as we would need to trigger an error on a `kfunc`/BPF helper. A well crafted BPF program could probably hook to one of the `kfunc` and filter the calls (failing those coming from `bpfilter`).

You could change the conditional used to check for error, so the counters would be updated on success, that would be better than nothing. 

Furthermore, could you run `make fixstyle` to apply the same style as the whole codebase?

src/bpfilter/cgen/cgen.c
amscanne: I've done it in this way.

src/bpfilter/cgen/cgen.h
amscanne: Done

src/bpfilter/cgen/cgen.h
amscanne: Done.

amscanne:
> > I've gone this route and dropped the NFT code (since I think it opens a can of worms there). Do you know of a good way to force failures on these paths and test? Everything seems to be the way I expect it, but I'd like to test with real faults on these paths.
> 
> Unfortunately, no, as we would need to trigger an error on a `kfunc`/BPF helper. A well crafted BPF program could probably hook to one of the `kfunc` and filter the calls (failing those coming from `bpfilter`).
> 
> You could change the conditional used to check for error, so the counters would be updated on success, that would be better than nothing.
> 
> Furthermore, could you run `make fixstyle` to apply the same style as the whole codebase?

I've fixed the style as suggested. I ran some manual tests with the same counter in the valid path, and it seemed fine.

I don't want to complicate this specific pull request, but thinking out loud: did adding a dynamic layer of functions for fault injection? (E.g. dynamically add a function indirection where the builtins are all bounced through a trampoline, which could probabalistically or deterministically inject various faults.) Is such a feature generally useful? Maybe alongside being able to control the fail open/closed behavior for each chain? (With the fault injection basically giving you the ability to test that.)

qdeslandes:
> I don't want to complicate this specific pull request, but thinking out loud: did adding a dynamic layer of functions for fault injection? (E.g. dynamically add a function indirection where the builtins are all bounced through a trampoline, which could probabalistically or deterministically inject various faults.) Is such a feature generally useful? Maybe alongside being able to control the fail open/closed behavior for each chain? (With the fault injection basically giving you the ability to test that.)

That's a great idea, and it would definitely be useful for more than testing `kfunc` error handling (e.g. validate user-provided BPF bytecode). Please don't hesitate to give it a go if you're interested. I think most of the pieces are already available (map support, custom function generation).

qdeslandes:
LGTM. The CI is mad because of `clang-tidy`, if you rebase on `main` + add the following to `cgen.c` it will fix it:
```
diff --git a/src/bpfilter/cgen/stub.c b/src/bpfilter/cgen/stub.c
index 25bb6a3..9be6ce0 100644
--- a/src/bpfilter/cgen/stub.c
+++ b/src/bpfilter/cgen/stub.c
@@ -19,6 +19,7 @@
 #include <endian.h>
 #include <stddef.h>
 
+#include "bpfilter/cgen/fixup.h"
 #include "bpfilter/cgen/jmp.h"
 #include "bpfilter/cgen/printer.h"
 #include "bpfilter/cgen/program.h"
 ```

qdeslandes:
Done! Thanks @amscanne for doing this!

Issue #151, opened 2024-10-28
Allow fetching rules and counters via bfcli


ryantimwilson:
bpfilter daemon keeps a list of active rules and it also writes counters to BPF maps. Sure, we can access the rules/counters via bpftool - counter maps are named bf_cmap_XXX - but it is more convenient to use an API.

```
$ sudo bpftool map dump name bf_cmap_000201
[{
        "key": 0,
        "value": {
            "packets": 71220,
            "bytes": 112981553
        }
    },{
        "key": 1,
        "value": {
            "packets": 71288,
            "bytes": 112993982
        }
    }
```

To do this, we need to implement the BF_REQ_GET_RULES and BF_REQ_GET_RULES APIs in bpfilter backend and fetch the results via bfcli.

bfcli API could look like (where chain/rule are optional):
```
$ bfcli get-rules <chain> <rule>
$ bfcli get-counters <chain> <rule>
```

Note `bfcli` would need to introduce subcommands as `bfcli` is only used for setting rules now.

qdeslandes:
Some thoughts on this:
- `BF_REQ_GET_RULES` API is as old as this repository, the daemon changed a lot since then, you *might* have to modify that logic a bit.
- `bfcli get-rules` would be a problem on its own: there is no logic to convert the ruleset back to the format used by `bfcli`. Maybe it would be worth to piggy-back on `bf_xxx_dump()` logic? You can print almost anything with `bf_xxx_dump()`, except for a matcher's payload (but that's easily solvable).
- `bfcli get-counters`: `nft` prints the counters with the rule's definition, that could be a way to do it.

ryantimwilson:
Tommy Unger is working on adding get-counters functionality as part of his ramp-up!

tommy-u:
I have a few noob questions here. I'm currently working on `get-counters` because we agreed that `get-rules` will require further conversation.

1. Could it be that we actually want `<chain>` and `<rule>` to be required arguments to `get-counters`? If `<chain>` and `<rule>` are optional, what is the expected functionality when one, the other, or both are omitted? Perhaps omitting both shows counters from all chains and all rules. Omitting `<rule>` shows all counters from all rules in the given `<chain>`. Omitting `<chain>` shows all counters corresponding to matching rules across all chains (is this useful?).
[...]

PR #176, opened 2024-12-13
bpfilter: add useful information to 'error:' log in the event of a misused '--verbose` flag

ryanbsull:


src/core/opts.c
```diff
@@ -140,8 +140,10 @@ static error_t _bf_opts_parser(int key, char *arg, struct argp_state *state)
         break;
     case 'v':
         r = bf_verbose_to_str(arg, &opt);
-        if (r < 0)
+        if (r < 0) {
+            bf_info("valid --verbose options: [debug, bpf, bytecode]");
             return bf_err_r(EINVAL, "unknown --verbose option '%s'", arg);
```
qdeslandes: List the valid options from the error message directly, i.e. "unknown --verbose option '%s'. Valid options are ...". Otherwise the user might miss it and jump directly the the error message.

qdeslandes (changes requested):
Thanks for your contribution @ryanbsull this is a great addition. There's a minor change to do but otherwise that's fine by me!

src/core/opts.c
ryanbsull: Hi, changed it up, lmk if that's all good, also was looking for a quick tutorial in the docs (a "Quick Start" example??). I watched your talk here: [bpfilter: a BPF-based packet filtering framework](https://www.youtube.com/watch?v=fzaPEm4PXn0) and took some notes that I feel could be useful for the docs, but lmk

qdeslandes:
Merged! Thanks @ryanbsull !

src/core/opts.c
qdeslandes: A Quick Start section would definitely be valuable in the doc. Something like starting the daemon, using `bfcli` to define one or more filtering rules, and retrieve the counters (counters are fetched through `bpftool` for now, see #151). Let me know what you would like to see in the doc (feel free to create a new issue for this) and I'm happy for you to do it if you want to! :)

PR #219, opened 2025-02-19
Better BPF objects pinning

qdeslandes:
All the BPF objects generated by `bpfilter` are pinned into `/sys/fs/bpf`. This behaviour is error-prone as some objects don't have unique filenames (e.g. cgroup filters). This PR aims to fix this issue by:
- Generating a predictable and unique ID for each `bf_program`, based on the hook and the hook options.
- If the BPF objects must be pinned, `bf_program` will do so in `/sys/fs/bpf/bpfilter/$ID`: isolate `bpfilter`'s BPF objects into a subdirectory, and pin all the objects of a given program into a folder named after its ID. Removing the directory will unpin all the object for a given program (program included).

To do so, all the BPF objects are pinned relative to a directory (`/sys/fs/bpf/bpfilter/$ID`), this design is more scalable than the previous implementation.

A few fixes have been added:
- Remove unused `bf_bpf__map_create()`, `union bf_flavor_attach_attr`, `attach_prog_pre_unload()` callback, `attach_prog_post_unload()` callback
- Allow `bf_map` to be created without key size or value size, as long as it's defined before the map is created
- Use `asroot` script to run end-to-end tests: this script is already used for `make benchmark`, it runs a given command with `sudo` if EUID is not `0`
- Fix the regex used to filter out `tests/unit` files from the `make check` target so it doesn't rely on the build directory name


qdeslandes:
I'll merge with the CI failure, the error is due to `cp` not finding the source file to copy, because the file is not created unless the CI runs on `main`. Subsequent run of the CI from a PR will be able to use the source file from the run on `main`.

PR #233, opened 2025-03-17
modify commands to get a working build from README

adelavegaf:
Following the instructions in the README & in the docs verbatim to build from source isn't working for me (both in ubuntu & fedora).

It's likely that I'm doing things wrong since it's my first time with cmake. But in case it's helpful for others, this commit adapts the README with the changes I had to make to the commands to get the daemon up & running.


qdeslandes:
LGTM, thanks @adelavegaf !

PR #252, opened 2025-05-05
Use argp global variables to generate version flag in bpfilter and bfcli

SkohTV:
[argp](https://www.gnu.org/software/libc/manual/html_node/Argp.html) has a feature to autogenerate the `--version` flag using [global variables](https://www.gnu.org/software/libc/manual/html_node/Argp-Global-Variables.html)
This PR aims to replace the manual implementation in `bpfilter` and `bfcli` binairies with this feature

### DIFF
--- a/src/bfcli/main.c
+++ b/src/bfcli/main.c
@@ -27,10 +27,20 @@
 #include "libbpfilter/bpfilter.h"
 #include "version.h"
 
+static void _bfc_print_version(FILE *stream, struct argp_state *state)
+{
+    UNUSED(state);
+
+    (void)fprintf(stream, "bfcli version %s, libbpfilter version %s\n",
+                  BF_VERSION, bf_version());
+}
+
 int main(int argc, char *argv[])
 {
     _clean_bfc_opts_ struct bfc_opts opts = bfc_opts_default();
     int r;
+
+    argp_program_version_hook = &_bfc_print_version;
 
     r = bfc_opts_parse(&opts, argc, argv);
     if (r < 0)
--- a/src/bpfilter/opts.c
+++ b/src/bpfilter/opts.c
@@ -16,7 +16,6 @@
 #include "core/front.h"
 #include "core/helper.h"
 #include "core/logger.h"
-#include "version.h"
 
 #define BF_DEFAULT_BPFFS_PATH "/sys/fs/bpf"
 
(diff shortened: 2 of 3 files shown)


SkohTV:
Should be all good
Do we also want a test to assure that this flag won't be broken in future versions ?

src/bfcli/main.c
```diff
@@ -30,6 +30,7 @@
 int main(int argc, char *argv[])
 {
     _clean_bfc_opts_ struct bfc_opts opts = bfc_opts_default();
+    argp_program_version_hook = &bf_print_version;
```
qdeslandes: This is not a variable definition, it should be lower, e.g.:

```
int r;

argp_program_version_hook = ...;

r = bfc_opts_parser(...);
```

src/bfcli/opts.c
```diff
@@ -187,6 +189,14 @@ static const struct bfc_opts_cmd _bfc_opts_cmds[] = {
...
 };
 
+void bf_print_version(FILE *stream, struct argp_state *state)
+{
+    UNUSED(state);
+
+    (void)fprintf(stream, "bfcli version %s, libbpfilter version %s\n",
+                  BF_VERSION, bf_version());
+}
+
```
qdeslandes: Move this back to `bfcli/main.c` as `static void _bfc_print_version...)`: it's not related to the options per se, only a helper function.

src/bpfilter/opts.c
```diff
@@ -33,6 +32,8 @@ static const char *_bf_verbose_strs[] = {
     [BF_VERBOSE_BYTECODE] = "bytecode",
 };
 
+const char *argp_program_version = "bpfilter version " BF_VERSION;
```
qdeslandes: In `bfcli/main.c`, the callback is set in the main function, this should be done similarly:

```
int main()
{
   ...

   bf_logger_setup();

   argp_program_version = "bpfilter version " BF_VERSION;

   ...
}
```

qdeslandes (changes requested):

qdeslandes:
> Should be all good Do we also want a test to assure that this flag won't be broken in future versions ?

Not necessarily, because only two things can be tested here:
- `--version` exists: it always exists as it's provided by argp
- The returned version is the right one: parsing and checking the printed version would require more lines of code than the actual feature, it's not worth it

qdeslandes:
Also please change your commits subject for consistency:

- `Use argp_program_version to autogenerate version flag for bpfilter` -> `daemon: use argp_program_version to autogenerate version flag for bpfilter`
- `Use argp_program_version_hook to autogenerate version flag for bfcli` -> `cli: use argp_program_version_hook to autogenerate version flag for bfcli`

qdeslandes (approved):

qdeslandes:
@SkohTV LGTM, thanks for your contribution!

PR #277, opened 2025-05-22
Dynamically compute `YY_READ_BUF_SIZE` to reduce time spent parsing large chains

### LINKED ISSUE #257: Performance issue when loading large ruleset with bfcli
It appears that loading a large ruleset with bfcli can be (very) slow. For example, loading a simple chain (see below) with a set containing 1M IPv4 takes XX seconds (Fedora 42 on M3 Pro laptop). 

```
chain my_chain BF_HOOK_XDP ACCEPT
    rule
        ip4.saddr in {1.1.1.1,...,100.100.100.1)
        DROP
```

A quick profiling of bfcli reveals `yyparse` is the issue.

There are a couple of ways to reproduce and investigate this issue:
- `make benchmark` will run `loadChainLargeSet` (merged in #256) and print the run time of two ruleset using a 10000 IPv4 set and a 100000 IPv4 set.
- Generate a chain with a set containing many IPv4 using [this script](https://gist.github.com/qdeslandes/e735fe3b8a507734892ba3ef10dc8ae2), then use `tools/perfrec --target firefox bfcli ruleset set --from-file ./ruleset.bpfilter`. Open the output file (`perf.data`) with [profiler.firefox.com](https://profiler.firefox.com/).

SkohTV:
Fix #257 

We noticed that loading large rulesets in bfcli took a long time, and after tracking it down, the solution to increase the `YY_READ_BUF_SIZE` was found.
Since we can't modify it directly, we use a compiler definition in `src/bfcli/CMakeLists.txt` to make it refer to a global variable that we can manipulate easily.

The current implementation of `compute_buf_size()` is extrapolated through tries and the [graph](https://github.com/facebook/bpfilter/issues/257#issuecomment-2891493844), it's implementation is heuristic and could be improved in the future.
Also we *could* add a lower/upper bound on the buffer size, but I don't know how useful that is.

The `build/bfcli/generated/bfcli/lexer.c` mention that "`YY_BUF_SIZE is 2*YY_READ_BUF_SIZE in the general case`".
But the code it's used in doesn't seems to care if that's not a case, and I didn't face any performance change when increasing this buffer, so I didn't touched it.

Not sure if this was the best approach, but seems pretty good nonetheless.

### DIFF
--- a/src/bfcli/helper.c
+++ b/src/bfcli/helper.c
@@ -8,15 +8,41 @@
 
 #include <errno.h>
 #include <stdio.h>
+#include <sys/stat.h>
 
 #include "bfcli/lexer.h"
 #include "bfcli/parser.h"
 #include "core/logger.h"
 
+// To speed up parsing very large rulesets, we can increase YY_READ_BUF_SIZE
+int yy_read_buf_size;
+
+#define _BF_LEX_MIN_BUF_POW 14
+#define _BF_LEX_MAX_BUF_POW 20
+
+static int _bf_compute_lexer_buf_size(size_t len)
+{
+    int val = _BF_LEX_MIN_BUF_POW;
+
+    if (len) {
+        val = (int)sizeof(len) * 8 - __builtin_clzl(len);
+        val = bf_min(bf_max(val - 3, _BF_LEX_MIN_BUF_POW), _BF_LEX_MAX_BUF_POW);
+    }
+
+    return 1 << val;
+}
+
 int bfc_parse_file(const char *file, struct bfc_ruleset *ruleset)
 {
     FILE *rules;
+    struct stat stt;
     int r;
+
+    if (stat(file, &stt) == -1)
+        return bf_err_r(errno, "failed to stat file %s:", file);
+
+    yy_read_buf_size = _bf_compute_lexer_buf_size(stt.st_size);
+    bf_dbg("yy_read_buf_size choosen is %d", yy_read_buf_size);
 
     rules = fopen(file, "r");
     if (!rules)
@@ -36,7 +62,12 @@ int bfc_parse_file(const char *file, struct bfc_ruleset *ruleset)
 int bfc_parse_str(const char *str, struct bfc_ruleset *ruleset)
 {
     YY_BUFFER_STATE buffer;
+    unsigned long str_size;
     int r;
+
+    str_size = strlen(str);
+    yy_read_buf_size = _bf_compute_lexer_buf_size(str_size);
+    bf_dbg("yy_read_buf_size choosen is %d", yy_read_buf_size);
 
     buffer = yy_scan_string(str);
 
--- a/tools/benchmarks/main.cpp
+++ b/tools/benchmarks/main.cpp
@@ -30,6 +30,8 @@ void loadChainLargeSet(::benchmark::State &state)
 BENCHMARK(loadChainLargeSet)
     ->Arg(10000)
     ->Arg(100000)
+    ->Arg(200000)
+    ->Arg(500000)
     ->Iterations(10)
     ->Unit(benchmark::kMillisecond);
 
--- a/src/bfcli/CMakeLists.txt
+++ b/src/bfcli/CMakeLists.txt
@@ -65,6 +65,7 @@ add_executable(bfcli
 target_compile_definitions(bfcli
     PRIVATE
         BF_CONTACT="${BF_CONTACT}"
+        "YY_READ_BUF_SIZE=(yy_read_buf_size)"
 )
 
 target_include_directories(bfcli
--- a/src/bfcli/lexer.l
+++ b/src/bfcli/lexer.l
@@ -11,6 +11,8 @@
     #include "core/verdict.h"
     #include "core/hook.h"
     #include "core/matcher.h"
+
+    extern int yy_read_buf_size;
 %}
 
 %option noyywrap
--- a/src/core/helper.h
+++ b/src/core/helper.h
@@ -194,6 +194,13 @@ extern const char *strerrordesc_np(int errnum);
         _a < _b ? _a : _b;                                                     \
     })
 
+#define bf_max(a, b)                                                           \
+    ({                                                                         \
+        __typeof__(a) _a = (a);                                                \
+        __typeof__(b) _b = (b);                                                \
+        _a > _b ? _a : _b;                                                     \
+    })
+
 /**
  * Free a pointer and set it to NULL.
  *


src/bfcli/helper.c
```diff
@@ -8,16 +8,38 @@
 
 #include <errno.h>
 #include <stdio.h>
+#include <sys/stat.h>
 
 #include "bfcli/lexer.h"
 #include "bfcli/parser.h"
 #include "core/logger.h"
 
+int yy_read_buf_size;
```
qdeslandes: Add a comment to describe what it's used for

src/bfcli/helper.c
```diff
@@ -8,16 +8,38 @@
...
+
+int compute_buf_size(unsigned long chain_length)
```
qdeslandes: `static int _bf_compute_lexer_buf_size(size_t)`

tools/benchmarks/main.cpp
qdeslandes, line 1: This change can be merged into the other commit.

src/bfcli/helper.c
```diff
@@ -8,16 +8,38 @@
...
+
+    return (2 << i);
+}
+
 int bfc_parse_file(const char *file, struct bfc_ruleset *ruleset)
 {
     FILE *rules;
     int r;
 
+    struct stat stt;
```
qdeslandes: Define variables at the beginning of the function.

src/bfcli/helper.c
```diff
@@ -38,6 +60,9 @@ int bfc_parse_str(const char *str, struct bfc_ruleset *ruleset)
     YY_BUFFER_STATE buffer;
     int r;
 
+    unsigned long str_size = strlen(str);
```
qdeslandes: Define variables at the beginning of the function.

src/bfcli/helper.c
```diff
@@ -8,16 +8,38 @@
...
+    return (2 << i);
+}
+
 int bfc_parse_file(const char *file, struct bfc_ruleset *ruleset)
 {
     FILE *rules;
     int r;
 
+    struct stat stt;
+    stat(file, &stt);
```
qdeslandes: Check `stat()` return value, it can fail.

src/bfcli/helper.c
```diff
@@ -8,16 +8,38 @@
...
+    yy_read_buf_size = compute_buf_size(stt.st_size);
```
qdeslandes: Log the buffer size.

src/bfcli/helper.c
```diff
@@ -38,6 +60,9 @@ int bfc_parse_str(const char *str, struct bfc_ruleset *ruleset)
     YY_BUFFER_STATE buffer;
     int r;
 
+    unsigned long str_size = strlen(str);
+    yy_read_buf_size = compute_buf_size(str_size);
```
qdeslandes: Log the buffer size.

src/bfcli/helper.c
```diff
@@ -8,16 +8,38 @@
...
+    // buf_size = 2 ^ (round(log2(chain_length / 11)))
+
+    chain_length >>= 4;
+
+    int i = 0;
+    for (; chain_length > 1; i++) {
+        chain_length >>= 1;
+    }
+
+    return (2 << i);
```
qdeslandes: The computation here is a bit messy, what about something like:

```
#define _BF_LEX_MIN_BUF_POW 14
#define _BF_LEX_MAX_BUF_POW 20

int compute_buf_size(unsigned long len)
{
    int v = _BF_LEX_MIN_BUF_POW;

    if (len) {
        v = sizeof(len) * 8 - __builtin_clzl(len);
        v = bf_min(bf_max(v - 3, _BF_LEX_MIN_BUF_POW), _BF_LEX_MAX_BUF_POW);
    }

    return 1 << v;
}
```

By default, the size of the buffer should be 16 KiB (`1 << 14` bytes). `__builtin_clzl()` is supported by both Clang and [GCC](https://gcc.gnu.org/onlinedocs/gcc/Bit-Operation-Builtins.html), but the result is undefined if `len` is 0, so we check for this.

If `len` is not zero, `sizeof(len) * 8 - __builtin_clzl(len)` is the position of the MSB set in `len`. Then we scale `v` down a bit to avoid creating large buffers too quickly (you did it with `chain_length >>= 4`) 

Your comment mentioned dividing `len` by 11 (not sure why 11 specifically, though), so we shift `v` by 3 (divide by 8) which is the closest, while ensuring the buffer is at least 16 KiB and at most 1 MiB.

With the following in `core/helper.h`:
```
#define bf_max(a, b)                                                           \
({                                                                             \
    __typeof__(a) _a = (a);                                                    \
    __typeof__(b) _b = (b);                                                    \
    _a > _b ? _a : _b;                                                         \
})
```

This function will create the following buffers:
```
--------------------------------------------------------------------------------
Benchmark                                      Time             CPU   Iterations
--------------------------------------------------------------------------------
Chain length: 1100742, buffer size: 131072
loadChainLargeSet/100000/iterations:1       60.7 ms         1.49 ms            1

Chain length: 2223656, buffer size: 262144
loadChainLargeSet/200000/iterations:1        122 ms         2.63 ms            1

Chain length: 3330352, buffer size: 262144
loadChainLargeSet/300000/iterations:1        208 ms         3.85 ms            1

Chain length: 4449838, buffer size: 524288
loadChainLargeSet/400000/iterations:1        246 ms         3.90 ms            1

Chain length: 5559920, buffer size: 524288
loadChainLargeSet/500000/iterations:1        319 ms         6.25 ms            1

Chain length: 6676056, buffer size: 524288
loadChainLargeSet/600000/iterations:1        413 ms         4.79 ms            1

Chain length: 7834106, buffer size: 524288
loadChainLargeSet/700000/iterations:1        497 ms         6.05 ms            1

Chain length: 9046914, buffer size: 1048576
loadChainLargeSet/800000/iterations:1        497 ms         7.06 ms            1

Chain length: 10263712, buffer size: 1048576
loadChainLargeSet/900000/iterations:1        586 ms         12.0 ms            1

Chain length: 84, buffer size: 16384
firstRuleDropCounter                        14.9 ns         14.9 ns     47000000 nInsn=131

Chain length: 311, buffer size: 16384
dropAfterXRules/8                           20.9 ns         20.9 ns     33000000 nInsn=253

Chain length: 1126, buffer size: 16384
dropAfterXRules/32                          42.0 ns         41.9 ns     17000000 nInsn=637

Chain length: 4419, buffer size: 16384
dropAfterXRules/128                          166 ns          165 ns      5000000 nInsn=2.173k

Chain length: 17859, buffer size: 16384
dropAfterXRules/512                          820 ns          818 ns      1000000 nInsn=8.317k

Chain length: 72668, buffer size: 16384
dropAfterXRules/2048                       10327 ns        10287 ns      1000000 nInsn=32.893k
```

I think this function scales nicely! Here are the results of the benchmarks suite (200000 and 500000 versions have been added in this PR, so there is no comparison point):

Screenshot 2025-05-26 at 21 01 11

qdeslandes (changes requested):

src/bfcli/helper.h
```diff
@@ -8,6 +8,16 @@
 
 #include "core/list.h"
 
+#define _BF_LEX_MIN_BUF_POW 14
+#define _BF_LEX_MAX_BUF_POW 20
```
qdeslandes: Those should be defined above `_bf_compute_lexer_buf_size()`, we don't have any use for it outside `helper.c`. Also, the `_` prefix indicates the symbols are local to the current translation unit.

src/bfcli/helper.h
```diff
@@ -8,6 +8,16 @@
...
+
+#define bf_max(a, b)                                                           \
+    ({                                                                         \
+        __typeof__(a) _a = (a);                                                \
+        __typeof__(b) _b = (b);                                                \
+        _a > _b ? _a : _b;                                                     \
+    })
```
qdeslandes: `bf_max()` can be moved to `core/helper.h` instead, as `bf_min()` is already defined there.

qdeslandes:
> Should be all good I used `bf_dbg()` to log the buffer size, if the message or log level needs an update, let me know

The log is fine. 

I've added a couple of nits, other LGTM.

SkohTV:
Nits should be resolved

qdeslandes:
Alright, this PR is now merged! 

Thanks for diving into this, this change introduces a huge performance boost for a very low cost. Also, good job on the clever `extern` usage to configure `YY_READ_BUF_SIZE`!

Issue #214, opened 2025-02-12
Support ICMPv6 packet filtering


qdeslandes:
Allow users to filter on ICMP (v6) type and code:

```
rule
    icmpv6.type xxx
    icmpv6.code xxx
    ACCEPT
```

See [iptables](https://ipset.netfilter.org/iptables-extensions.man.html#lbBD) and [nftables](https://wiki.nftables.org/wiki-nftables/index.php/Quick_reference-nftables_in_10_minutes#Icmpv6) implementation.

qdeslandes:
Done by @vinxcls in #284

PR #284, opened 2025-06-06
cli,daemon: add support for icmpv6.type and icmpv6.code

vinxcls:
Issue #214 

bfcli and bpfilter now support ICMPv6 type and code.
e2e tests have been added. cli tests were excluded as they currently appear not to support IPv6.
Minimal update of ruleset file.


qdeslandes (approved):
That's good for me, thanks!

Issue #216, opened 2025-02-12
Support IPv6 next header filtering


qdeslandes:
Filter packets based on the "next header" content of the IPv6 header:

```
rule
    ipv6.nexthdr eq frag
    DROP
rule
    ipv6.nexthdr in {route, prot}
    counter
    DROP
```

qdeslandes:
Done by @vinxcls

PR #286, opened 2025-06-09
cli,daemon,doc: Add support for ip6.nexthdr

vinxcls:
Issue #216

I have done initial work to support extra headers in IPv6. To fully parse the data inside these headers, a second kernel query is required (which I have not implemented yet). The extra header is valid only if the L3 protocol is IPv6.

Then I have added support for IPv6 next header filtering, included e2e tests and updated the documentation accordingly.

Please let me know if everything looks good or if there are any suggestions or nits.

Thank you!

### DIFF
--- a/src/bfcli/parser.y
+++ b/src/bfcli/parser.y
@@ -80,6 +80,7 @@
 %token <sval> MATCHER_IP4_NET
 %token <sval> MATCHER_IP6_ADDR
 %token <sval> MATCHER_IP6_NET
+%token <sval> MATCHER_IP6_NEXTHDR
 %token <sval> MATCHER_PORT MATCHER_PORT_RANGE
 %token <sval> MATCHER_ICMP
 %token <sval> STRING
@@ -605,6 +606,21 @@ matcher         : matcher_type matcher_op MATCHER_META_IFINDEX
 
                     if (bf_matcher_new(&matcher, $1, $2, &set_id, sizeof(set_id)))
                         bf_parse_err("failed to create a new matcher\n");
+
+                    $$ = TAKE_PTR(matcher);
+                }
+                | matcher_type matcher_op MATCHER_IP6_NEXTHDR
+                {
+                    _free_bf_matcher_ struct bf_matcher *matcher = NULL;
+                    enum bf_matcher_ipv6_nh nexthdr;
+
+                    if (bf_matcher_ipv6_nh_from_str($3, &nexthdr))
+                        bf_parse_err("Unknown IPv6 next-header '%s', ignoring\n", $3);
+
+                    if (bf_matcher_new(&matcher, $1, $2, &nexthdr, sizeof(nexthdr)) < 0)
+                        bf_parse_err("failed to create a new matcher\n");
+
+                    free($3);
 
                     $$ = TAKE_PTR(matcher);
                 }
--- a/src/bpfilter/cgen/matcher/ip6.c
+++ b/src/bpfilter/cgen/matcher/ip6.c
@@ -27,6 +27,13 @@
     (((uint32_t)(a) << 24) | ((uint32_t)(b) << 16) | ((uint32_t)(c) << 8) |    \
      (uint32_t)(d))
 #define _BF_MASK_LAST_BYTE 15
+#define BF_IPV6_EH_HOPOPTS(x) ((x) << 0)
+#define BF_IPV6_EH_ROUTING(x) ((x) << 1)
+#define BF_IPV6_EH_FRAGMENT(x) ((x) << 2)
+#define BF_IPV6_EH_AH(x) ((x) << 3)
+#define BF_IPV6_EH_DSTOPTS(x) ((x) << 4)
+#define BF_IPV6_EH_MH(x) ((x) << 5)
+
 
 static int _bf_matcher_generate_ip6_addr(struct bf_program *program,
                                          const struct bf_matcher *matcher)
@@ -205,6 +212,47 @@ static int _bf_matcher_generate_ip6_net(struct bf_program *program,
     return 0;
 }
 
+static int _bf_matcher_generate_ip6_nexthdr(struct bf_program *program,
+                                            const struct bf_matcher *matcher)
+{
+    const uint8_t ehdr = matcher->payload[0];
+
+    if ((matcher->op != BF_MATCHER_EQ) && (matcher->op != BF_MATCHER_NE))
+        return -EINVAL;
+
+    switch (ehdr) {
+    case IPPROTO_HOPOPTS:
+    case IPPROTO_ROUTING:
+    case IPPROTO_DSTOPTS:
+    case IPPROTO_FRAGMENT:
+    case IPPROTO_AH:
+    case IPPROTO_MH:
+        uint8_t eh_mask = (BF_IPV6_EH_HOPOPTS(ehdr == IPPROTO_HOPOPTS) |
+                           BF_IPV6_EH_ROUTING(ehdr == IPPROTO_ROUTING) |
+                           BF_IPV6_EH_FRAGMENT(ehdr == IPPROTO_FRAGMENT) |
+                           BF_IPV6_EH_AH(ehdr == IPPROTO_AH) |
+                           BF_IPV6_EH_DSTOPTS(ehdr == IPPROTO_DSTOPTS) |
+                           BF_IPV6_EH_MH(ehdr == IPPROTO_MH));
+        EMIT(program, BPF_LDX_MEM(BPF_DW, BPF_REG_1, BPF_REG_10,
+                                  BF_PROG_CTX_OFF(ipv6_eh)));
+        EMIT(program, BPF_ALU64_IMM(BPF_AND, BPF_REG_1, eh_mask));
+        EMIT_FIXUP_JMP_NEXT_RULE(
+            program,
+            BPF_JMP_IMM((matcher->op == BF_MATCHER_EQ) ? BPF_JEQ : BPF_JNE,
+                        BPF_REG_1, 0, 0));
+        break;
+    default:
+        /* check l4 protocols using BPF_REG_8 */
+        EMIT_FIXUP_JMP_NEXT_RULE(
+            program,
+            BPF_JMP_IMM((matcher->op == BF_MATCHER_EQ) ? BPF_JNE : BPF_JEQ,
+                        BPF_REG_8, ehdr, 0));
+        break;
+    }
+
+    return 0;
+}
+
 int bf_matcher_generate_ip6(struct bf_program *program,
                             const struct bf_matcher *matcher)
 {
@@ -224,6 +272,9 @@ int bf_matcher_generate_ip6(struct bf_program *program,
     case BF_MATCHER_IP6_SNET:
     case BF_MATCHER_IP6_DNET:
         r = _bf_matcher_generate_ip6_net(program, matcher);
+        break;
+    case BF_MATCHER_IP6_NEXTHDR:
+        r = _bf_matcher_generate_ip6_nexthdr(program, matcher);
         break;
     default:
         return bf_err_r(-EINVAL, "unknown matcher type %d", matcher->type);
--- a/src/bpfilter/cgen/stub.c
+++ b/src/bpfilter/cgen/stub.c
@@ -270,7 +270,9 @@ int bf_stub_parse_l3_hdr(struct bf_program *program)
         // Process EH
         EMIT(program, BPF_MOV64_REG(BPF_REG_1, BPF_REG_10));
         EMIT(program, BPF_ALU64_IMM(BPF_ADD, BPF_REG_1, BF_PROG_CTX_OFF(arg)));
-        EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_PARSE_IPV6_EH);
+        EMIT_FIXUP_ELFSTUB(program, program->ipv6_nexthdr ?
+                                    BF_ELFSTUB_PARSE_IPV6_NH :
+                                    BF_ELFSTUB_PARSE_IPV6_EH);
         EMIT(program, BPF_MOV64_REG(BPF_REG_8, BPF_REG_0));
 
         ehjmp = bf_jmpctx_get(program, BPF_JMP_A(0));
--- a/src/core/matcher.c
+++ b/src/core/matcher.c
@@ -147,6 +147,7 @@ static const char *_bf_matcher_type_strs[] = {
     [BF_MATCHER_IP6_SNET] = "ip6.snet",
     [BF_MATCHER_IP6_DADDR] = "ip6.daddr",
     [BF_MATCHER_IP6_DNET] = "ip6.dnet",
+    [BF_MATCHER_IP6_NEXTHDR] = "ip6.nexthdr",
     [BF_MATCHER_TCP_SPORT] = "tcp.sport",
     [BF_MATCHER_TCP_DPORT] = "tcp.dport",
     [BF_MATCHER_TCP_FLAGS] = "tcp.flags",
@@ -222,6 +223,8 @@ static const char *_bf_matcher_tcp_flags_strs[] = {
     [BF_MATCHER_TCP_FLAG_ECE] = "ECE", [BF_MATCHER_TCP_FLAG_CWR] = "CWR",
 };
 
+static_assert(ARRAY_SIZE(_bf_matcher_tcp_flags_strs) == _BF_MATCHER_TCP_FLAG_MAX);
+
 const char *bf_matcher_tcp_flag_to_str(enum bf_matcher_tcp_flag flag)
 {
     bf_assert(0 <= flag && flag < _BF_MATCHER_TCP_FLAG_MAX);
@@ -238,6 +241,45 @@ int bf_matcher_tcp_flag_from_str(const char *str,
     for (size_t i = 0; i < _BF_MATCHER_TCP_FLAG_MAX; ++i) {
         if (bf_streq(_bf_matcher_tcp_flags_strs[i], str)) {
             *flag = i;
+            return 0;
+        }
+    }
+
+    return -EINVAL;
+}
+
+static const char *_bf_matcher_ipv6_nh_strs[] = {
+    [BF_IPV6_NH_HOP] = "hop",
+    [BF_IPV6_NH_TCP] = "tcp",
+    [BF_IPV6_NH_UDP] = "udp",
+    [BF_IPV6_NH_ROUTING] = "route",
+    [BF_IPV6_NH_FRAGMENT] = "frag",
+    [BF_IPV6_NH_AH] = "ah",
+    [BF_IPV6_NH_ICMPV6] = "icmpv6",
+    [BF_IPV6_NH_DSTOPTS] = "dst",
+    [BF_IPV6_NH_MH] = "mh",
+};
+
+static_assert(ARRAY_SIZE(_bf_matcher_ipv6_nh_strs) == _BF_MATCHER_IPV6_NH_MAX);
+
+const char *bf_matcher_ipv6_nh_to_str(enum bf_matcher_ipv6_nh nexthdr)
+{
+    bf_assert(0 <= nexthdr && nexthdr < _BF_MATCHER_IPV6_NH_MAX);
+
+    return _bf_matcher_ipv6_nh_strs[nexthdr];
+}
+
+int bf_matcher_ipv6_nh_from_str(const char *str,
+                                enum bf_matcher_ipv6_nh *nexthdr)
+{
+    bf_assert(str && nexthdr);
+
+    for (size_t i = 0; i < _BF_MATCHER_IPV6_NH_MAX; ++i) {
+        /* sparse array */
+        if (!_bf_matcher_ipv6_nh_strs[i])
+            continue;
+        if (bf_streq(_bf_matcher_ipv6_nh_strs[i], str)) {
+            *nexthdr = i;
             return 0;
         }
     }
--- a/src/core/matcher.h
+++ b/src/core/matcher.h
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include <linux/in.h>
+#include <linux/in6.h>
 #include <stddef.h>
 #include <stdint.h>
 
@@ -73,6 +75,8 @@ enum bf_matcher_type
     BF_MATCHER_IP6_DADDR,
     /// Matches IPv6 destination network.
     BF_MATCHER_IP6_DNET,
+    /// Matches IPv6 next header
+    BF_MATCHER_IP6_NEXTHDR,
     /// Matches against the TCP source port
     BF_MATCHER_TCP_SPORT,
     /// Matches against the TCP destination port
@@ -117,6 +121,23 @@ struct bf_matcher_ip6_addr
     uint8_t addr[16];
     /// 128-bits IPv6 mask.
     uint8_t mask[16];
+};
+
+/**
+ * Define the IPv6 next headers.
+ */
+enum bf_matcher_ipv6_nh
+{
+    BF_IPV6_NH_HOP = IPPROTO_HOPOPTS,
+    BF_IPV6_NH_TCP = IPPROTO_TCP,
+    BF_IPV6_NH_UDP = IPPROTO_UDP,
+    BF_IPV6_NH_ROUTING = IPPROTO_ROUTING,
+    BF_IPV6_NH_FRAGMENT = IPPROTO_FRAGMENT,
+    BF_IPV6_NH_AH = IPPROTO_AH,
+    BF_IPV6_NH_ICMPV6 = IPPROTO_ICMPV6,
+    BF_IPV6_NH_DSTOPTS = IPPROTO_DSTOPTS,
+    BF_IPV6_NH_MH = IPPROTO_MH,
+    _BF_MATCHER_IPV6_NH_MAX,
 };
 
 /**
@@ -279,3 +300,21 @@ const char *bf_matcher_tcp_flag_to_str(enum bf_matcher_tcp_flag flag);
  */
 int bf_matcher_tcp_flag_from_str(const char *str,
                                  enum bf_matcher_tcp_flag *flag);
+
+/**
+ * Convert a IPv6 next-header to a string.
+ *
+ * @param nexthdr IPv6 next-header to convert.
+ * @return String representation of the IPv6 next-header.
+ */
+const char *bf_matcher_ipv6_nh_to_str(enum bf_matcher_ipv6_nh nexthdr);
+
+/**
+ * Convert a string to the corresponding IPv6 next-header.
+ *
+ * @param str String containing the name of the IPv6 next-header.
+ * @param nexthdr IPv6 next-header value, if the parsing succeeds.
+ * @return 0 on success, or negative errno value on failure.
+ */
+int bf_matcher_ipv6_nh_from_str(const char *str,
+                                enum bf_matcher_ipv6_nh *nexthdr);
--- a/tests/e2e/cli.sh
+++ b/tests/e2e/cli.sh
@@ -422,12 +422,23 @@ suite_icmpv6_chain_set() {
     log "[SUITE] icmpv6: chain set"
     expect_success "can parse icmpv6 type and code by TC chain" \
 	    ${FROM_NS} ${BFCLI} ruleset set --from-str \"chain xdp BF_HOOK_TC_INGRESS\{ifindex=${NS_IFINDEX}\} ACCEPT rule icmpv6.type eq 128 icmpv6.code eq 0 counter DROP\"
-    expect_success "can parse icmpv6 type and code by TC chain" \
-	    ${FROM_NS} ${BFCLI} ruleset set --from-str \"chain xdp BF_HOOK_XDP\{ifindex=${NS_IFINDEX}\} ACCEPT rule icmp.type eq 8 icmp.code eq 0 counter DROP\"
+    expect_success "can parse icmpv6 type and code by XDP chain" \
+	    ${FROM_NS} ${BFCLI} ruleset set --from-str \"chain xdp BF_HOOK_XDP\{ifindex=${NS_IFINDEX}\} ACCEPT rule icmpv6.type eq 8 icmp.code eq 0 counter DROP\"
     expect_success "flushing the ruleset" \
         ${FROM_NS} ${BFCLI} ruleset flush
 }
 with_daemon suite_icmpv6_chain_set
+
+suite_ipv6_nexthdr_chain_set() {
+    log "[SUITE] ipv6 next-header: chain set"
+    expect_success "can parse ipv6 next-header by TC chain" \
+	    ${FROM_NS} ${BFCLI} ruleset set --from-str \"chain xdp BF_HOOK_TC_INGRESS\{ifindex=${NS_IFINDEX}\} ACCEPT rule ip6.nexthdr eq hop counter DROP\"
+    expect_success "can parse ipv6 next-header by XDP chain" \
+	    ${FROM_NS} ${BFCLI} ruleset set --from-str \"chain xdp BF_HOOK_XDP\{ifindex=${NS_IFINDEX}\} ACCEPT rule ip6.nexthdr not tcp counter DROP\"
+    expect_success "flushing the ruleset" \
+        ${FROM_NS} ${BFCLI} ruleset flush
+}
+with_daemon suite_ipv6_nexthdr_chain_set
 
 suite_chain_set() {
     log "[SUITE] chain: set"
--- a/doc/usage/bfcli.rst
+++ b/doc/usage/bfcli.rst
@@ -478,7 +478,12 @@ IPv6
       - ``ip6.dnet``
       - ``in``
       - ``{$IP/$MASK[,...]}``
-
+    * - :rspan:`1` Next header
+      - :rspan:`1` ``ip6.nexthdr``
+      - ``eq``
+      - :rspan:`3` ``$NEXT_HEADER``
+      - :rspan:`3` ``$NEXT_HEADER`` can be one of the following strings: ``ah``, ``dst``, ``frag``, ``hop``, ``icmpv6``, ``mh``, ``route``, ``tcp``, ``udp``
+    * - ``not``
 
 TCP
 ###
--- a/src/bfcli/lexer.l
+++ b/src/bfcli/lexer.l
@@ -29,6 +29,7 @@
 %s STATE_MATCHER_IP4_NET
 %s STATE_MATCHER_IP6_ADDR
 %s STATE_MATCHER_IP6_NET
+%s STATE_MATCHER_IP6_NEXTHDR
 %s STATE_MATCHER_PORT
 %s STATE_MATCHER_ICMP
 %s STATE_MATCHER_TCP_FLAGS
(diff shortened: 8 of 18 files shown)


src/bpfilter/cgen/stub.c
```diff
@@ -257,6 +257,39 @@ int bf_stub_parse_l4_hdr(struct bf_program *program)
...
+                          BPF_MOV64_IMM(BPF_REG_4, 2));
+        EMIT_SWICH_OPTION(&swich, IPPROTO_MH,
+                          BPF_JMP_IMM(BPF_JEQ, BPF_REG_7, htobe16(ETH_P_IPV6), 2),
+                          BPF_MOV64_IMM(BPF_REG_4, 0),
+                          BPF_JMP_A(1),
+                          BPF_MOV64_IMM(BPF_REG_4, 2));
+        /* managed by default case
+         * EMIT_SWICH_OPTION(&swich, IPPROTO_NONE,
+         *                  BPF_MOV64_IMM(BPF_REG_8, 0));
+         */
```
qdeslandes: The IPv6 extension headers should not be parsed from a stub function, mainly because it's part of the IPv6 protocol and not an L4 protocol header. Furthermore, an IPv6 packet can contain more than 1 extension header, and the logic above only parses the first one.

Supporting IPv6 next header field requires two main changes:
- Handle the next header field when parsing the L3 protocol: account for the next header field(s) (if any) when calculating the size of the IPv6 header, which gives us the correct offset where the L4 header is located. This is cumbersome as we must read the next header field, check if it's a proper protocol or an extension header, and continue as long as we read extension headers.
- Support `ipv6.nexthdr`, which is the purpose of this PR. 

With #1 missing, #2 won't be able to properly filter the IPv6 extension headers.

qdeslandes:
@vinxcls I have a solution wrt issue #1 in my previous comment, I should be able to merge it this week.

I'm improving the packet preprocessing steps for IPv6, so the extension headers are skipped until the actual L4 header is found. Hence, the L4 header stored in `BPF_REG_8` will be accurate even if the IPv6 packet contains extension headers.

For this PR, it means all the logic will be defined in a matcher, it will be a bit more complex though:
- If packet is IPv6 and `ipv6hdr.nexthdr != BPF_REG_8` it means the packet contains extension headers
- Create a dynamic pointer slice of starting at `ipv6hdr + 1` with size=2, it should create a slice containing the first extension header
- Check the header for the matcher's operator
- Continue with the next extension header (if `nexthdr != BPF_REG_8`)

This logic will be a pain to write in BPF bytecode, especially with the loop to process all the headers, but the fix I mentioned above provides a way to write it in C and integrate it into the final BPF program (hopefully that use case will be easy to implement).

vinxcls:
I was trying to implement the same issue in BPF bytecode, but I don't have a working solution... I'd prefer to wait for your merge and continue from there to complete this PR.

Thank you

qdeslandes:
Hi @vinxcls ,

The IPv6 extension headers are not properly processed by bpfilter (i.e. if extension headers are defined, the BPF program will skip then to process the L4 header). See https://bpfilter.io/developers/generation.html#elf-stubs for the actually mechanism.

src/bfcli/parser.y
```diff
@@ -479,6 +481,114 @@ matcher         : matcher_type matcher_op MATCHER_META_IFINDEX
...
+                        else if (bf_streq($3, "tcp"))
+                            value = IPPROTO_TCP;
+                        else if (bf_streq($3, "udp"))
+                            value = IPPROTO_UDP;
+                        else if (bf_streq($3, "icmpv6"))
+                            value = IPPROTO_ICMPV6;
+                        else {
+                            value = 0x100;
+                            bf_parse_err("unsupported %s value '%s'\n",
+                                         bf_matcher_type_to_str($1), nexthdr);
```
qdeslandes: `bf_parse_err()` is enough, as it will print the error message and stop `bfcli`, so `value < 0x100` will never be reached

src/bpfilter/cgen/stub.c
```diff
@@ -229,20 +229,22 @@ int bf_stub_parse_l3_hdr(struct bf_program *program)
...
 
-        /* Fast path for TCP and UDP: quickly recognize the most used protocol
+        /* Fast path for TCP, UDP and ICMPv6: quickly recognize the most used protocol
          * to process them as fast as possible. */
         tcpjmp = bf_jmpctx_get(program,
                                BPF_JMP_IMM(BPF_JEQ, BPF_REG_8, IPPROTO_TCP, 0));
         udpjmp = bf_jmpctx_get(program,
                                BPF_JMP_IMM(BPF_JEQ, BPF_REG_8, IPPROTO_UDP, 0));
+        icmp6jmp = bf_jmpctx_get(program,
+                                 BPF_JMP_IMM(BPF_JEQ, BPF_REG_8, IPPROTO_ICMPV6, 0));
```
qdeslandes: We can remove this. ICMPv6 will still be processed properly through the catch-all below, it doesn't need to be fast-tracked

src/bpfilter/cgen/matcher/ip6.c
```diff
@@ -155,6 +156,68 @@ static int _bf_matcher_generate_ip6_addr(struct bf_program *program,
...
+}
+
+static int _bf_matcher_generate_ip6_nexthdr(struct bf_program *program,
+                                            const struct bf_matcher *matcher)
+{
+    const uint8_t nexthdr = matcher->payload[0];
+
+    EMIT(program, BPF_MOV64_REG(BPF_REG_1, BPF_REG_10));
+    EMIT(program, BPF_ALU64_IMM(BPF_ADD, BPF_REG_1, BF_PROG_CTX_OFF(arg)));
+    EMIT(program, BPF_MOV64_IMM(BPF_REG_2, nexthdr));
```
qdeslandes: If `matcher->op` is `BF_MATCHER_IN`, `matcher->payload[0]` is the set's ID, not a `nexthdr` value.

src/bpfilter/cgen/matcher/ip6.c
```diff
@@ -155,6 +156,68 @@ static int _bf_matcher_generate_ip6_addr(struct bf_program *program,
...
+                                 BPF_JMP_IMM(BPF_JNE, BPF_REG_0, 0, 0));
+        break;
+    case BF_MATCHER_IN:
+        return _bf_matcher_generate_ip6_nexthdr_set(program, matcher);
+    default:
+        return -EINVAL;
+    }
+
+    return 0;
+}
```
qdeslandes: Pleased about the ELF stub usage here, however, I think there is a better approach. 

During pre-processing of the IPv6 packet, we already go through all the extension headers (if any). Maybe we should use this pre-processing to store a bitmask of the extension headers available (but not the L4 protocol, as it's stored in a register already)? This way we don't have to loop over those again.

For example:
```
// Matcher payload
struct XXX {
    uint8_t eh;
    uint8_t l4_proto;
};

// During preprocessing
uint8_t eh = 0;
...
case IPPROTO_HOP:
  eh |= BF_IPV6_EH_HOP;
  ...
  break;
...

ctx->ipv6_eh = eh;

// During matching
if (matcher->payload & BF_IPV6_EH_HOP) {
    EMIT(program, r2 = ctx->ipv6_eh & BF_IPV6_EH_HOP);
    EMIT(program, if r2 == 0 jmp + XXX);
}

if (matcher->payload & BF_IPV6_EH_AH)
...

if (matcher->payload->l4_proto) {
    // Check L4 proto
}
```

I think this PR should focus on the `eq` and `not` operators first and implement those efficiently (ideally without re-parsing the EH and without the ELF stubs as this has an overhead). 

For the `in` operator, maybe it should be discussed in a second step, as I'm not sure it's really needed.

vinxcls: Got it! I’ll probably push the updates tomorrow.  🙂
Thanks for your feedback!

src/bfcli/parser.y
vinxcls: Removed

src/bpfilter/cgen/matcher/ip6.c
vinxcls: The ``in`` matcher is now removed

src/bpfilter/cgen/stub.c
vinxcls: Done!

src/bpfilter/bpf/parse_ipv6_eh.bpf.c
```diff
@@ -12,42 +12,87 @@
...
+
+        ctx->l4_offset += (ext->hdrlen + offset) << shift;
+        next_hdr_type = ext->nexthdr;
     }
 
+exit:
+    ctx->ipv6_nh = nh_mask;
+
     return next_hdr_type;
 }
```
qdeslandes: I'll answer to your comment about the performance of the stub here directly.

Firstly, the end-to-end tests running slower is not indicative of the performance of the BPF program. Obviously, if it takes 5 minutes to run `make e2e`, there is an issue somewhere, but don't read too much into it if it's slightly slower.

To evaluate the performance of bpfilter or the generated BPF program, refer to `make benchmark`. However, benchmarking is *hard*, as it greatly depends on the machine and how it's configured (hyperthreading, turbo boost, concurrent processes...).

Locally, I use a LattePanda Delta 3 x86-64 SBC running on Fedora 42, which is dedicated to benchmarking bpfilter. Nothing except for bpfilter's required packages are installed, hyperthreading is disabled as well as turbo boost and Intel sleep states. This setup gives me fairly reproducible benchmarks.

Now to the actual performance: `make benchmark` doesn't run benchmarks against the network packets generated dynamically (using Python's scapy) like end-to-end tests does, so I used [this patch](https://gist.github.com/qdeslandes/cb6e67957133c37fb4921f3dcfba6e97) to create a new benchmark for IPv6 extension headers processing. I've run this benchmark for the baseline bpfilter (without your changes), your changes ("PR"), and some changes I've made on top of your PR ("Sequential"). Keep in mind, this benchmark is running on a slow SBC, so you will definitely have different results:

| Test       | Stub size (#insn) | Runtime (ns) |
|-----------:|------------------:|-------------:|
| Baseline   | 209               | 78.0         |
| PR         | 491               | 79.6         |
| Sequential | 228               | 78.6         |

The runtime difference between the three solutions is quite small, the main difference is the number of instructions in the program. Because of the various jumps (`goto`) you use, clang will generate a lot of BPF instructions (330 instructions to be precise, see `llvm-objdump -d $ELF`). I've used a simpler implementation which is slightly faster (and more readable):

```c
__u8 bf_parse_ipv6(struct bf_runtime *ctx)
{
    struct ipv6hdr *ip6hdr = ctx->l3_hdr;
    __u8 next_hdr_type = ip6hdr->nexthdr;

    ctx->l4_offset = ctx->l3_offset + sizeof(struct ipv6hdr);

    for (int i = 0; i < BF_IPV6_EXT_MAX_CHAIN; ++i) {
        struct ipv6_opt_hdr _ext;
        struct ipv6_opt_hdr *ext =
            bpf_dynptr_slice(&ctx->dynptr, ctx->l4_offset, &_ext, sizeof(_ext));
        if (!ext)
            return 0;

        switch (next_hdr_type) {
        case IPPROTO_HOPOPTS:
            ctx->ipv6_nh |= BF_IPV6_NH_HOPOPTS;
            next_hdr_type = ext->nexthdr;
            ctx->l4_offset += (ext->hdrlen + 1) * 8;
            break;
        case IPPROTO_DSTOPTS:
            ctx->ipv6_nh |= BF_IPV6_NH_DSTOPTS;
            next_hdr_type = ext->nexthdr;
            ctx->l4_offset += (ext->hdrlen + 1) * 8;
            break;
        case IPPROTO_ROUTING:
            ctx->ipv6_nh |= BF_IPV6_NH_ROUTING;
            next_hdr_type = ext->nexthdr;
            ctx->l4_offset += (ext->hdrlen + 1) * 8;
            break;
        case IPPROTO_MH:
            ctx->ipv6_nh |= BF_IPV6_NH_MH;
            next_hdr_type = ext->nexthdr;
            ctx->l4_offset += (ext->hdrlen + 1) * 8;
            break;
        case IPPROTO_AH:
            ctx->ipv6_nh |= BF_IPV6_NH_AH;
            next_hdr_type = ext->nexthdr;
            ctx->l4_offset += (ext->hdrlen + 2) * 4;
            break;
        case IPPROTO_FRAGMENT:
            ctx->ipv6_nh |= BF_IPV6_NH_FRAGMENT;
            next_hdr_type = ext->nexthdr;
            ctx->l4_offset += ext->hdrlen + 8;
            break;
        default:
            return next_hdr_type;
        }
    }

    return next_hdr_type;
}
```

A couple of comments related to your implementation:
- `nh_mask |= ...;` then `ctx->ipv6_nh = nh_mask;` is not faster, it only generates more instructions. clang will store `nh_mask` on the stack, so `nh_mask |= ...` leads to a read from the stack to a register (`BPF_LDX_MEM()`), then an update of the value (`BPF_ALU64_IMM()`), then a write to the stack (`BPF_STX_MEM()`). When all EH are processed, `ctx->ipv6_nh = nh_mask;` will trigger a read from `nh_mask` into a register, then a write from the register to `ctx->ipv6_nh`. Hence, doing `ctx->ipv6_nh |= ...` directly is actually simpler.
- `goto` will make the program jump back and forth between various places. It's not a big issue by itself, but clang will generate a more complex program to ensure the registers are properly populated depending on the jump source, etc... Usually, more bytecode = slower program.

This stub sent me into a very interesting rabbit hole!

src/bpfilter/cgen/matcher/ip6.c
```diff
@@ -213,18 +238,24 @@ int bf_matcher_generate_ip6(struct bf_program *program,
...
+        EMIT(program,
+             BPF_LDX_MEM(BPF_DW, BPF_REG_6, BPF_REG_10, BF_PROG_CTX_OFF(l3_hdr)));
+
         r = _bf_matcher_generate_ip6_addr(program, matcher);
         break;
     case BF_MATCHER_IP6_SNET:
     case BF_MATCHER_IP6_DNET:
+        EMIT(program,
+             BPF_LDX_MEM(BPF_DW, BPF_REG_6, BPF_REG_10, BF_PROG_CTX_OFF(l3_hdr)));
+
```
qdeslandes: This change can be removed. It creates noise in the diff and won't have any measurable impact on the performance.

tests/e2e/cli.sh
```diff
@@ -422,13 +422,24 @@ suite_icmpv6_chain_set() {
     log "[SUITE] icmpv6: chain set"
     expect_success "can parse icmpv6 type and code by TC chain" \
 	    ${FROM_NS} ${BFCLI} ruleset set --from-str \"chain xdp BF_HOOK_TC_INGRESS\{ifindex=${NS_IFINDEX}\} ACCEPT rule icmpv6.type eq 128 icmpv6.code eq 0 counter DROP\"
-    expect_success "can parse icmpv6 type and code by TC chain" \
-	    ${FROM_NS} ${BFCLI} ruleset set --from-str \"chain xdp BF_HOOK_XDP\{ifindex=${NS_IFINDEX}\} ACCEPT rule icmp.type eq 8 icmp.code eq 0 counter DROP\"
+    expect_success "can parse icmpv6 type and code by XDP chain" \
+	    ${FROM_NS} ${BFCLI} ruleset set --from-str \"chain xdp BF_HOOK_XDP\{ifindex=${NS_IFINDEX}\} ACCEPT rule icmpv6.type eq 8 icmp.code eq 0 counter DROP\"
```
qdeslandes, line 426: Good catch!

src/core/matcher.h
```diff
@@ -119,6 +121,23 @@ struct bf_matcher_ip6_addr
...
+    BF_IPV6_NH_ROUTING = 1,
+    BF_IPV6_NH_FRAGMENT = 2,
+    BF_IPV6_NH_DSTOPTS = 3,
+    BF_IPV6_NH_MH = 4,
+    BF_IPV6_NH_AH = 5,
+    BF_IPV6_NH_TCP = 6,
+    BF_IPV6_NH_UDP = 7,
+    BF_IPV6_NH_ICMPV6 = 8,
+    _BF_MATCHER_IPV6_NH_MAX,
+};
```
qdeslandes: Similar enums don't set a specific value, no need to do it here

src/core/matcher.c
```diff
@@ -244,3 +247,40 @@ int bf_matcher_tcp_flag_from_str(const char *str,
...
+    bf_assert(0 <= nexthdr && nexthdr < _BF_MATCHER_IPV6_NH_MAX);
+
+    return _bf_matcher_ipv6_nh_strs[nexthdr];
+}
+
+int bf_matcher_ipv6_nh_from_str(const char *str,
+                                enum bf_matcher_ipv6_nh *nexthdr)
+{
+    bf_assert(str);
+    bf_assert(nexthdr);
```
qdeslandes: You can group the NULL checks together (that's what the codebase should do, but some places are not up to date): `bf_assert(str && nexthdr)`.

src/core/matcher.c
```diff
@@ -222,6 +223,8 @@ static const char *_bf_matcher_tcp_flags_strs[] = {
     [BF_MATCHER_TCP_FLAG_ECE] = "ECE", [BF_MATCHER_TCP_FLAG_CWR] = "CWR",
 };
 
+static_assert(ARRAY_SIZE(_bf_matcher_tcp_flags_strs) == _BF_MATCHER_TCP_FLAG_MAX);
```
qdeslandes, line 226: 👍

src/bpfilter/cgen/matcher/ip6.c
```diff
@@ -205,6 +205,31 @@ static int _bf_matcher_generate_ip6_net(struct bf_program *program,
...
+    case BF_MATCHER_NE:
+        EMIT_FIXUP_JMP_NEXT_RULE(program,
+                                 BPF_JMP_IMM(BPF_JNE, BPF_REG_6, 0, 0));
+        break;
+    default:
+        return -EINVAL;
+    }
+
+    return 0;
+}
```
qdeslandes: I think the approach should be slightly different. With the current implementation:
- Either the user matches against one of the supported protocols, or they can't use `ip6.nexthdr`
- Because of the 16-bits mask, we can't filter on more than 16 different `nexthdr` values
- Any new supported protocol requires a change to the stub

Here's what I would suggest, let me know what you think:
- Because `eq` and `in` allows a single `nexthdr` to match on, we can convert the value of the user-provided `nexthdr` (as `IPPROTO_XXX` value), no need for a bitmask or a custom enum value
- Only parse EH in the stub, ignore the L4 protocol (except for returning its ID)
- During the bytecode generation, we check which `IPPROTO_XXX` value we need to match against: if it's an EH identifier, we check against `ctx->ipv6_eh` with a logical `and` as you did. If it's an actual protocol value, we know it won't be in `ctx->ipv6_eh` but in `r8` already, so we compare the raw `IPPROTO_XXX` value against it.

qdeslandes (changes requested):

src/bpfilter/bpf/parse_ipv6_eh.bpf.c
vinxcls: Just to dig info this rabbit hole... today I had another idea to implement the same, basically removing the switch case and the bytecode is near 150 instructions instead of more 330. But the program needs much more time to be loaded and the branch prediction made by the switch case + clang seems more efficient of this new implementation. If you want i can show here the PoC.

qdeslandes: Sure, happy to have a look!

src/bpfilter/cgen/matcher/ip6.c
vinxcls: It's ok for you to convert ``ctx->ipv6_eh`` from ``__u16`` to ``__u64`` in ``bf_runtime`` and implement in the marcher something like this?
```
for (int i = 0; i < 6; ++i) {
    if (((ctx->ipv6_eh >> (8 * i)) & 0xff) == target)
        return true;
}
return false;
```

src/bpfilter/bpf/parse_ipv6_eh.bpf.c
vinxcls: Here it is... I have reworked a bit and I think with this version the total instructions are 95 and it fixes also the loading overthead.

```c
#define BF_IPV6_NH_FAST_PATH        0x1C0

#define BF_IPV6_NH_HOPOPTS(x)       (x << 0)
#define BF_IPV6_NH_ROUTING(x)       (x << 1)
#define BF_IPV6_NH_FRAGMENT(x)      (x << 2)
#define BF_IPV6_NH_DSTOPTS(x)       (x << 3)
#define BF_IPV6_NH_MH(x)            (x << 4)
#define BF_IPV6_NH_AH(x)            (x << 5)
#define BF_IPV6_NH_TCP(x)           (x << 6)
#define BF_IPV6_NH_UDP(x)           (x << 7)
#define BF_IPV6_NH_ICMPV6(x)        (x << 8)

__u8 bf_parse_ipv6(struct bf_runtime *ctx)
{
    struct ipv6hdr *ip6hdr = ctx->l3_hdr;
    __u8 next_hdr_type = ip6hdr->nexthdr;
    ctx->ipv6_nh = 0;
    ctx->l4_offset = ctx->l3_offset + sizeof(struct ipv6hdr);

    for (int i = 0; i < BF_IPV6_EXT_MAX_CHAIN; ++i) {
        struct ipv6_opt_hdr _ext;
        struct ipv6_opt_hdr *ext =
            bpf_dynptr_slice(&ctx->dynptr, ctx->l4_offset, &_ext, sizeof(_ext));

        if (!ext) {
            next_hdr_type = 0;
            break;
        }

        ctx->ipv6_nh |= BF_IPV6_NH_TCP((next_hdr_type == IPPROTO_TCP)) |
                   BF_IPV6_NH_UDP((next_hdr_type == IPPROTO_UDP)) |
                   BF_IPV6_NH_ICMPV6((next_hdr_type == IPPROTO_ICMPV6));

        if ((ctx->ipv6_nh & BF_IPV6_NH_FAST_PATH) != 0)
            break;

        ctx->ipv6_nh |= BF_IPV6_NH_HOPOPTS((next_hdr_type == IPPROTO_HOPOPTS)) |
                   BF_IPV6_NH_DSTOPTS((next_hdr_type == IPPROTO_DSTOPTS)) |
                   BF_IPV6_NH_ROUTING((next_hdr_type == IPPROTO_ROUTING)) |
                   BF_IPV6_NH_MH((next_hdr_type == IPPROTO_MH));

        __u8 offset = EH_COMMON_OFFSET;
        __u8 shift = EH_COMMON_SHIFT;

        if (next_hdr_type == IPPROTO_AH) {
            ctx->ipv6_nh |= BF_IPV6_NH_AH(1);
            offset = EH_AH_OFFSET;
            shift = EH_AH_SHIFT;
        } else if (next_hdr_type == IPPROTO_FRAGMENT) {
            ctx->ipv6_nh |= BF_IPV6_NH_FRAGMENT(1);
            offset = EH_FRAG_OFFSET;
            shift = EH_FRAG_SHIFT;
        }

        if (ctx->ipv6_nh == 0)
            break;

        ctx->l4_offset += (ext->hdrlen + offset) << shift;
        next_hdr_type = ext->nexthdr;
    }

    return next_hdr_type;
}
```

qdeslandes: I see... The issue here is that the fast path is part of the stub, and the call to the stub itself induces overhead.

That's actually how I implemented the stub originally: it would perform all the IPv6 header processing, but the function call introduced a 5ns to 8ns overhead. It has to create a new stack frame, create a dynamic pointer in *any case*, then check for fast path on every loop.

src/bpfilter/cgen/matcher/ip6.c
qdeslandes: I'm not sure I see the benefit over a logical `and` which would check all EH at once?

vinxcls: I suppose it's not possible to detect the occurrence of one byte with a mask and a logical ``and``.

I can try to implement something like this ``ipv6_eh ^ pattern`` and then apply this algorithm: https://graphics.stanford.edu/~seander/bithacks.html#ZeroInWord

src/bpfilter/bpf/parse_ipv6_eh.bpf.c
vinxcls: I suppose the use of ``bpf_dynptr_slice()`` is unavoidable.

src/bpfilter/cgen/matcher/ip6.c
vinxcls: Done

src/core/matcher.h
vinxcls: I have used the ``IPPROTO_XXX`` instead, hope it's ok

src/core/matcher.c
vinxcls: Done

src/bpfilter/cgen/matcher/ip6.c
vinxcls: Due to limitations with 64-bit constants, I divided the algorithm into two parts. I believe there's room for improvement, so I'm open to any suggestions.

src/bfcli/parser.y
```diff
@@ -608,6 +609,20 @@ matcher         : matcher_type matcher_op MATCHER_META_IFINDEX
...
                 }
+                | matcher_type matcher_op MATCHER_IP6_NEXTHDR
+                {
+                    _free_bf_matcher_ struct bf_matcher *matcher = NULL;
+                    enum bf_matcher_ipv6_nh nexthdr;
+
+                    if (bf_matcher_ipv6_nh_from_str($3, &nexthdr))
+                        bf_parse_err("Unknown IPv6 next-header '%s', ignoring\n", $3);
+                    else if (bf_matcher_new(&matcher, $1, $2, &nexthdr, sizeof(nexthdr)) < 0)
+                        bf_parse_err("failed to create a new matcher\n");
```
qdeslandes: Split this in two `if`, first convert the text into a matcher, then create the matcher. It will be easier to read and maintain over time.

src/bpfilter/cgen/matcher/ip6.c
```diff
@@ -175,18 +175,18 @@ static int _bf_matcher_generate_ip6_net(struct bf_program *program,
...
              BPF_STX_MEM(BPF_W, BPF_REG_10, BPF_REG_3, BF_PROG_SCR_OFF(4)));
 
         EMIT(program, BPF_LDX_MEM(BPF_DW, BPF_REG_2, BPF_REG_6, offset));
         EMIT(program,
-            BPF_STX_MEM(BPF_DW, BPF_REG_10, BPF_REG_2, BF_PROG_SCR_OFF(8)));
+             BPF_STX_MEM(BPF_DW, BPF_REG_10, BPF_REG_2, BF_PROG_SCR_OFF(8)));
         EMIT(program, BPF_LDX_MEM(BPF_DW, BPF_REG_2, BPF_REG_6, offset + 8));
         EMIT(program,
-            BPF_STX_MEM(BPF_DW, BPF_REG_10, BPF_REG_2, BF_PROG_SCR_OFF(16)));
+             BPF_STX_MEM(BPF_DW, BPF_REG_10, BPF_REG_2, BF_PROG_SCR_OFF(16)));
```
qdeslandes: Is this change due to clang-format?

src/bpfilter/bpf/parse_ipv6_eh.bpf.c
```diff
@@ -12,41 +12,53 @@
...
+        if (next_hdr_type == IPPROTO_AH) {
+            offset = EH_AH_OFFSET;
+            shift = EH_AH_SHIFT;
+        } else if (next_hdr_type == IPPROTO_FRAGMENT) {
+            offset = EH_FRAG_OFFSET;
+            shift = EH_FRAG_SHIFT;
+        } else {
+            offset = EH_COMMON_OFFSET;
+            shift = EH_COMMON_SHIFT;
         }
```
qdeslandes: As you've moved away from the `switch...case` construct, you're only handling TCP, UDP and ICMP. But this stub should be able to accept any L4 protocol (hence the default in the switch case).

src/bpfilter/cgen/matcher/ip6.c
```diff
@@ -205,6 +205,73 @@ static int _bf_matcher_generate_ip6_net(struct bf_program *program,
...
+
+        //Merge results
+        EMIT(program, BPF_ALU64_REG(BPF_OR, BPF_REG_4, BPF_REG_5));
+
+        EMIT_FIXUP_JMP_NEXT_RULE(
+            program,
+            BPF_JMP_IMM((matcher->op == BF_MATCHER_EQ) ? BPF_JEQ : BPF_JNE,
+                        BPF_REG_4, 0, 0));
+    }
+    return 0;
```
qdeslandes, line 253: I think there is a misunderstanding here (or maybe I got this wrong, in which case feel free to correct me): 

There are two different cases `ip6.nexthdr` bytecode should handle:
1. The payload is an identifier for an EH
2. The payload is an identifier for a L4 protocol

Both are valid values for IPv6's next header field.

During pre-processing, if an EH is detected, the stub will create a bitmap of the EH defined in the packet:
- `1 << 0`: hop by hop
- `1 << 1`: fragment
- ...

Now, when generation the bytecode, we need to handle case 1 and 2. Case 2 we compare the payload against r8 as you did. For case 1 we will check if the requested EH bit is set in `ctx->ipv6_eh`: 
```
r1 = ctx->ipv6_eh
r2 = 1 << 0 # Assuming ip6.nexthdr eq hop
r1 &= r2
if r1 == 0: goto next rule # No match
... # Match
```

src/core/matcher.c
```diff
@@ -172,8 +170,7 @@ const char *bf_matcher_type_to_str(enum bf_matcher_type type)
 
 int bf_matcher_type_from_str(const char *str, enum bf_matcher_type *type)
 {
-    bf_assert(str);
-    bf_assert(type);
+    bf_assert(str && type);
```
qdeslandes: As a general rule, do not fix the style of a function you do not modify. At some point it will be fixed, but passing over it in this PR only creates noise.

src/core/matcher.c
```diff
@@ -202,8 +199,7 @@ const char *bf_matcher_op_to_str(enum bf_matcher_op op)
 
 int bf_matcher_op_from_str(const char *str, enum bf_matcher_op *op)
 {
-    bf_assert(str);
-    bf_assert(op);
+    bf_assert(str && op);
```
qdeslandes: Same

src/core/matcher.c
```diff
@@ -232,8 +230,7 @@ const char *bf_matcher_tcp_flag_to_str(enum bf_matcher_tcp_flag flag)
 int bf_matcher_tcp_flag_from_str(const char *str,
                                  enum bf_matcher_tcp_flag *flag)
 {
-    bf_assert(str);
-    bf_assert(flag);
+    bf_assert(str && flag);
```
qdeslandes: Same

qdeslandes (changes requested):

src/bpfilter/cgen/matcher/ip6.c
vinxcls: yes

src/core/matcher.c
vinxcls: ok

src/bpfilter/cgen/matcher/ip6.c
vinxcls: Maybe I did a mistake... with the last commit bfcli converts the string values into the actual protocol numbers. As a result, the ``ipv6_eh`` field contains the raw bytes of the extension header chain.
For example, if a packet has a Hop-by-Hop, then a Routing header, and finally TCP, the parser currently sets ``ctx->ipv6_eh`` like this: ``ctx->ipv6_eh = IPPROTO_HOPOPTS | (IPPROTO_ROUTING << 8)``, while ``BPF_REG_8`` contains the value ``IPPROTO_TCP``.

This setup complicates the matching logic quite a bit, as you can see.
If you're okay, I can revert the recent changes and switch to using a bitmask to represent the EHs that should be matched.

So, is it better to have a small stub and a more complex matcher or the opposite? 

With a small stub we're around 60 instructions... whereas the large one is closer to 100.  
For the matcher, the complex version needs 16 instructions, while the simpler one takes 12.
Those results are keeping the chain history inside ``ipv6_eh``, as we can process each element of the chain (especially with duplicate headers).  
Instead if we just use a bitmask that shows only the header of the chain (but not the occurrence), the stub would grow to ~80 instructions, but the matcher would be less than 10 instructions.

IMHO, it's better to keep the stub as small as possible and delegate the complexity to the matcher. However, if the EH history isn't required, then shifting the logic into a larger stub (e.g., using a bitmask) can be a good tradeoff, as it simplifies the matcher and still performs well.

@qdeslandes what's your opinion about this?

src/bpfilter/bpf/parse_ipv6_eh.bpf.c
vinxcls: Ok, I’ll figure out a way to manage only extension headers, avoiding any reliance on L4 protocols.

src/bpfilter/cgen/matcher/ip6.c
qdeslandes: Thanks for your patience @vinxcls , this PR had slipped my mind.

I understand your logic. As a general rule, because the stub (this stub at least) is always inserted in the final program, it's better to shift the cost towards the matcher. Especially with `ip6.nexthdr` which I don't expect to be used very frequently.

That being said, the matcher as-is is very complex, and we don't need to maintain an EH history. IMO the cost of the bitmask is justified here, especially as:
- 20 extra instructions is not much, and 99% of the traffic will use the fast path and skip the stub altogether
- Eventually, we can define a stub which stores EH into the runtime struct and one which doesn't, then insert the correct one during generation depending on `ip6.nexthdr` usage

Nonetheless, your implementation is quite clever!

Let me know what you think about this :)

src/bpfilter/cgen/matcher/ip6.c
```diff
@@ -205,6 +209,51 @@ static int _bf_matcher_generate_ip6_net(struct bf_program *program,
...
+        return -EINVAL;
+
+    if ((ehdr == IPPROTO_TCP) || (ehdr == IPPROTO_UDP) ||
+        (ehdr == IPPROTO_ICMPV6)) {
+        /* check l4 protocols using BPF_REG_8 */
+        EMIT_FIXUP_JMP_NEXT_RULE(
+            program,
+            BPF_JMP_IMM((matcher->op == BF_MATCHER_EQ) ? BPF_JNE : BPF_JEQ,
+                        BPF_REG_8, ehdr, 0));
+    } else {
```
qdeslandes: This condition should be reversed:
```
if (is_eh(ehdr)) {
    ....
} else {
    ....
}
```

Eventually, the user will be able to use other protocols than TCP/UDP/ICMPv6 (even provide an Internet Protocol number directly), and we won't have to modify the bytecode generation.

src/bpfilter/cgen/stub.c
```diff
@@ -231,9 +232,16 @@ int bf_stub_parse_l3_hdr(struct bf_program *program)
...
         struct bpf_insn ld64[2] = {BPF_LD_IMM64(BPF_REG_2, _BF_LOW_EH_BITMASK)};
+        struct bpf_insn negative_one[2] = {
+            BPF_LD_IMM64(BPF_REG_2, _BF_NEGATIVE_ONE)};
         _clean_bf_jmpctx_ struct bf_jmpctx _ = bf_jmpctx_get(
             program, BPF_JMP_IMM(BPF_JNE, BPF_REG_7, htobe16(ETH_P_IPV6), 0));
 
+        EMIT(program, negative_one[0]);
+        EMIT(program, negative_one[1]);
+        EMIT(program, BPF_STX_MEM(BPF_W, BPF_REG_10, BPF_REG_2,
+                                  BF_PROG_CTX_OFF(ipv6_eh)));
```
qdeslandes: W.r.t. my previous (long) comment: this is the kind of bytecode we need to limit, as it will affect every packet, especially before the TCP/UDP fast path.

Issue #188, opened 2025-01-21
Publish `bpfilter`'s DEB package


qdeslandes:
Package `bpfilter` into a DEB package and publish it to official repositories for Ubuntu.

PR #315, opened 2025-08-20
doc: make install requires root

AliGhaffarian:
When trying to build from source, make command requires root:

```
CMake Error at src/bpfilter/cmake_install.cmake:52 (file):
  file INSTALL cannot copy file
  "/home/user/Clones/bpfilter/build/output/sbin/bpfilter" to
  "/usr/local/sbin/bpfilter": Permission denied.
Call Stack (most recent call first):
  cmake_install.cmake:52 (include)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ sudo dnf install -y clang cmake gcc libbpf-devel libnl3-devel bison flex sed xxd
 
 # Configure the project and build bpfilter
 cmake -S $SOURCES_DIR -B $BUILD_DIR -DNO_DOCS=ON -DNO_TESTS=ON -DNO_CHECKS=ON -DNO_BENCHMARKS=ON
-make -C $BUILD_DIR install
+sudo make -C $BUILD_DIR install
 ```
 
 ### Usage


qdeslandes:
Thanks! Could you update the commit subject to be more in line with the existing commits? E.g "doc: make install requires root"

README.md
```diff
@@ -56,7 +56,7 @@ sudo dnf install -y clang cmake gcc libbpf-devel libnl3-devel bison flex sed xxd
 
 # Configure the project and build bpfilter
 cmake -S $SOURCES_DIR -B $BUILD_DIR -DNO_DOCS=ON -DNO_TESTS=ON -DNO_CHECKS=ON -DNO_BENCHMARKS=ON
-make -C $BUILD_DIR install
+sudo make -C $BUILD_DIR install
```
qdeslandes, line 59: The command should actually be `make -C $BUILD_DIR`: this line is used to build bpfilter, not install it. The following commands run bpfilter from the build directory.

AliGhaffarian: (thank you for having patience with me:) ) so this should be broken into this?
```
make -C $BUILD_DIR
sudo make -C $BUILD_DIR install
```

or should we just remove the `install` argument?  
i guess we can just remove the `install` argument and when #188 is resolved it doesn't matter anymore.

PR #317, opened 2025-08-27
doc: don't install while building

AliGhaffarian:
In the [build from source](https://github.com/facebook/bpfilter?tab=readme-ov-file#build-from-sources) section of the readme file, install argument is passed to make


qdeslandes:
Merged, thanks @AliGhaffarian for your contribution!

PR #319, opened 2025-09-01
build: mark git as optional dependency

sudipm-mukherjee:
A minimum build enironments might not have git installed and the build fails.
Mark 'git' as optional so that bpfilter can be built from a tarball.


sudipm-mukherjee:
> Hi @sudipm-mukherjee , thanks for doing this. Could you add a check after `find_package(Git)` and print a log (e.g. "git not found, using default project version ${DEFAULT_PROJECT_VERSION}") before returning if git is not found? Otherwise, the default `0.0.0` version could be confusing. Additionally, `-DDEFAULT_PROJECT_VERSION=x.y.z` can be used when running CMake to build without Git but with a valid version number (see [bpfilter.spec](https://src.fedoraproject.org/rpms/bpfilter/blob/rawhide/f/bpfilter.spec#_55)).

Thanks for checking. Changes done.
yeah, I noticed DEFAULT_PROJECT_VERSION after making other changes to set the version.  :)

qdeslandes (approved):

PR #321, opened 2025-09-04
[RESEND] bpfilter: xlate: use const argument when possible

AliGhaffarian:
resolves the following todo
src/bpfilter/xlate/front.h: @todo  Make bf_front_ops.request_handler take a const struct bf_request.


qdeslandes:
Excellent, thanks for fixing this todo @AliGhaffarian ! :)

qdeslandes (approved):

PR #328, opened 2025-09-14
Enforce clang-format on all source files

AliGhaffarian:
The changes are automated with this simple script:
```sh
#!/bin/bash

tmpfile=$(mktemp)
for file in $(find ./src -name '*.c');do
	clang-format $file > $tmpfile
	cat $tmpfile > $file
done
```
No change in functionality is intended.


AliGhaffarian:
Thanks, I will go through the `fixstyle` target you mentioned and see how it works.

PR #330, opened 2025-09-18
Logger enhancements

AliGhaffarian:
This pr contains two changes
- dry the logger's prefix
- on debug build, log the name of the function that invoked the logger

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,9 +28,6 @@ option(NO_BENCHMARKS "Disable the benchmark" 0)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-set(CMAKE_C_FLAGS_RELEASE "")       # No NDEBUG in release mode
-set(CMAKE_CXX_FLAGS_RELEASE "")     # No NDEBUG in release mode
-
 set(BF_CONTACT "Quentin Deslandes <qde@naccy.de>")
 
 if (NOT CMAKE_BUILD_TYPE)
--- a/src/bpfilter/xlate/cli.c
+++ b/src/bpfilter/xlate/cli.c
@@ -126,7 +126,9 @@ int _bf_cli_ruleset_set(const struct bf_request *request,
     bf_rpack_node_t child, node;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
+
+    bf_assert(request);
 
     bf_ctx_flush(BF_FRONT_CLI);
 
@@ -202,7 +204,9 @@ int _bf_cli_chain_set(const struct bf_request *request,
     bf_rpack_node_t root, child;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
+
+    bf_assert(request);
 
     r = bf_rpack_new(&pack, bf_request_data(request),
                      bf_request_data_len(request));
@@ -382,7 +386,9 @@ int _bf_cli_chain_load(const struct bf_request *request,
     bf_rpack_node_t child;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
+
+    bf_assert(request);
 
     r = bf_rpack_new(&pack, bf_request_data(request),
                      bf_request_data_len(request));
@@ -433,7 +439,9 @@ int _bf_cli_chain_attach(const struct bf_request *request,
     bf_rpack_node_t child;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
+
+    bf_assert(request);
 
     r = bf_rpack_new(&pack, bf_request_data(request),
                      bf_request_data_len(request));
@@ -477,7 +485,9 @@ int _bf_cli_chain_update(const struct bf_request *request,
     bf_rpack_node_t child;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
+
+    bf_assert(request);
 
     r = bf_rpack_new(&pack, bf_request_data(request),
                      bf_request_data_len(request));
@@ -514,7 +524,9 @@ int _bf_cli_chain_flush(const struct bf_request *request,
     _cleanup_free_ char *name = NULL;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
+
+    bf_assert(request);
 
     r = bf_rpack_new(&pack, bf_request_data(request),
                      bf_request_data_len(request));
--- a/src/bpfilter/xlate/ipt/ipt.c
+++ b/src/bpfilter/xlate/ipt/ipt.c
@@ -709,8 +709,7 @@ static int _bf_ipt_ruleset_set(const struct bf_request *req)
 static int _bf_ipt_set_counters_handler(const struct xt_counters_info *counters,
                                         size_t len)
 {
-    bf_assert(counters);
-
+    UNUSED(counters);
     UNUSED(len);
 
     return 0;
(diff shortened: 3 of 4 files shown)


src/libbpfilter/include/bpfilter/logger.h
```diff
@@ -51,6 +51,18 @@ enum bf_log_level
     _BF_LOG_MAX,
 };
 
+#ifdef BUILD_TYPE_DEBUG
+#define logger_prefix_fmt "%s%-7s[%s]%s: "
+#define logger_prefix_args(level, color)                                       \
```
qdeslandes: Prefix macros with `_bf_`

src/libbpfilter/include/bpfilter/logger.h
```diff
@@ -51,6 +51,18 @@ enum bf_log_level
     _BF_LOG_MAX,
 };
 
+#ifdef BUILD_TYPE_DEBUG
```
qdeslandes: We should rely on the standard `NDEBUG` here instead of a custom definition. This requires removing:

```
set(CMAKE_C_FLAGS_RELEASE "")       # No NDEBUG in release mode
set(CMAKE_CXX_FLAGS_RELEASE "")     # No NDEBUG in release mode
```

from `CMakeLists.txt`.

CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ add_library(bf_global_flags INTERFACE)
 target_compile_options(bf_global_flags
     INTERFACE
         -std=gnu17 -Wall -Wextra -fPIC
-        $<$<CONFIG:debug>:-O0 -g3 -ggdb -fno-omit-frame-pointer -fsanitize=address -fsanitize=undefined>
+	$<$<CONFIG:debug>:-O0 -g3 -ggdb -fno-omit-frame-pointer -fsanitize=address -fsanitize=undefined -DBUILD_TYPE_DEBUG>
```
qdeslandes: You change the indentation.

src/libbpfilter/include/bpfilter/logger.h
```diff
@@ -51,6 +51,18 @@ enum bf_log_level
...
+#define logger_prefix_fmt "%s%-7s[%s]%s: "
```
qdeslandes: The line number in addition to the function would be useful.

src/libbpfilter/include/bpfilter/logger.h
AliGhaffarian: Removing the mentioned lines from `CMakeLists.txt` didn't define `NDEBUG`, probably because compile options are defined manually (not sure). I defined it in place of previously defined `BUILD_TYPE_DEBUG` in `CMakeLists.txt`.

src/libbpfilter/include/bpfilter/logger.h
```diff
@@ -51,6 +51,19 @@ enum bf_log_level
...
+#define _bf_logger_prefix_fmt_args(level, color)                               \
+    bf_logger_get_color((color), BF_STYLE_BOLD), bf_log_level_to_str(level),   \
+        __func__, __LINE__,                                                    \
+        bf_logger_get_color(BF_COLOR_RESET, BF_STYLE_RESET)
+#else
+#define _bf_logger_prefix_fmt "%s%-7s%s: "
+#define _bf_logger_prefix_fmt_args(level, color)                               \
+    bf_logger_get_color((color), BF_STYLE_BOLD), bf_log_level_to_str(level),   \
+        bf_logger_get_color(BF_COLOR_RESET, BF_STYLE_RESET)
+#endif
```
qdeslandes, line 65: You used it the other way around, `NDEBUG` is the standard macro for `not debug`, so if it's defined it means we're not in debug mode.

CMake defines it by default in if `CMAKE_BUILD_TYPE=release`, but we override the default C flags with:
```
set(CMAKE_C_FLAGS_RELEASE "")       # No NDEBUG in release mode
set(CMAKE_CXX_FLAGS_RELEASE "")     # No NDEBUG in release mode
```

What you need to do is:
- Remove the lines disabled `NDEBUG` in release
- Reverse your condition is `NDEBUG` is defined in release mode

AliGhaffarian: oh i see, thanks

src/bpfilter/xlate/ipt/ipt.c
```diff
@@ -709,7 +709,11 @@ static int _bf_ipt_ruleset_set(const struct bf_request *req)
 static int _bf_ipt_set_counters_handler(const struct xt_counters_info *counters,
                                         size_t len)
 {
+#ifdef NDEBUG
+    UNUSED(counters);
+#else
     bf_assert(counters);
+#endif
```
qdeslandes: `bf_assert()` shouldn't have been used in the first place. If it's unused in the function, just use `UNUSED(x)`, no need for `bf_assert()`.

AliGhaffarian: assuming you were referring to all the instances in which this change was made, i replaced `bf_assert(x)` with `UNUSED(x)` where `x` was unused in the function other than `bf_assert(x)`.

src/bpfilter/xlate/cli.c
```diff
@@ -126,7 +126,7 @@ int _bf_cli_ruleset_set(const struct bf_request *request,
     bf_rpack_node_t child, node;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
```
qdeslandes, line 129: We still need to assert on `request`: it's used and should not be NULL.

src/bpfilter/xlate/cli.c
```diff
@@ -202,7 +202,7 @@ int _bf_cli_chain_set(const struct bf_request *request,
     bf_rpack_node_t root, child;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
```
qdeslandes, line 207: We still need to assert on `request`: it's used and should not be NULL.

src/bpfilter/xlate/cli.c
```diff
@@ -382,7 +382,7 @@ int _bf_cli_chain_load(const struct bf_request *request,
     bf_rpack_node_t child;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
```
qdeslandes, line 389: We still need to assert on `request`: it's used and should not be NULL.

src/bpfilter/xlate/cli.c
```diff
@@ -433,7 +433,7 @@ int _bf_cli_chain_attach(const struct bf_request *request,
...
```
qdeslandes, line 442: We still need to assert on `request`: it's used and should not be NULL.

src/bpfilter/xlate/cli.c
```diff
@@ -477,7 +477,7 @@ int _bf_cli_chain_update(const struct bf_request *request,
...
```
qdeslandes, line 488: We still need to assert on `request`: it's used and should not be NULL.

src/bpfilter/xlate/cli.c
```diff
@@ -514,7 +514,7 @@ int _bf_cli_chain_flush(const struct bf_request *request,
     _cleanup_free_ char *name = NULL;
     int r;
 
-    bf_assert(request && response);
+    UNUSED(response);
```
qdeslandes, line 527: We still need to assert on `request`: it's used and should not be NULL.

qdeslandes (approved):
Thanks @AliGhaffarian , LGTM!

PR #343, opened 2025-09-28
libbpfilter: list: implement bf_list_push

AliGhaffarian:


src/libbpfilter/include/bpfilter/list.h
```diff
@@ -286,6 +285,14 @@ static inline bool bf_list_is_tail(const bf_list *list,
     return node == list->tail;
 }
 
+/**
+ * Push node at the end of the list.
```
qdeslandes: `@brief`

src/libbpfilter/list.c
```diff
@@ -133,6 +133,25 @@ int bf_list_pack(const bf_list *list, bf_wpack_t *pack)
     return bf_wpack_is_valid(pack) ? 0 : -EINVAL;
 }
 
+void bf_list_push(bf_list *list, bf_list_node *node)
```
qdeslandes: Should be `int bf_list_push(bf_list *list, void **data)`: it adds a new node at the end of the list and steal the `data` pointer (copy `*data` and `*data = NULL`).

AliGhaffarian: There already is `int bf_list_add_tail(bf_list *list, void *data)`. Should we change it to steal the `data` pointer since there is a [todo](https://github.com/facebook/bpfilter/blob/cc3396125280d691c40ec0c56cc09a978ed81f12/src/libbpfilter/include/bpfilter/list.h#L18) mentioning this? This way `bf_list_add_tail()` essentially becomes `int bf_list_push(bf_list *list, void **data)`.

src/libbpfilter/include/bpfilter/list.h
AliGhaffarian: There are other functions missing `@brief` (for example, [here](https://github.com/facebook/bpfilter/blob/cc3396125280d691c40ec0c56cc09a978ed81f12/src/libbpfilter/include/bpfilter/list.h#L211C1-L211C20), [here](https://github.com/facebook/bpfilter/blob/cc3396125280d691c40ec0c56cc09a978ed81f12/src/libbpfilter/include/bpfilter/list.h#L183) and [here](https://github.com/facebook/bpfilter/blob/cc3396125280d691c40ec0c56cc09a978ed81f12/src/libbpfilter/include/bpfilter/list.h#L322)). Should `@brief` be added to them as well?

qdeslandes: Feel free to fix functions in the same file, but in a different commit.

src/libbpfilter/list.c
qdeslandes: Focus on `int bf_list_push(bf_list *list, void **data)` only, as `bf_list_add_tail()` is part of the API and changing it will break stuff. I will update it as soon as I ensured every usage of `bf_list_add_tail()` moved to `bf_list_push()`.

src/libbpfilter/list.c
```diff
@@ -133,6 +133,36 @@ int bf_list_pack(const bf_list *list, bf_wpack_t *pack)
...
+
+    if (!list->head) {
+        list->head = node;
+        list->tail = node;
+    } else {
+        list->tail->next = node;
+        list->tail = node;
+    }
+
+    ++list->len;
```
qdeslandes: `bf_list_add_tail()` would be simpler, this way you wouldn't have to handle the node's pointers.

tests/unit/libbpfilter/list.c
```diff
@@ -398,3 +398,40 @@ Test(list, node_take_data)
...
+    bf_list_init(&list, &free_ops);
+    expect_assert_failure(bf_list_push(NULL, NULL));
+    expect_assert_failure(bf_list_push(NULL, (void **)NOT_NULL));
+    expect_assert_failure(bf_list_push(NOT_NULL, NULL));
+
+    int *dummy_ptr = NULL;
+
+    int dummy_int = 1;
+    dummy_ptr = malloc(sizeof(dummy_int));
+    *dummy_ptr = dummy_int;
```
qdeslandes: Variables should be defined at the beginning of the current scope. Also, declare pointers with `_cleanup_free_` so it's automatically freed if not NULL.

src/libbpfilter/list.c
AliGhaffarian: Since when calling `bf_list_add_tail()` we need to deref the `void **data`, i also added a `dummpy_void_ptr_ptr` to test the `bf_list_push(NULL, NOT_NULL)` case (`NOT_NULL` itself is a constant and we can't really pass it's address).

PR #346, opened 2025-10-08
feature: rate limit

### LINKED ISSUE #215: Support traffic rate limitation
> This issue is defined to keep track of the investigation, feasibility, and progress on rate limit traffic using `bpfilter`. Design work is required before any implementation.

Allow users to rate limit specific packets, such as `ip6tables -A INPUT -p icmpv6 -m limit --limit 20/m --limit-burst 3 -j ACCEPT` with `iptables`.

See [iptables](https://ipset.netfilter.org/iptables-extensions.man.html#lbBI) and [nftables](https://wiki.nftables.org/wiki-nftables/index.php/Quick_reference-nftables_in_10_minutes#Limit) documentation.

SkohTV:
Fix https://github.com/facebook/bpfilter/issues/215

Hi, sorry for the long hiatus,
A bunch of stuff came up, and I didn't have the time to do the big rebase+refactor.

**Changes**:

- Did the big rebase.
- Changed `ratelimit` from an `action` to a `matcher`, supports for `eq` with an `uint32`
- Kept the map, because some persistence is required for rate limiting.
- Also kept and slightly changed the `elfstub`.

  | 10 first | All after
-- | -- | --
`chain my_chain BF_HOOK_XDP{ifindex=2} DROP rule meta.ratelimit eq 10 ACCEPT` | ACCEPT | DROP
`chain my_chain BF_HOOK_XDP{ifindex=2} ACCEPT rule meta.ratelimit eq 10 DROP` | DROP | ACCEPT
`chain my_chain BF_HOOK_XDP{ifindex=2} DROP rule meta.ratelimit not eq 10 ACCEPT` | DROP | ACCEPT
`chain my_chain BF_HOOK_XDP{ifindex=2} ACCEPT rule meta.ratelimit not eq 10 DROP` | ACCEPT | DROP

If that implementation seems good, I'll go ahead and add QoL / the documentation

###### I'd really like to get this PR over the finish line at some point :)


src/bfcli/parser.y
```diff
@@ -55,16 +55,18 @@
     })
 
     enum bf_rule_option_flag {
-        BF_RULE_OPTION_LOG      = 1 << 0,
-        BF_RULE_OPTION_COUNTER  = 1 << 1,
-        BF_RULE_OPTION_MARK     = 1 << 2,
+        BF_RULE_OPTION_LOG        = 1 << 0,
+        BF_RULE_OPTION_COUNTER    = 1 << 1,
+        BF_RULE_OPTION_RATELIMIT  = 1 << 2,
```
qdeslandes: The rate limit entry can be added as the last enum option.

src/bfcli/lexer.l
```diff
@@ -84,6 +85,16 @@ log             { BEGIN(STATE_LOG_OPTS); return LOG; }
...
 }
 
+    /* Rate limit  */
+ratelimit       { BEGIN(STATE_RATELIMIT_OPTS); return RATELIMIT; }
+<STATE_RATELIMIT_OPTS>{
+    [0-9]+ {
+        BEGIN(INITIAL);
+        yylval.sval = strdup(yytext);
+        return RATELIMIT_VAL;
+    }
```
qdeslandes: We should expect any float value (even negative, we'll error out later on). We also need a time window , e.g. 10/second.

src/bfcli/parser.y
```diff
@@ -315,6 +319,18 @@ rule_option     : LOG LOG_HEADERS
                         .flags = BF_RULE_OPTION_COUNTER,
                     };
                 }
+                | RATELIMIT RATELIMIT_VAL
+                {
+                    _cleanup_free_ char *in = $2;
+                    char *tmp = in;
+                    uint32_t limit = atoi(tmp);
+
+                    printf("Got %d\n", limit);
```
qdeslandes: Remove

src/bfcli/parser.y
```diff
@@ -315,6 +319,18 @@ rule_option     : LOG LOG_HEADERS
                         .flags = BF_RULE_OPTION_COUNTER,
                     };
                 }
+                | RATELIMIT RATELIMIT_VAL
+                {
+                    _cleanup_free_ char *in = $2;
+                    char *tmp = in;
+                    uint32_t limit = atoi(tmp);
```
qdeslandes: `strtoul` with overflow check would be better, see https://github.com/facebook/bpfilter/blob/main/src/libbpfilter/matcher.c#L283.

src/bpfilter/cgen/program.c
```diff
@@ -117,6 +117,11 @@ int bf_program_new(struct bf_program **program, const struct bf_chain *chain)
     if (r < 0)
         return bf_err_r(r, "failed to create the counters bf_map object");
 
+    r = bf_map_new(&_program->rmap, "ratelimit_map", BF_MAP_TYPE_RATELIMIT,
+                   sizeof(uint32_t), sizeof(struct bf_counter), 1);
+    if (r < 0)
```
qdeslandes: `if (r)` should be enough.

src/bpfilter/cgen/program.c
```diff
@@ -599,6 +626,11 @@ static int _bf_program_generate_rule(struct bf_program *program,
         EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_UPDATE_COUNTERS);
     }
 
+    // TODO: the logic goes here
+    if (rule->ratelimit) {
+        // We probably should do something here, but the EMIT are scarry
```
qdeslandes: ELF stubs are probably the most straightforward way to an initial implementation https://bpfilter.io/developers/generation.html#elf-stubs (won't save you from writing a few `EMIT` though ;) )

qdeslandes (commented):
I went for an early review, mostly for the parsing part. On the BPF side, the map will contain the runtime values for a given rule's rate limit (current burst/allowance, limit, last update time...).

src/bpfilter/cgen/program.c
SkohTV: I just followed the pattern (https://github.com/facebook/bpfilter/blob/main/src/bpfilter/cgen/program.c#L115-L128)
Should we change them too to keep it uniform ?

src/bfcli/lexer.l
SkohTV: Do we really wanna handle float or do we also error them out later on ?
It feels weird writing `12.1/second` (even though it is technically correct), and we could have issues with float precision in some cases.

src/bpfilter/cgen/program.c
qdeslandes: All good, following the pattern is correct.

src/bfcli/lexer.l
qdeslandes: We don't want to handle floats, but it's easier to error out when parsing the matcher payload ([e.g. with a clear error message](https://github.com/facebook/bpfilter/blob/main/src/libbpfilter/matcher.c#L274-L294)) than in the lexer.

SkohTV:
Here is a working POC
Not going to far, to make it easier to change/correct things if required

Things done:
- Added `ratelimit XX/XX` (ex: `ratelimit 5/sec`) to **bfcli**
- Created a map to store `current` (packets that passed in this unit of time) and `last_time` (last packet timestamp)
- Added an elfstub (and a bunch of emits) to allow only X packets to be ACCEPTED, packets over the rate limit will be DROPPED
  
Things remaining:
- Make the unit (`sec`) actually be used (atm we only do it per second)
- Handle cli syntax better (errors on negative or wrong syntax)
  
Testing:
```
$ sudo ./result/sbin/bpfilter --transient --verbose=bpf

$ sudo ./result/sbin/bfcli chain set --from-str \
	'chain my_chain BF_HOOK_XDP{ifindex=2} ACCEPT
	rule ip4.proto icmp log internet ratelimit 5/sec ACCEPT'

$ ping 8.8.8.8 -c 10 -i 0.02
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=120 time=24.7 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=120 time=18.9 ms
64 bytes from 8.8.8.8: icmp_seq=3 ttl=120 time=14.4 ms
64 bytes from 8.8.8.8: icmp_seq=4 ttl=120 time=17.1 ms
64 bytes from 8.8.8.8: icmp_seq=5 ttl=120 time=13.5 ms

--- 8.8.8.8 ping statistics ---
10 packets transmitted, 5 received, 50% packet loss, time 184ms
rtt min/avg/max/mdev = 13.462/17.706/24.692/3.982 ms, pipe 2
```

There's a bunch of doc to do and leftover comments to remove, but that should be enough to swi(t)ch this PR from draft to "ready-for-review" now that the feature is working !

src/bfcli/parser.y
qdeslandes: New options should be added at the end, otherwise it breaks the ABI.

src/bfcli/parser.y
```diff
@@ -305,6 +309,30 @@ rule_option     : LOG LOG_HEADERS
...
+                {
+                    _cleanup_free_ char *in = $2;
+                    char *tmp = in;
+                    char *saveptr;
+
+                    if (tmp[0] == '-')
+                        bf_parse_err("ratelimit should be positive");
+
+                    errno = 0;
+                    uint32_t limit = strtoul(strtok_r(tmp, "/", &saveptr), NULL, 0);
```
qdeslandes: Variables should be defined at the beginning of the scope.

src/bfcli/parser.y
```diff
@@ -305,6 +309,30 @@ rule_option     : LOG LOG_HEADERS
...
+
+                    $$ = (struct bf_rule_options){
+                        .ratelimit = limit,
+                        .flags = BF_RULE_OPTION_RATELIMIT,
+                    };
+
+                    $$ = (struct bf_rule_options){
+                        .ratelimit = limit,
+                        .flags = BF_RULE_OPTION_RATELIMIT,
+                    };
```
qdeslandes, line 398: Duplicate.

src/bpfilter/bpf/ratelimit.bpf.c
```diff
@@ -0,0 +1,34 @@
...
+
+    ratelimit = bpf_map_lookup_elem(map, &key);
+    if (!ratelimit) {
+        bpf_printk("failed to fetch the rule's ratelimit");
+        return 1;
+    }
+
+    if (current_time != ratelimit->last_time) {
+        ratelimit->current = 0;
+    }
```
qdeslandes, line 30: No braces needed.

src/bpfilter/cgen/prog/map.c
```diff
@@ -308,6 +310,12 @@ static struct bf_btf *_bf_map_make_btf(const struct bf_map *map)
         btf__add_field(kbtf, "packets", 1, 0, 0);
         btf__add_field(kbtf, "bytes", 1, 64, 0);
         break;
+    case BF_MAP_TYPE_RATELIMIT: // I have no clue what I'm doing here
+        btf__add_int(kbtf, "u64", 8, 0);
+        btf->key_type_id = btf__add_int(kbtf, "u32", 4, 0);
+        btf->value_type_id = btf__add_struct(kbtf, "bf_", 16);
+        btf__add_field(kbtf, "limit", 1, 0, 0);
+        break;
```
qdeslandes: No need for this yet.

src/bpfilter/cgen/elfstub.h
```diff
@@ -122,19 +122,22 @@ enum bf_elfstub_id
...
+     * - `map`: pointer to the logs ring buffer.
+     * - `layers`: user-requested headers to log.
+     * - `l3_proto`: layer 3 (internet) protocol identifier.
+     * - `l4_proto`: layer 4 (transport) protocol identifier.
      *
      * **Return** 0 on success, or 1 on error.
      */
     BF_ELFSTUB_LOG,
 
+    // Return 0 on ACCEPT, 1 on DROP
```
qdeslandes: Document the full elfstub.

src/bpfilter/cgen/program.c
```diff
@@ -606,6 +630,21 @@ static int _bf_program_generate_rule(struct bf_program *program,
         EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_UPDATE_COUNTERS);
     }
 
+    if (rule->ratelimit) {
+        EMIT_LOAD_RATELIMIT_FD_FIXUP(program, BPF_REG_1);
+        EMIT(program, BPF_MOV32_IMM(BPF_REG_2, rule->index));
+        EMIT(program, BPF_MOV32_IMM(BPF_REG_3, rule->ratelimit));
+        EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_RATELIMIT);
+
+        _clean_bf_jmpctx_ struct bf_jmpctx ctx =
```
qdeslandes: Variables are defined at the beginning of the scope.

src/bpfilter/cgen/program.c
```diff
@@ -606,6 +630,21 @@ static int _bf_program_generate_rule(struct bf_program *program,
...
+            bf_jmpctx_get(program, BPF_JMP_IMM(BPF_JEQ, BPF_REG_0, 0, 0));
+
+        EMIT(program,
+             BPF_MOV64_IMM(BPF_REG_0,
+                           program->runtime.ops->get_verdict(BF_VERDICT_DROP)));
+        EMIT(program, BPF_EXIT_INSN());
```
qdeslandes: We need to think this through. With the current implementation: if the rate limit is reached, all extra packet matching the rule are dropped. What if the rule already drops the packet? Then you drop all the matching packets, meaning the rate limit actually matches because the outcome is the same. 

Currently, actions (mark, log, counter) don't have any impact on whether the rule matches or not. This one does. Would this make more sense as a matcher instead? It would allow for negative matches:
- `rule meta.ratelimit 4pkt/s ACCEPT`: if the packet is within the limit, accept it
- `rule meta.ratelimit not 4pkt/s DROP`: packets within the limit are ignored, while packets above are matched and dropped (`not` or another keyword that fits better).

Seems like this would be more in line with nftables. What do you think? I'm open to alternatives here.

qdeslandes (changes requested):

src/bpfilter/cgen/program.c
SkohTV: Like this ?

```diff
if (rule->ratelimit) {
    EMIT_LOAD_RATELIMIT_FD_FIXUP(program, BPF_REG_1);
    EMIT(program, BPF_MOV32_IMM(BPF_REG_2, rule->index));
    EMIT(program, BPF_MOV32_IMM(BPF_REG_3, rule->ratelimit));
    EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_RATELIMIT);

+   {
        _clean_bf_jmpctx_ struct bf_jmpctx ctx =
            bf_jmpctx_get(program, BPF_JMP_IMM(BPF_JEQ, BPF_REG_0, 0, 0));

        EMIT(program,
             BPF_MOV64_IMM(BPF_REG_0,
                           program->runtime.ops->get_verdict(BF_VERDICT_DROP)));
        EMIT(program, BPF_EXIT_INSN());
+   }
}
```

src/bpfilter/cgen/program.c
SkohTV: I'm not super familiar with nftables, but I really like the idea of "negative matches", I think it makes rate limiting very versatile and nice to use
Changing it to a matcher seems to make more sense

qdeslandes:
> This PR has had no activity for 14 days. It will be closed in 14 more days unless it is updated. Comment or push to keep it open.

That's on me, looking at this today :)

src/bfcli/lexer.l
qdeslandes, line 1: `meta.limit` would be more suited for this.

src/bfcli/lexer.l
```diff
@@ -177,6 +178,16 @@ meta\.flow_hash     { BEGIN(STATE_MATCHER_META_FLOW_HASH); yylval.sval = strdup(
     }
 }
 
+meta\.ratelimit  { BEGIN(STATE_MATCHER_META_RATELIMIT); yylval.sval = strdup(yytext); return MATCHER_TYPE; }
+<STATE_MATCHER_META_RATELIMIT>{
+    (eq) { yylval.sval = strdup(yytext); return MATCHER_OP; }
+    {int}|[0-9a-zA-Z-]+ {
```
qdeslandes: You need to ensure a unit is present, for clarity.

src/libbpfilter/CMakeLists.txt
```diff
@@ -117,6 +117,7 @@ bf_target_add_elfstubs(libbpfilter
         "pkt_log"
         "flow_hash"
         "sock_addr_log"
+        "ratelimit"
```
qdeslandes: `limit` here too.

src/libbpfilter/include/bpfilter/matcher.h
```diff
@@ -67,6 +67,8 @@ enum bf_matcher_type
     BF_MATCHER_META_FLOW_HASH,
     /// Matches packets based on flow probability (consistent per flow).
     BF_MATCHER_META_FLOW_PROBABILITY,
+    /// Matches a number of packets per unit of time
+    BF_MATCHER_META_RATELIMIT,
```
qdeslandes: `BF_MATCHER_META_LIMIT`

src/libbpfilter/matcher.c
```diff
@@ -408,6 +408,32 @@ static void _bf_print_probability(const void *payload)
...
+
+    int r;
+
+    r = bf_if_index_from_str(raw_payload, (uint32_t *)payload);
+    if (r) {
+        return bf_err_r(
+            r, "\"%s %s\" Blabla error message something something, not '%s'",
+            bf_matcher_type_to_str(type), bf_matcher_op_to_str(op),
+            raw_payload);
+    }
```
qdeslandes: Not sure this is gonna land in prod ;)

src/libbpfilter/matcher.c
```diff
@@ -408,6 +408,32 @@ static void _bf_print_probability(const void *payload)
...
+
+    return 0;
+}
+
+static void _bf_print_ratelimit(const void *payload)
+{
+    assert(payload);
+
+    (void)fprintf(stdout, "%" PRIu32, *(uint32_t *)payload);
+}
```
qdeslandes: Ditto

src/libbpfilter/matcher.c
```diff
@@ -924,6 +950,16 @@ static struct bf_matcher_meta _bf_matcher_metas[_BF_MATCHER_TYPE_MAX] = {
                                    _bf_print_probability),
                 },
         },
+    [BF_MATCHER_META_RATELIMIT] =
+        {
+            .layer = BF_MATCHER_NO_LAYER,
+            // Not sure if we need unsupported_hooks
```
qdeslandes: No, this should be versatile.

src/libbpfilter/matcher.c
```diff
@@ -1473,6 +1509,7 @@ static const char *_bf_matcher_type_strs[] = {
     [BF_MATCHER_META_MARK] = "meta.mark",
     [BF_MATCHER_META_FLOW_HASH] = "meta.flow_hash",
     [BF_MATCHER_META_FLOW_PROBABILITY] = "meta.flow_probability",
+    [BF_MATCHER_META_RATELIMIT] = "meta.ratelimit",
```
qdeslandes: limit

src/libbpfilter/cgen/program.h
```diff
@@ -232,6 +244,12 @@ struct bf_program
 
 #define _free_bf_program_ __attribute__((__cleanup__(bf_program_free)))
 
+struct bf_ratelimit
+{
+    uint64_t current;
+    uint64_t last_time;
+};
+
```
qdeslandes: Define in `runtime.h` and import it instead.

src/libbpfilter/bpf/ratelimit.bpf.c
```diff
@@ -0,0 +1,34 @@
...
+#include <bpf/bpf_helpers.h>
+#include <stddef.h>
+
+#define BF_TIME_S 1000000000
+
+struct bf_ratelimit
+{
+    __u64 current;
+    __u64 last_time;
+};
```
qdeslandes: Define in `runtime.h` and import it instead.

src/libbpfilter/bpf/ratelimit.bpf.c
```diff
@@ -0,0 +1,34 @@
...
+    __u32 key = 0;
+
+    ratelimit = bpf_map_lookup_elem(map, &key);
+    if (!ratelimit) {
+        bpf_printk("failed to fetch the rule's ratelimit");
+        return 1;
+    }
+
+    if (current_time != ratelimit->last_time)
+        ratelimit->current = 0;
```
qdeslandes: Your implementation assume `seconds` as time unit is implicit, but it should be explicit instead, it will ease support for other time units later on.

src/libbpfilter/cgen/prog/map.c
```diff
@@ -194,6 +195,8 @@ int bf_map_new(struct bf_map **map, const char *name, enum bf_map_type type,
         [BF_MAP_TYPE_COUNTERS] = BF_BPF_MAP_TYPE_ARRAY,
         [BF_MAP_TYPE_PRINTER] = BF_BPF_MAP_TYPE_ARRAY,
         [BF_MAP_TYPE_LOG] = BF_BPF_MAP_TYPE_RINGBUF,
+        [BF_MAP_TYPE_RATELIMIT] =
```
qdeslandes: Array: the number of rate-limiting matchers is known during generation and doesn't change.

qdeslandes (changes requested):
Changing the rate limit to a matcher is the good call, it will provide more flexibility.

The implementation has been simplified a lot, so much that so of the functionalities we would deem necessary have been removed.

You will need to think about two things now:
- How do you handle the unit? For now, supporting only `s` (e.g. `meta.limit 10/s`) would be enough, but allowing a unit to be passed will allow us to expand support for rate limiting later on.
- How to handle access to the `bf_ratelimit` (or whatever the name is) structure from a rule? Rules have IDs, but nothing prevents a rule from having multiple rate-limiting matchers. An BPF array map is a good pick, but you need to ensure a `meta.limit` matcher is mapped to the correct `bf_ratelimit` structure in the map.

Also, you'll have to rebase on main :)

src/bfcli/lexer.l
SkohTV: Done :+1:

src/libbpfilter/matcher.c
SkohTV: Looks like someone (me) forgot to change his debug error message

src/libbpfilter/cgen/program.h
SkohTV: Done :+1:

src/libbpfilter/bpf/ratelimit.bpf.c
SkohTV: I took this pattern from [update_counters.bpf.c](https://github.com/facebook/bpfilter/blob/main/src/libbpfilter/bpf/update_counters.bpf.c#L15), perhaps the same thing can be done for it

src/libbpfilter/bpf/limit.bpf.c
```diff
@@ -0,0 +1,42 @@
...
+    case 's':
+        if (current_time_ns > ratelimit->last_time + BF_TIME_S) {
+            ratelimit->current = 0;
+            ratelimit->last_time = current_time_ns;
+        }
+        break;
+    default:
+        bpf_printk("the time unit '%c' is not recognized", letter);
+        return 1;
+    }
```
qdeslandes: This should not be needed. The rate limit (and its unit) passed by the user should be normalized to ns, then we only compare duration and rate limits in ns.

src/libbpfilter/cgen/runtime.h
```diff
@@ -179,5 +179,20 @@ struct bf_runtime
...
+struct bf_ratelimit
+{
+    /** Current timestamp (at the last call of the BPF) */
+    __u64 current;
+
+    /** Last time the rate limit was reset */
+    __u64 last_time;
+};
+
+// I don't think we need a static_assert ?
```
qdeslandes: What for?

src/libbpfilter/matcher.c
```diff
@@ -23,9 +23,11 @@
 #include <limits.h>
 #include <math.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
+#include "asm-generic/errno-base.h"
```
qdeslandes: This header is included in `errno.h`, only `errno.h` should be included. This is a false positive from static analyzers.

src/libbpfilter/matcher.c
```diff
@@ -408,6 +410,40 @@ static void _bf_print_probability(const void *payload)
...
+}
+
+static void _bf_print_limit(const void *payload)
+{
+    assert(payload);
+
+    (void)fprintf(stdout, "%d/%c" PRIu32, *(uint16_t *)payload,
+                  *(uint32_t *)payload >> 16);
+}
+
```
qdeslandes, line 445: The `meta.limit` matcher is composed of a duration and a unit. Whatever the unit is, we can normalize the duration to the corresponding duration in nanoseconds. That way we don't have to store the unit.

When printing the duration, we should convert it back to the correct unit (that is, the most suitable one). We might end up with the user giving us `1000ms` and seeing `1s` when printing it back, but it's not an issue.

src/libbpfilter/bpf/limit.bpf.c
```diff
@@ -0,0 +1,42 @@
...
+
+#include "cgen/runtime.h"
+
+#define BF_TIME_S 1000000000
+
+__u8 bf_ratelimit(void *map, const __u16 limit, __u8 letter)
+{
+    struct bf_ratelimit *ratelimit;
+    __u64 current_time_ns = bpf_ktime_get_ns();
+    __u32 key = 0; // NTODO: should not be 0, but dynamic (WIP)
```
qdeslandes: The key should be the rule index. If the chain changes (leading to a rule index update), the program will be re-generated, so there's no stale IDs.

qdeslandes (changes requested):

src/libbpfilter/cgen/matcher/meta.c
```diff
@@ -157,6 +158,33 @@ _bf_matcher_generate_meta_flow_probability(struct bf_program *program,
...
 }
 
+static int _bf_matcher_generate_meta_limit(struct bf_program *program,
+                                           const struct bf_matcher *matcher)
+{
+    uint64_t tmp = *(uint64_t *)bf_matcher_payload(matcher);
+    uint32_t limit = tmp;
+    uint32_t letter = tmp >> 32;
+
+    static uint32_t key = -1;
```
SkohTV, line 167: I don't love `static`, but it seems to be the cleanest way to do it
We don't really have access to the `rule->index` here

SkohTV (commented):
I've implemented your fixes @qdeslandes 

We don't keep the unit anymore, we keep the duration in `ns`
And each rate limit has a unique id

I could also:
\- Add support for `ms` unit, it shouldn't be a lot more work
\- Add a few tests

src/libbpfilter/bpf/limit.bpf.c
```diff
@@ -0,0 +1,34 @@
...
+#include <stddef.h>
+
+#include "cgen/runtime.h"
+
+__u8 bf_ratelimit(void *map, const __u32 limit, __u32 duration, __u32 key)
+{
+    struct bf_ratelimit *ratelimit;
+    __u64 current_time_ns = bpf_ktime_get_ns();
+
+    bpf_printk("%d, %d - %d", limit, duration, key);
```
qdeslandes, line 19: Remove this debug log, `bpf_printk` is very expensive.

src/libbpfilter/cgen/matcher/meta.c
qdeslandes: Yes, `static` is not the way to go here. A better approach would be to follow `bf_set` example:
- `bf_chain` contains a list of `bf_ratelimit`
- The `meta.limit` matcher references an index in the `bf_chain.ratelimit` field (or whatever the name is)
- During generation, `bf_chain_get_ratelimit(chain, ratelimit_idx)` is used to get the correct rate limit config from the chain

qdeslandes (changes requested):

src/libbpfilter/cgen/matcher/meta.c
SkohTV: So the map (`rmap`) should be swapped with a bf_list (`bf_chain.ratelimit`) ?

qdeslandes: No, both should be used:
- `rmap` is the rate limit data at runtime. Each entry refers to a single rate limit matcher.
- `bf_chain.ratelimit` is a list of rate limit elements defined in the chain, used for generation.
- During generation, the index of the rate limit element in the chain is used so the generated rate limit matcher will look for the correct rate limit runtime data in `rmap`.

qdeslandes (changes requested):

SkohTV:
I think I figured it out
I've kept it as 2 commits for now, to simplify the review, I'll squash them afterward

PR #353, opened 2026-01-09
Fix daemon SEGV when invalid chain type is passed

pzmarzly:
I was playing with bpfilter, and typo-ed a bfcli command. To my surprise, it caused a SIGSEGV crash in daemon process. That's because bfcli parser returned `-22` (EINVAL), which was passed as-is to the daemon, which used it to index into an array, leading to out-of-bounds array access.

I'm introducing 3 changes here:
- `libbpfilter: Add some out-of-bounds checks` - adding some `bf_assert` checks in helper functions that read from arrays, so that in debug builds we get a cleaner signal
- `bfcli: Fix parsing unknown hook types` - instead of passing EINVAL along, print an error to the user and abort
- `daemon: Reject invalid hook types` - some hardening on server side, to turn this crash into an error with easy-to-understand message

### DIFF
--- a/src/libbpfilter/chain.c
+++ b/src/libbpfilter/chain.c
@@ -74,7 +74,14 @@ int bf_chain_new(struct bf_chain **chain, const char *name, enum bf_hook hook,
     int r;
 
     bf_assert(chain && name);
-    bf_assert(policy < _BF_TERMINAL_VERDICT_MAX);
+    if (hook < 0 || hook >= _BF_HOOK_MAX) {
+        bf_warn("invalid hook value %d", hook);
+        return -EINVAL;
+    }
+    if (policy < 0 || policy >= _BF_TERMINAL_VERDICT_MAX) {
+        bf_warn("invalid policy value %d", policy);
+        return -EINVAL;
+    }
 
     _chain = malloc(sizeof(*_chain));
     if (!_chain)
@@ -127,10 +134,14 @@ int bf_chain_new_from_pack(struct bf_chain **chain, bf_rpack_node_t node)
     r = bf_rpack_kv_enum(node, "hook", &hook);
     if (r)
         return bf_rpack_key_err(r, "bf_chain.hook");
+    if (hook < 0 || hook >= _BF_HOOK_MAX)
+        return bf_err_r(-EINVAL, "invalid hook value %d", hook);
 
     r = bf_rpack_kv_enum(node, "policy", &policy);
     if (r)
         return bf_rpack_key_err(r, "bf_chain.policy");
+    if (policy < 0 || policy >= _BF_TERMINAL_VERDICT_MAX)
+        return bf_err_r(-EINVAL, "invalid policy value %d", policy);
 
     r = bf_rpack_kv_array(node, "sets", &array);
     if (r)


src/libbpfilter/hook.c
```diff
@@ -41,6 +41,7 @@ static_assert(ARRAY_SIZE(_bf_hook_strs) == _BF_HOOK_MAX,
 
 const char *bf_hook_to_str(enum bf_hook hook)
 {
+    bf_assert(hook >= 0 && hook < _BF_HOOK_MAX);
```
qdeslandes: `bf_assert()` is phased out in favour of `assert()`. It was used to unit test assertions, it's not needed any more.

qdeslandes: This approach (although it's used elsewhere in the codebase) is not ideal. 

I would suggest something like:
```
if (hook < 0 || hook >= _BF_HOOK_MAX) {
    bf_warn("invalid hook value %d", hook);
    return "(invalid)";
}
```

src/libbpfilter/hook.c
```diff
@@ -74,6 +75,7 @@ enum bf_flavor bf_hook_to_flavor(enum bf_hook hook)
     static_assert(ARRAY_SIZE(flavors) == _BF_HOOK_MAX,
                   "missing entries in bf_flavor array");
 
+    bf_assert(hook >= 0 && hook < _BF_HOOK_MAX);
```
qdeslandes: Similar comment that `const char *bf_hook_to_str(enum bf_hook hook)`. We can check the range, warn, and return `-EINVAL`. This applies to the other functions.

src/bfcli/parser.y
```diff
@@ -182,12 +182,12 @@ verdict         : VERDICT
...
 hook            : HOOK
                 {
-                    enum bf_hook hook = bf_hook_from_str($1);
+                    int hook = bf_hook_from_str($1);
                     if (hook < 0)
                         bf_parse_err("unknown hook '%s'\n", $1);
 
                     free($1);
-                    $$ = hook;
+                    $$ = (enum bf_hook)hook;
```
qdeslandes, line 190: Not necessary. `bf_hook_from_str()` returns a `enum bf_hook` type, we should store it as such. `return -EINVAL` is a special error case which can be dealt with in using `(hook < 0)` even if `hook` is `enum bf_hook` type.

src/libbpfilter/chain.c
```diff
@@ -74,7 +74,8 @@ int bf_chain_new(struct bf_chain **chain, const char *name, enum bf_hook hook,
     int r;
 
     bf_assert(chain && name);
-    bf_assert(policy < _BF_TERMINAL_VERDICT_MAX);
+    bf_assert(hook >= 0 && hook < _BF_HOOK_MAX);
+    bf_assert(policy >= 0 && policy < _BF_TERMINAL_VERDICT_MAX);
```
qdeslandes: Use `if` here, we should only assert on `NULL`.

qdeslandes (changes requested):
About the commits titles:
- `libbpfilter: Add some out-of-bounds checks` -> `lib: add some out-of-bounds checks`
- `bfcli: Fix parsing unknown hook types` -> `cli: fix parsing unknown hook types`
- `daemon: Reject invalid hook types` -> `daemon: reject invalid hook types`

Most of those comments are due to stuff that is not properly formalized and described in the doc, mostly because is evolved over time (hence, not applied to the whole codebase).

src/bfcli/parser.y
pzmarzly: Sadly, seems necessary. Storing negative values in `enum` is UB, and GCC optimizes that `if` away, even in -O0.

Godbolt: https://godbolt.org/z/MvWzob36e
Meta internal explanation: M8N33B

qdeslandes: Oh no :( 

Alright, let's find a different approach then. Another approach could be:
`int bf_hook_from_str(const char *str, enum bf_hook *ret)`. This way we have a return value to check, and we can use `ret` on success. It's a bit more verbose, but at least our back's covered.

The other option would be to use `int bf_hook_from_str(const char *str)` and cast the return to `enum bf_hook` after it's checked. I find the suggestion above better as there's clear distinction between the return value and the parsed value. What do you think?

Oooooor, we can return the sentinel value on error: `_BF_HOOK_MAX` if the string can't be parsed.

pzmarzly: > `int bf_hook_from_str(const char *str, enum bf_hook *ret)`

I like this approach the most. Now I'm thinking about redoing these commits from scratch, and adding validation only to the two points where we accept user input:
- for CLI, change `bf_hook_from_str` like you described
- for daemon, change `bf_rpack_kv_enum(node, key, value)` to `bf_rpack_kv_enum(node, key, value, min, max)` and validate the input value when parsing the request instead of later

Seems other parsing, for example `bf_matcher_type_from_str`, already uses that approach.

PR #354, opened 2026-01-09
docs: Update getting started guide

pzmarzly:
The command does not work - the syntax it uses doesn't match what bpfilter expects, probably due to format changing over time.

Also, for some reason NF_LOCAL_OUT isn't filtering packets on my laptop (Fedora 43)? Maybe because nftables is also in use? What do you think about showing a XDP example instead?

### DIFF
--- a/doc/usage/index.rst
+++ b/doc/usage/index.rst
@@ -76,13 +76,14 @@ Now that the daemon is up and running, we will use ``bfcli`` to send a filtering
 .. code-block:: bash
 
 	$ sudo bfcli ruleset set --from-str "
-	chain BF_HOOK_NF_LOCAL_OUT policy ACCEPT
-	    rule
-	        ip4.proto icmp
-	        DROP
+        chain my_chain BF_HOOK_XDP{ifindex=$IFINDEX} ACCEPT
+            rule
+                ip4.proto eq icmp
+                counter
+                DROP
 	"
 
-We split the chain over multiple lines, to it's easier to read. Alternatively, you can write the chain in a file and call ``bfcli ruleset set --from-file $MYFILE``. We choose to create a chain attached to ``BF_HOOK_NF_LOCAL_OUT`` which is called for every packet leaving this host with ``ACCEPT`` as the default policy: if a packet doesn't match any of the rules defined, it will be accepted by default.
+We split the chain over multiple lines, so it's easier to read. Alternatively, you can write the chain in a file and call ``bfcli ruleset set --from-file $MYFILE``. We choose to create a chain attached to ``BF_HOOK_XDP`` with ``ACCEPT`` as the default policy: if a packet doesn't match any of the rules defined, it will be accepted by default.
 
 Our chain contains a single rule matching against the IPv4's ``protocol`` field. Packets matching this rule will be ``DROP`` ed.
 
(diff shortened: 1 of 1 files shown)


doc/usage/index.rst
```diff
@@ -75,16 +75,21 @@ Now that the daemon is up and running, we will use ``bfcli`` to send a filtering
...
 .. code-block:: bash
 
+    $ default_iface=$(ip --json route | jq -r '.[] | select(.dst == "default") | .dev')
+    $ default_iface_index=$(cat /sys/class/net/$default_iface/ifindex)
 	$ sudo bfcli ruleset set --from-str "
-	chain BF_HOOK_NF_LOCAL_OUT policy ACCEPT
-	    rule
-	        ip4.proto icmp
-	        DROP
+        chain my_chain BF_HOOK_XDP{ifindex=$default_iface_index} ACCEPT
```
qdeslandes: `{ifindex=$default_iface_index}` -> `{ifindex=$IFINDEX}`, we expect the user to be able to find the iface it wants to attach the chain to by itself. Let's keep the example simple.

doc/usage/index.rst
```diff
@@ -75,16 +75,21 @@ Now that the daemon is up and running, we will use ``bfcli`` to send a filtering
...
+            rule
+                ip4.proto eq icmp
+                counter
+                DROP
+            rule
+                ip6.nexthdr eq icmpv6
+                counter
+                DROP
```
qdeslandes: No need for IPv6, the example pings in IPv4. Let's keep a short example just for the demo.

doc/usage/index.rst
```diff
@@ -75,16 +75,21 @@ Now that the daemon is up and running, we will use ``bfcli`` to send a filtering
...
 	"
 
-We split the chain over multiple lines, to it's easier to read. Alternatively, you can write the chain in a file and call ``bfcli ruleset set --from-file $MYFILE``. We choose to create a chain attached to ``BF_HOOK_NF_LOCAL_OUT`` which is called for every packet leaving this host with ``ACCEPT`` as the default policy: if a packet doesn't match any of the rules defined, it will be accepted by default.
-
-Our chain contains a single rule matching against the IPv4's ``protocol`` field. Packets matching this rule will be ``DROP`` ed.
+We split the chain over multiple lines, to it's easier to read. Alternatively, you can write the chain in a file and call ``bfcli ruleset set --from-file $MYFILE``. We choose to create a chain attached to ``BF_HOOK_XDP`` with ``ACCEPT`` as the default policy: if a packet doesn't match any of the rules defined, it will be accepted by default.
```
qdeslandes: Keep the line explaining what the chain does.

qdeslandes: "to it's easier to read" -> "so it's easier to read"

doc/usage/index.rst
pzmarzly: Not if you don't have any IPv4 address. For example, on my Hetzner VPS:

```
[root@ipv6-only-test1 ~]# ping facebook.com
PING facebook.com (2a03:2880:f177:185:face:b00c:0:25de) 56 data bytes
64 bytes from edge-star-mini6-shv-02-fra3.facebook.com
(2a03:2880:f177:185:face:b00c:0:25de): icmp_seq=1 ttl=55 time=6.26 ms
```

That's because ping uses `AI_ADDRCONFIG` flag to filter out unsupported transports from DNS results.

qdeslandes: Of course, a basic chain wouldn't be applicable for every use case, but the purpose here is to keep it very (very) simple, just so users have an idea of the structure.

We should keep a single rule, for clarity purposes.

qdeslandes (changes requested):

doc/usage/index.rst
pzmarzly: OK

PR #366, opened 2026-01-14
bfcli: fix parsing log flags

pzmarzly:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #368
* #367
* __->__ #366

ReviewStack: https://reviewstack.dev/facebook/bpfilter/pull/366

---

`ctest --test-dir build -R e2e.daemon.netns_to_host`  was flaky, which was annoying me, as it was failing on unrelated changes.

With some Claude-assisted debugging, I found out it's because `uint8_t log` sometimes contains garbage data, that we later OR into in `log |= BF_FLAG(header);` line. When that happens, the printed output doesn't match what `awk` is expecting.

### DIFF
--- a/src/bfcli/parser.y
+++ b/src/bfcli/parser.y
@@ -287,7 +287,7 @@ rule_option     : LOG LOG_HEADERS
                     char *tmp = in;
                     char *saveptr;
                     char *token;
-                    uint8_t log;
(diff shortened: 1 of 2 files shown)


qdeslandes (approved):
LGTM, good catch! Thanks!

PR #368, opened 2026-01-14
daemon: validate enum values coming from user

pzmarzly:
Right now a request with an invalid enum value will most likely cause a SIGSEGV on daemon side. Let's protect ourselves a bit, and reject such requests.

See https://github.com/facebook/bpfilter/pull/353 for example of how easy it was to crash the daemon.

### DIFF
--- a/src/libbpfilter/chain.c
+++ b/src/libbpfilter/chain.c
@@ -73,8 +73,11 @@ int bf_chain_new(struct bf_chain **chain, const char *name, enum bf_hook hook,
     size_t ridx = 0;
     int r;
 
-    bf_assert(chain && name);
-    bf_assert(policy < _BF_TERMINAL_VERDICT_MAX);
+    assert(chain && name);
+    if (hook >= _BF_HOOK_MAX)
+        return bf_err_r(-EINVAL, "unknown hook type");
+    if (policy >= _BF_TERMINAL_VERDICT_MAX)
+        return bf_err_r(-EINVAL, "unknown policy type");
 
     _chain = malloc(sizeof(*_chain));
     if (!_chain)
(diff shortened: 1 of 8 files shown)


src/libbpfilter/chain.c
```diff
@@ -75,6 +75,7 @@ int bf_chain_new(struct bf_chain **chain, const char *name, enum bf_hook hook,
 
     bf_assert(chain && name);
     bf_assert(policy < _BF_TERMINAL_VERDICT_MAX);
+    bf_assert(hook < _BF_HOOK_MAX);
```
qdeslandes: `assert()`

tests/unit/libbpfilter/pack.c
```diff
@@ -399,7 +399,7 @@ static void wpack_enum(void **state)
     assert_ok(bf_rpack_new(&rpack, data, data_len));
     root = bf_rpack_root(rpack);
 
-    assert_ok(bf_rpack_kv_enum(root, "enum_val", &enum_val));
+    assert_ok(bf_rpack_kv_enum(root, "enum_val", &enum_val, 0, 10));
     assert_int_equal(enum_val, 5);
```
qdeslandes, line 405: Add test for out-of-range values.

qdeslandes (changes requested):

src/libbpfilter/chain.c
```diff
@@ -73,8 +73,9 @@ int bf_chain_new(struct bf_chain **chain, const char *name, enum bf_hook hook,
     size_t ridx = 0;
     int r;
 
-    bf_assert(chain && name);
-    bf_assert(policy < _BF_TERMINAL_VERDICT_MAX);
+    assert(chain && name);
+    assert(policy < _BF_TERMINAL_VERDICT_MAX);
+    assert(hook < _BF_HOOK_MAX);
```
qdeslandes: Use `if (...)` here, so we only use `assert()` to prevent NULL pointers.

src/libbpfilter/include/bpfilter/pack.h
```diff
@@ -304,19 +304,29 @@ typedef union
...
  *
  * @param node Node to read from.
  * @param key Key of the node to read. Can't be NULL.
  * @param value Pointer to the enumerator value to read into.
+ * @param first First valid enumerator value (inclusive).
+ * @param max Maximum enumerator value (exclusive).
  * @return 0 on success, or a negative error value on failure.
  */
-#define bf_rpack_kv_enum(node, key, value)                                     \
+#define bf_rpack_kv_enum(node, key, value, first, max)                         \
```
qdeslandes, line 317: Please update `bf_rpack_enum` too :)

qdeslandes (approved):

qdeslandes:
Good for me, thanks!

Issue #358, opened 2026-01-12
Log a rule's action


qdeslandes:
When logging, bpfilter should store the rule's action as part of the logged data. While it is currently possible to find out which action was taken (by fetching the ruleset and checking the rule based on the log's rule ID), it would be more convenient to store it directly in the log.

Plan:
- Add a new field to bf_log and [update the `log.bpf.c` stub](https://github.com/facebook/bpfilter/blob/main/src/bpfilter/bpf/log.bpf.c) to set the rule's action
  - The rule's ID is not accessible through the stub (yet), it could be an extra parameter (though `log()` currently expect 5 arguments)
- Update bfcli to print the rule's action

PR #387, opened 2026-01-21
lib,cli,daemon: log the matching rule's verdict

jordalgo:
This is so users don't have to use the rule id to then look up the
action/verdict that was taken.

Example output:
```
[13:20:01.176233] Rule #0 matched 142 bytes with verdict CONTINUE
  TCP       : 35438 → 10057 [psh,ack]
              seq=1825888209 ack=1600908100 win=863
```
```
[13:37:50.908611] Rule #0 matched 86 bytes with verdict ACCEPT
  TCP       : 61452 → 10057 [ack]
              seq=2463015047 ack=813395332  win=406
```

Issue: https://github.com/facebook/bpfilter/issues/358

### DIFF
--- a/src/bpfilter/cgen/elfstub.h
+++ b/src/bpfilter/cgen/elfstub.h
@@ -122,14 +122,14 @@ enum bf_elfstub_id
     /**
      * Log user-requested packet headers to a ring buffer.
      *
-     * `__u8 bf_log(struct bf_runtime *ctx, __u32 rule_id, __u8 headers, __u16 l3_proto, __u8 l4_proto)`
+     * `__u8 bf_log(struct bf_runtime *ctx, __u32 rule_id, __u8 headers, __u32 verdict, __u32 l3_l4_proto)`
      *
      * **Parameters**
      * - `ctx`: address of the `bf_runtime` context of the program.
-     * - `rule_id`: id of the matched rule
+     * - `rule_id`: id of the matched rule.
      * - `headers`: user-requested headers to log.
-     * - `l3_proto`: layer 3 (internet) protocol identifier.
-     * - `l4_proto`: layer 4 (transport) protocol identifier.
+     * - `verdict`: verdict of the matched rule.
+     * - `l3_l4_proto`: layer 3 (internet) and layer 4 (transport) protocols packed (l3 << 16 | l4)
      *
      * **Return** 0 on success, or 1 on error.
      */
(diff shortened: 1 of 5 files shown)


jordalgo:
Need advice on adding tests for this.

src/libbpfilter/include/bpfilter/runtime.h
```diff
@@ -97,6 +99,9 @@ struct bf_log
     /** ID of the rule triggering the log. */
     __u32 rule_id;
 
+    /** Verdict of the rule triggering the log. */
+    enum bf_verdict verdict;
```
qdeslandes: Use `__32` for the verdict, we don't want to pull-in extra headers in here as this will eventually cause trouble with the ElfStubs compilation.

src/bpfilter/cgen/elfstub.h
```diff
@@ -122,14 +122,14 @@ enum bf_elfstub_id
...
      * **Parameters**
      * - `ctx`: address of the `bf_runtime` context of the program.
-     * - `map`: pointer to the logs ring buffer.
-     * - `layers`: user-requested headers to log.
-     * - `l3_proto`: layer 3 (internet) protocol identifier.
-     * - `l4_proto`: layer 4 (transport) protocol identifier.
+     * - `rule_id`: id of the matched rule.
+     * - `headers`: user-requested headers to log.
+     * - `verdict`: verdict of the matched rule.
+     * - `l3_l4_proto`: layer 3 (internet) and layer 4 (transport) protocols packed.
```
qdeslandes: Describe the bytes layout `(l3 << 16 | l4)`.

qdeslandes (changes requested):
Please fix your commit titles (see https://bpfilter.io/developers/style.html), and add a brief description to the commit modifying the `bf_log` stub arguments (for future reference).

> Need advice on adding tests for this.

The `log` action is not automatically testing for now. Feel free to add an e2e test script to create a chain and parse the output of bfcli. Realistically, if this was my PR, I would manually test a few cases (including edge cases to validate byte order), and paste the results in the PR, that would be enough.

src/libbpfilter/include/bpfilter/runtime.h
jordalgo: Done.

src/bpfilter/cgen/elfstub.h
jordalgo: Done.

PR #398, opened 2026-01-24
cli: use -h for help

pzmarzly:
argp has default behavior of using `-?`/`--help` for help, `-V`/`--version` for version, and `--usage` for short help. This is not customizable, but can be opted out of with `ARGP_NO_HELP`. Then you need to handle these 3 flags yourself, which this PR does.

Now `-h` / `--help` is for help, `-V` / `--version` is for version, and `--usage` for short help. I would love to remove `--usage` as well, but argp error messages say `Try `bfcli --help' or `bfcli --usage' for more information.` and this is non-customizable.

Wdyt? Is behaving like a "standard" CLI worth this extra code?

### DIFF
--- a/src/bfcli/opts.c
+++ b/src/bfcli/opts.c
@@ -32,6 +32,16 @@
 #define BFC_HELP_ENTRY(action, help)                                           \
     {.name = _bfc_action_strs[action], .doc = (help), .flags = OPTION_DOC}
 
+/**
+ * @brief Get key value for long-only flag.
+ *
+ * By passing non-printable character to argp, we can create
+ * long-only flag. This macro shifts the IDs into non-printable space.
+ *
+ * @param id Unique ID number.
+ */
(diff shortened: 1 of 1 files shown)


src/bfcli/opts.c
```diff
@@ -312,6 +371,30 @@ struct bfc_opts_opt
     void (*parser)(struct argp_state *state, const char *arg,
                    struct bfc_opts *opts);
 } _bfc_options[] = {
+    {
+        .id = BFC_OPT_HELP,
+        .key = 'h',
+        .name = "help",
+        .arg = NULL,
```
qdeslandes: `.arg` can be omitted as it's NULL.

src/bfcli/opts.c
qdeslandes, line 617: While testing, I noticed `_BFC_NAME_LEN` being insufficient if we use the absolute path to `bfcli`, could you update it to use `(PATH_MAX + 32)` instead? It should give use enough room for the path, the object, and the action keywords.

Issue #386, opened 2026-01-21
Handle empty sets in a chain


qdeslandes:
bpfilter doesn't allow a set in a ruleset to be empty:
```
$ sudo ./build/output/sbin/bfcli ruleset set --from-str "
chain test BF_HOOK_XDP{ifindex=2} ACCEPT
    rule
        (ip4.saddr) in {}
        ACCEPT
"
error  : failed to set ruleset: Invalid argument
```

Instead of failing, bpfilter should ignore those sets and remove them from the rule(s). If the rule is empty (no matcher), then it should be discarded. A warning message should be printed to inform the user, as the matcher won't appear in `bfcli ruleset get` anymore.

PR #400, opened 2026-01-26
lib: chain: disable codegen for rules referencing empty sets

yaakov-stein:
Adding support for empty sets based on #386. Previously, having any matcher with an empty set caused an error and the entire ruleset was rejected. Now, we treat the matcher with an empty set as a no-op (always false) and skip codegen (both flags and rule generation) for any rule that contains a matcher with an empty set . We alert the user to this by sending a warning for each rule that will be ignored.

Test Plan:
- Unit tests for disabled flag in `bf_chain_new`
- e2e tests creating rulesets with empty sets
- Manual testing (checking that the bytecode is the same both without the rule that contains the empty set and without that rule)


src/libbpfilter/chain.c
```diff
@@ -19,11 +19,51 @@
...
+                                         const struct bf_rule *rule)
+{
+    bf_list_foreach (&rule->matchers, matcher_node) {
+        struct bf_matcher *matcher = bf_list_node_get_data(matcher_node);
+
+        if (bf_matcher_get_type(matcher) != BF_MATCHER_SET)
+            continue;
+
+        uint32_t set_id = *(uint32_t *)bf_matcher_payload(matcher);
+        struct bf_set *set = bf_list_get_at(&chain->sets, set_id);
```
yaakov-stein: Currently, this iterates through sets for each rule (since a set is stored as a list and we can't look it up in O(1)). We could create a set mapping (0->valid, -1->invalid) in `bf_chain_new `and pass it through to enable this to be O(1), but I opted against that to keep the code a bit simpler (happy to switch that though).

src/libbpfilter/include/bpfilter/rule.h
```diff
@@ -64,6 +64,10 @@ struct bf_rule
     uint64_t mark;
 
     bool counters;
+
+    /** If true, skip this rule during flag calculation and code generation. */
+    bool disabled;
```
yaakov-stein, line 69: Note: We're not serializing or deserializing this (similar to flags in `bf_chain`) as it will be re-calculated after deserialization.

src/bpfilter/cgen/program.c
```diff
@@ -133,6 +133,15 @@ int bf_program_new(struct bf_program **program, const struct bf_chain *chain)
         struct bf_set *set = bf_list_node_get_data(set_node);
         _free_bf_map_ struct bf_map *map = NULL;
 
+        // Skip map creation for empty sets - add NULL placeholder
+        if (bf_list_is_empty(&set->elems)) {
```
pzmarzly: Maybe add a helper to set.h called `bf_set_is_empty`? The fact that sets use lists under the hood could be considered private, and IIRC we want to change it one day to use true hashsets instead

src/bpfilter/cgen/program.c
```diff
@@ -965,6 +986,14 @@ static int _bf_program_load_sets_maps(struct bf_program *new_prog)
...
         struct bf_set *set = bf_list_node_get_data(set_node);
         struct bf_map *map = bf_list_node_get_data(map_node);
+
+        // Skip null maps (empty sets)
+        if (!map) {
+            set_node = bf_list_node_next(set_node);
+            map_node = bf_list_node_next(map_node);
+            continue;
+        }
+
```
pzmarzly: If I understand correctly, this iteration is the only reason why you need to store NULLs in `prog->sets`. If you chose to skip storing anything for empty sets, the iteration could look like this:

```suggestion
        struct bf_set *set = bf_list_node_get_data(set_node);

        // Skip empty sets
        if (bf_set_is_empty(set) {
            set_node = bf_list_node_next(set_node);
            continue;
        }
        
        struct bf_map *map = bf_list_node_get_data(map_node);
```

src/libbpfilter/chain.c
pzmarzly: Minor nit: elsewhere we call it `set_index`:

```
        case BF_FIXUP_TYPE_SET_MAP_FD:
            map = bf_list_get_at(&program->sets, fixup->attr.set_index);
            if (!map) {
                return bf_err_r(-ENOENT, "can't find set map at index %lu",
                                fixup->attr.set_index);
            }
```

pzmarzly (commented):
LGTM, minor nits.

Can you update this comment in `program.h`:
```
    /// List of set maps
    bf_list sets;
```
to say that this list can contain NULLs if bf_set was empty, and that indexes in this list match indexes in the bf_chain's `sets`?

src/bpfilter/cgen/program.c
qdeslandes: This can be simplified as:
```
if (!bf_set_is_empty(set)) {
    (void)snprintf(name, BPF_OBJ_NAME_LEN, "set_%04x", (uint8_t)set_idx);
    r = bf_map_new_from_set(&map, name, set);
    if (r < 0)
        return r;
}

r = bf_list_push(&_program->sets, &map);
if (r < 0)
    return r;

set_idx++;
```

Two advantages:
- Still a single "valid path", we only stop the loop's flow on error. If everything works as expected, the flow is still the same.
- Use `bf_list_push()` instead of `bf_list_add_tail()` so we don't have to call `TAKE_PTR()`.

qdeslandes:
```
_free_bf_map_ struct bf_map *map = NULL;

if (!bf_rpack_is_nil(array_node))
    r = bf_list_empty(&_program->sets, bf_map_new_from_pack, map, dir_fd, array_node);
else
    r = bf_list_add_tail(&_program->sets, NULL);

if (r)
    return bf_err_r(r, "failed to unpack bf_map into bf_program.sets");
```

This should be simpler :) Also, variables should always be defined at the top of the block (so `bf_map *map` before `if (bf_rpack_is_nil())`).

src/bpfilter/cgen/program.c
```diff
@@ -1158,7 +1193,10 @@ int bf_program_pin(struct bf_program *prog, int dir_fd)
     }
 
     bf_list_foreach (&prog->sets, set_node) {
-        r = bf_map_pin(bf_list_node_get_data(set_node), dir_fd);
+        struct bf_map *map = bf_list_node_get_data(set_node);
+        if (!map)
+            continue;
```
qdeslandes, line 1202: Empty line below

src/libbpfilter/include/bpfilter/set.h
```diff
@@ -118,4 +118,12 @@ int bf_set_pack(const struct bf_set *set, bf_wpack_t *pack);
 
 void bf_set_dump(const struct bf_set *set, prefix_t *prefix);
 
+/**
+ * @brief Check if a set is empty.
+ *
+ * @param set Initialised set. Can't be NULL.
+ * @return true if the set has no elements, false otherwise.
```
qdeslandes: `true` -> `True`.

src/libbpfilter/chain.c
```diff
@@ -19,11 +19,51 @@
 #include "bpfilter/set.h"
 #include "bpfilter/verdict.h"
 
-int _bf_chain_check_rule(struct bf_chain *chain, const struct bf_rule *rule)
+/**
+ * Check if a rule references an empty set.
```
qdeslandes: `@brief`

src/libbpfilter/chain.c
```diff
@@ -19,11 +19,51 @@
...
+ *
+ * @param chain Chain containing the sets list.
+ * @param rule Rule to check.
+ * @return 0 if no issues, 1 if rule references an empty set (should be
+ *         disabled), or negative errno if rule references a non-existent set.
+ */
+static int _bf_rule_references_empty_set(const struct bf_chain *chain,
+                                         const struct bf_rule *rule)
+{
+    bf_list_foreach (&rule->matchers, matcher_node) {
```
qdeslandes, line 36: Assert on `chain` and `rule` as those must be not NULL.

src/libbpfilter/chain.c
```diff
@@ -19,11 +19,51 @@
...
+        struct bf_matcher *matcher = bf_list_node_get_data(matcher_node);
+
+        if (bf_matcher_get_type(matcher) != BF_MATCHER_SET)
+            continue;
+
+        uint32_t set_index = *(uint32_t *)bf_matcher_payload(matcher);
+        struct bf_set *set = bf_list_get_at(&chain->sets, set_index);
```
qdeslandes: Variables should be defined at the top of the scope.

src/libbpfilter/chain.c
```diff
@@ -19,11 +19,51 @@
...
+
+        if (!set)
+            return bf_err_r(-EINVAL, "rule %u references non-existent set",
```
qdeslandes, line 48: This spans over multiple lines, so it should be around braces `{}`.

src/libbpfilter/chain.c
```diff
@@ -19,11 +19,51 @@
...
+                            rule->index);
+
+        if (bf_set_is_empty(set)) {
+            bf_warn("rule %u references empty set, rule will be skipped",
```
qdeslandes: "disabled" instead of "skipped"

src/libbpfilter/chain.c
```diff
@@ -19,11 +19,51 @@
...
 {
+    int r;
+
     assert(rule);
 
-    if (rule->log)
+    r = _bf_rule_references_empty_set(chain, rule);
+    if (r < 0)
+        return r;
+    rule->disabled = (r > 0);
```
qdeslandes: `rule->disabled = r`?

tests/e2e/matchers/set.sh
```diff
@@ -56,6 +56,18 @@ bfcli ruleset set --dry-run --from-str "chain xdp BF_HOOK_XDP ACCEPT rule (  icm
...
+bfcli ruleset set --dry-run --from-str "chain xdp BF_HOOK_XDP ACCEPT rule (ip4.saddr) in {} DROP" 2>&1 | grep -q "rule 0 references empty set"
+bfcli ruleset set --dry-run --from-str "chain xdp BF_HOOK_XDP ACCEPT rule (ip4.saddr, ip4.proto) in {} DROP" 2>&1 | grep -q "rule 0 references empty set"
+bfcli ruleset set --dry-run --from-str "chain xdp BF_HOOK_XDP ACCEPT
+    set myset (ip4.saddr) in {}
+    rule (ip4.saddr) in myset DROP
+" 2>&1 | grep -q "rule 0 references empty set"
+bfcli ruleset set --dry-run --from-str "chain xdp BF_HOOK_XDP ACCEPT
+    rule (ip4.saddr) in { 192.168.1.1 } DROP
+    rule (ip4.saddr) in {} ACCEPT
+" 2>&1 | grep -q "rule 1 references empty set"
```
qdeslandes: Greping for an error/warning message is not great as we might change it for any reason. Instead, use `bpftool` to check if a map is created.

tests/unit/libbpfilter/chain.c
```diff
@@ -100,13 +100,182 @@ static void get_set_from_matcher(void **state)
...
+        assert_ok(bf_rule_add_matcher(rule, BF_MATCHER_SET, BF_MATCHER_IN,
+                                      &set_index, sizeof(set_index)));
+        assert_ok(bf_list_add_tail(&rules, rule));
+        TAKE_PTR(rule);
+    }
+
+    assert_ok(bf_chain_new(&chain, "test", BF_HOOK_TC_EGRESS,
+                           BF_VERDICT_ACCEPT, &sets, &rules));
+
+    struct bf_rule *rule = bf_list_node_get_data(bf_list_get_head(&chain->rules));
```
qdeslandes: Define `struct bf_rule *rule` at the very beginning, then `bf_rule_new(new)`, then `bf_list_add_tail(list, rule)`, and you can reuse the pointer to `rule` to ensure it's disabled.

tests/unit/libbpfilter/chain.c
```diff
@@ -100,13 +100,182 @@ static void get_set_from_matcher(void **state)
...
```
qdeslandes: Same as above.

tests/unit/libbpfilter/chain.c
```diff
@@ -100,13 +100,182 @@ static void get_set_from_matcher(void **state)
...
+    assert_ok(bf_chain_new(&chain, "test", BF_HOOK_TC_EGRESS,
+                           BF_VERDICT_ACCEPT, &sets, &rules));
+
+    struct bf_rule *r0 = bf_list_node_get_data(bf_list_get_head(&chain->rules));
+    struct bf_rule *r1 = bf_list_get_at(&chain->rules, 1);
+    assert_true(r0->disabled);
+    assert_false(r1->disabled);
+
+    // BF_CHAIN_LOG should be set (from enabled rule 1)
+    assert_int_not_equal(chain->flags & BF_FLAG(BF_CHAIN_LOG), 0);
```
qdeslandes: Multiple test cases defined above can be summarized into this test: define 2 rules, one with an empty set, one without, both with logging and/or `ipv6.nexthdr` matching, then add them to a chain and validate the outcome.

src/libbpfilter/chain.c
qdeslandes: LGTM for now. Eventually, we'll need set and hash maps.

Issue #362, opened 2026-01-12
Use `SKB` or `HW` XDP driver depending on device capabilities


qdeslandes:
When an XDP program is attached to the kernel, the caller can pick the attachment mode:
- [`XDP_FLAGS_HW_MODE`](https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_XDP/#xdp_flags_hw_mode): offload the program to the NIC. Only select Netronome devices support offloading.
- [`XDP_FLAGS_DRV_MODE`](https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_XDP/#xdp_flags_drv_mode): attach the program in driver mode, requires the NIC to support it.
- [`XDP_FLAGS_SKB_MODE`](https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_XDP/#xdp_flags_skb_mode): generic mode, works regardless of driver support. This negates the speed advantages offered by XDP, as it is called further up the stack, once a socket buffer has been created.

bpfilter will always pick [`XDP_FLAGS_SKB_MODE`](https://github.com/facebook/bpfilter/blob/main/src/bpfilter/cgen/prog/link.c#L164) as it will always be supported, but that leaves a lot of performance on the table. Instead, bpfilter should probe the NIC to use the most appropriate mode.

qdeslandes:
Fixed by #401

PR #401, opened 2026-01-26
daemon: cgen: link: probe for best XDP attachment mode

yaakov-stein:
As mentioned in #362, instead of always using XDP_FLAGS_SKB_MODE, try hardware offload first, then driver mode, falling back to SKB mode only if neither is available. This improves performance on NICs that support XDP driver or hardware offloading.

(Note: I searched around for a way to do this without trial and error - HW->DRV->SKB -, but was unable to find a way to determine what XDP options are available dynamically. If anyone knows of a way to do that, please let me know.)

Test Plan:
- Manual Testing: Was able to test that on a server that XDP_FLAGS_DRV_MODE properly attaches in driver mode, and that XDP_FLAGS_SKB_MODE properly functions as the fallback. I was unable to get access to a Netronome NIC to test the HW offloading.

### DIFF
--- a/src/bpfilter/cgen/prog/link.c
+++ b/src/bpfilter/cgen/prog/link.c
@@ -146,6 +146,38 @@ void bf_link_dump(const struct bf_link *link, prefix_t *prefix)
     bf_dump_prefix_pop(prefix);
 }
 
+/**
+ * @brief Try to attach an XDP program, probing for the best available mode.
+ *
+ * Tries driver mode first, falling back to SKB mode.
+ *
+ * @param prog_fd File descriptor of the BPF program to attach.
+ * @param ifindex Network interface index to attach the program to.
+ * @param hook Hook to attach the program to.
+ * @return File descriptor of the created link on success, negative errno on failure.
+ */
+static int _bf_link_try_attach_xdp(int prog_fd, int ifindex, enum bf_hook hook)
+{
+    int r;
+
+    r = bf_bpf_link_create(prog_fd, ifindex, hook, XDP_FLAGS_DRV_MODE, 0, 0);
+    if (r >= 0) {
+        bf_info("attached XDP program in driver mode");
+        return r;
+    }
+
+    if (r != -ENOTSUP)
+        return r;
+
+    bf_dbg_r(r, "driver mode not available");
+
+    r = bf_bpf_link_create(prog_fd, ifindex, hook, XDP_FLAGS_SKB_MODE, 0, 0);
+    if (r >= 0)
+        bf_info("attached XDP program in SKB mode");
+
+    return r;
+}
+
 int bf_link_attach(struct bf_link *link, enum bf_hook hook,
                    struct bf_hookopts **hookopts, int prog_fd)
 {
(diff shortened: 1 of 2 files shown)


src/bpfilter/cgen/prog/link.c
```diff
@@ -160,8 +195,7 @@ int bf_link_attach(struct bf_link *link, enum bf_hook hook,
 
     switch (bf_hook_to_flavor(hook)) {
     case BF_FLAVOR_XDP:
-        r = bf_bpf_link_create(prog_fd, _hookopts->ifindex, hook,
-                               XDP_FLAGS_SKB_MODE, 0, 0);
+        r = _bf_link_attach_xdp(prog_fd, _hookopts->ifindex, hook);
```
qdeslandes: Rename to `_bf_link_try_attach_xdp` which is more explicit.

src/bpfilter/cgen/prog/link.c
```diff
@@ -146,6 +146,41 @@ void bf_link_dump(const struct bf_link *link, prefix_t *prefix)
...
+static int _bf_link_attach_xdp(int prog_fd, int ifindex, enum bf_hook hook)
+{
+    int r;
+
+    r = bf_bpf_link_create(prog_fd, ifindex, hook, XDP_FLAGS_HW_MODE, 0, 0);
+    if (r >= 0) {
+        bf_info("attached XDP program in hardware offload mode");
+        return r;
+    }
+    bf_dbg("hardware offload mode not available: %s", strerror(-r));
```
qdeslandes: Use `bf_dbg_r(r, fmt, ...)` to print message with an error code.

src/bpfilter/cgen/prog/link.c
```diff
@@ -146,6 +146,41 @@ void bf_link_dump(const struct bf_link *link, prefix_t *prefix)
...
+ * @param ifindex Network interface index to attach the program to.
+ * @param hook Hook to attach the program to.
+ * @return File descriptor of the created link on success, negative errno on failure.
+ */
+static int _bf_link_attach_xdp(int prog_fd, int ifindex, enum bf_hook hook)
+{
+    int r;
+
+    r = bf_bpf_link_create(prog_fd, ifindex, hook, XDP_FLAGS_HW_MODE, 0, 0);
+    if (r >= 0) {
```
qdeslandes: Check for `-ENOTSUP` explicitly, so we know this mode is not supported. Any other error should stop this function and return an error.

src/bpfilter/cgen/prog/link.c
```diff
@@ -146,6 +146,41 @@ void bf_link_dump(const struct bf_link *link, prefix_t *prefix)
...
+ * @param prog_fd File descriptor of the BPF program to attach.
+ * @param ifindex Network interface index to attach the program to.
+ * @param hook Hook to attach the program to.
+ * @return File descriptor of the created link on success, negative errno on failure.
+ */
+static int _bf_link_attach_xdp(int prog_fd, int ifindex, enum bf_hook hook)
+{
+    int r;
+
+    r = bf_bpf_link_create(prog_fd, ifindex, hook, XDP_FLAGS_HW_MODE, 0, 0);
```
qdeslandes: This can be removed as we don't have the hardware to validate it. There might be quirks to apply to offloaded programs that we are not aware of.

src/bpfilter/cgen/prog/link.c
yaakov-stein: Added this in, but just wanted to flag in case we hear of issues here - since we previously always tried SKB mode, if for whatever reason a driver returns an inconsistent error code, this code would cause the XDP attach to fail, even if SKB mode would work. I don't think that should ever happen, but figured I'd note it here just in case.

PR #402, opened 2026-01-27
lib: bpf: fix return value documentation for fd-returning functions

yaakov-stein:
Non-functional changes: `bf_bpf_btf_load`, `bf_bpf_map_create`, and `bf_bpf_link_create` return file descriptors on success. Update the documentation to match the behavior (and to match `bf_bpf_token_create`).

### DIFF
--- a/src/libbpfilter/include/bpfilter/bpf.h
+++ b/src/libbpfilter/include/bpfilter/bpf.h
@@ -102,8 +102,8 @@ int bf_bpf_prog_run(int prog_fd, const void *pkt, size_t pkt_len,
  *
  * @param bpffs_fd File descriptor of the BPF filesystem to create the token
  *        for.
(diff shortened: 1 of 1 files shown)


src/libbpfilter/include/bpfilter/bpf.h
```diff
@@ -113,7 +113,8 @@ int bf_bpf_token_create(int bpffs_fd);
  * @param btf_data Raw BTF data to send to the kernel. Can't be NULL.
  * @param token_fd File descriptor of the BPF token to use, or -1 if no token
  *        should be used.
- * @return 0 on success, or a negative error value on failure.
+ * @return A valid BTF file descriptor on success (which should be closed by
```
qdeslandes: I would replace "which should be closed by the caller" by "owned by the caller" to be more consistent with other functions, this applies to other changes as well.

PR #404, opened 2026-01-27
Add update-set operation for delta updates to sets

pzmarzly:
To avoid passing entire chain on every minor update, let's define a new operation that only takes a delta change, and applies it on bpfilter daemon side.

# Notes

Currently every change to a set will nonetheless trigger codegen regeneration and reloading of the BPF program, but we decided it's not an issue worth superoptimizing for. We have some ideas how to make it faster (turn deletions into bpf_map deletions, keep track of bpf_map free storage, grow it exponentially if needed).

# Example usage

```bash
$ bfcli ruleset set --from-str "
    chain test_xdp BF_HOOK_XDP{ifindex=2} ACCEPT
      set blocked_ips (ip4.saddr) in {192.168.2.1;192.168.2.2}
      rule
        (ip4.saddr) in blocked_ips
        counter
        DROP
  "
$ bfcli chain update-set \
    --name test_xdp \
    --set-name blocked_ips \
    --add 192.168.1.1 --add 192.168.1.2 \
    --remove 192.168.2.1
info   : updated set 'blocked_ips' in chain 'test_xdp'
$ bfcli ruleset get
chain test_xdp BF_HOOK_XDP{ifindex=2} ACCEPT
    counters policy 283 packets 154188 bytes; error 0 packets 0 bytes
    set blocked_ips (ip4.saddr) in {
        192.168.2.2
        192.168.1.1
        192.168.1.2
    }
    rule
        (ip4.saddr) in blocked_ips
        counters 0 packets 0 bytes
        DROP
```

### DIFF
--- a/doc/usage/bfcli.rst
+++ b/doc/usage/bfcli.rst
@@ -239,6 +239,37 @@ If you want to modify the hook options, use ``bfcli chain set`` instead.
               counters 0 packets 0 bytes
               DROP
 
+``chain update-set``
+~~~~~~~~~~~~~~~~~~~~
+
+Atomically update the content of a named set in a chain using delta operations. This is more efficient than replacing the entire chain when you only need to modify set membership.
+
+**Options**
+  - ``--name NAME``: name of the chain containing the set.
+  - ``--set-name NAME``: name of the set to update.
+  - ``--add ELEMENT``: element to add to the set. Can be specified multiple times.
+  - ``--remove ELEMENT``: element to remove from the set. Can be specified multiple times.
+
+At least one of ``--add`` or ``--remove`` must be specified.
+
+**Examples**
+
+.. code:: shell
+
+    $ # Create a chain with a named set
+    $ sudo bfcli chain set --from-str "
+          chain my_filter BF_HOOK_XDP{ifindex=2} ACCEPT
+              set blocklist (ip4.saddr) in { 192.168.1.100 }
+              rule
+                  (ip4.saddr) in blocklist
+                  DROP"
+
+    $ # Add addresses to the blocklist
+    $ sudo bfcli chain update-set --name my_filter --set-name blocklist --add 192.168.1.101 --add 192.168.1.102
+
+    $ # Remove an address from the blocklist
+    $ sudo bfcli chain update-set --name my_filter --set-name blocklist --remove 192.168.1.100
+
 ``chain flush``
 ~~~~~~~~~~~~~~~
 


src/libbpfilter/include/bpfilter/bpfilter.h
```diff
@@ -289,6 +290,40 @@ int bf_chain_attach(const char *name, const struct bf_hookopts *hookopts);
...
+ *   "to_add": { }, // bf_set object or nil
+ *   "to_remove": { } // bf_set object or nil
+ * }
+ * @endcode
+ *
+ * **Response payload format**
+ * The response doesn't contain data.
+ *
+ * @param chain_name Name of the chain containing the set. Can't be NULL.
+ * @param to_add Set containing elements to add. The set name and key format
```
yaakov-stein, line 315: It looks like `to_add` and `to_remove` aren't params here and the naming is incorrect in the `libbpfilter: add update-set operation` commit. I'm guessing that commit won't compile

src/libbpfilter/include/bpfilter/chain.h
```diff
@@ -145,3 +145,26 @@ int bf_chain_add_set(struct bf_chain *chain, struct bf_set *set);
...
+ * @param set_name Name of the set to update. Can't be NULL.
+ * @param to_add Set containing elements to add. Set name and key format must
+ *        match the existing set. Can be NULL if no elements to add.
+ * @param to_remove Set containing elements to remove. Set name and key format
+ *        must match the existing set. Can be NULL if no elements to remove.
+ * @return 0 on success, or a negative errno value on failure, including:
+ * - `-ENOENT`: no set found with the given name.
+ * - `-EINVAL`: set key format doesn't match existing set.
+ */
+int bf_chain_update_set(struct bf_chain *chain, const char *set_name,
```
yaakov-stein: Would this be simpler if we broke it up into base parts - `bf_chain_remove_from_set` and `bf_chain_add_to_set` (or something like that)? I'm wondering if that may be better for readability, future repurposing, and to avoid needing to have several methods with very similar names (update_set vs. set_update).

src/libbpfilter/chain.c
```diff
@@ -270,3 +270,117 @@ struct bf_set *bf_chain_get_set_for_matcher(const struct bf_chain *chain,
...
+            return bf_err_r(
+                -EINVAL,
+                "to_remove set type mismatch: target set use_trie=%d, to_remove use_trie=%d",
+                target_set->use_trie, to_remove->use_trie);
+        }
+
+        bf_list_foreach (&to_remove->elems, elem_node) {
+            const void *elem_to_remove = bf_list_node_get_data(elem_node);
+
+            bf_list_foreach (&target_set->elems, target_elem_node) {
```
yaakov-stein: Is it worth adding a helper `bf_set_remove_elem` - (in line with but the opposite of `bf_set_add_elem`)? Can make things easier if set implementation ever changes

src/bfcli/chain.c
```diff
@@ -264,3 +265,37 @@ int bfc_chain_flush(const struct bfc_opts *opts)
 
     return r;
 }
+
+int bfc_chain_update_set(const struct bfc_opts *opts)
```
yaakov-stein, line 269: Should we keep naming consistent with libbpfilter and have this be `bfc_chain_set_update`? In general, it looks like sometimes it is update_set (`BFC_ACTION_UPDATE_SET`) and sometimes its set_update (ie `BF_REQ_CHAIN_SET_UPDATE`) - is that necessary (seems a bit confusing)?

src/libbpfilter/include/bpfilter/bpfilter.h
pzmarzly: Great catch, I split commits incorrectly. Interestingly, it compiles, since it's just a headers file

src/libbpfilter/include/bpfilter/chain.h
pzmarzly: It would. Splitten in new version

src/bpfilter/cgen/cgen.c
```diff
@@ -401,8 +401,11 @@ int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain)
 
     bf_swap(cgen->program, new_prog);
 
-    bf_chain_free(&cgen->chain);
-    cgen->chain = TAKE_PTR(*new_chain);
+    // Don't free cgen->chain if it's the same object (in-place editing)
+    if (cgen->chain != *new_chain) {
+        bf_chain_free(&cgen->chain);
+        cgen->chain = TAKE_PTR(*new_chain);
+    }
```
qdeslandes: This is an issue: the function expects `struct bf_chain **chain`, meaning the caller owns the chain, which is no possible for in-place editing, because `cgen` owns it already.

src/libbpfilter/chain.c
```diff
@@ -270,3 +270,16 @@ struct bf_set *bf_chain_get_set_for_matcher(const struct bf_chain *chain,
 
     return bf_list_get_at(&chain->sets, set_id);
 }
+
+struct bf_set *bf_chain_get_set_by_name(struct bf_chain *chain, const char *set_name)
+{
+    assert(chain && set_name);
```
qdeslandes: Do 2 `assert()`.

src/libbpfilter/chain.c
```diff
@@ -270,3 +270,16 @@ struct bf_set *bf_chain_get_set_for_matcher(const struct bf_chain *chain,
...
+
+    bf_list_foreach (&chain->sets, set_node) {
+        struct bf_set *set = bf_list_node_get_data(set_node);
+        if (set->name && bf_streq(set->name, set_name))
```
qdeslandes: `bf_streq()` will check `set->name` for NULL, no need to do it.

src/libbpfilter/include/bpfilter/set.h
```diff
@@ -130,3 +130,30 @@ int bf_set_add_elem(struct bf_set *set, const void *elem);
...
+ * Removes all elements present in `to_remove` from `dest`. Both sets must
+ * have the same key format (n_comps, key components, and use_trie must match).
+ * Elements in `to_remove` that aren't present in `dest` are ignored.
+ *
+ * @param dest Destination set to remove elements from. Can't be NULL.
+ * @param to_remove Source set containing elements to remove. Can't be NULL.
+ * @return 0 on success, or a negative errno value on failure, including:
+ * - `-EINVAL`: set key format doesn't match between dest and to_remove.
+ */
+int bf_set_difference(struct bf_set *dest, const struct bf_set *to_remove);
```
qdeslandes: - `bf_set_add` and `bf_set_remove` would be clearer
- Expect `bf_set **` as input: it emphasizes that the function will take ownership of the data and it is responsible for freeing the set on success

src/libbpfilter/set.c
```diff
@@ -416,3 +416,103 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
 }
+
+/**
+ * @brief Check if two sets have the same shape (key format).
+ *
+ * @param first First set. Can't be NULL.
+ * @param second Second set. Can't be NULL.
+ * @return 0 if sets have matching shape, or a negative errno value on mismatch.
+ */
+static int _bf_set_check_shape(const struct bf_set *first, const struct bf_set *second)
```
qdeslandes: `_bf_set_cmp_key`? It more consistent with other comparison functions we have.

src/libbpfilter/set.c
```diff
@@ -416,3 +416,103 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
+{
+    assert(first && second);
```
qdeslandes: 2 `assert()` calls.

src/libbpfilter/set.c
```diff
@@ -416,3 +416,103 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
+                i, bf_matcher_type_to_str(first->key[i]),
+                bf_matcher_type_to_str(second->key[i]));
+        }
+    }
+
+    if (first->use_trie != second->use_trie) {
+        return bf_err_r(-EINVAL,
+                        "set type mismatch: first set use_trie=%d, second use_trie=%d",
+                        first->use_trie, second->use_trie);
+    }
```
qdeslandes: This should not be needed. If the keys are the same, both will either use or not use the trie.

src/libbpfilter/set.c
```diff
@@ -416,3 +416,103 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
+
+    return 0;
+}
+
+int bf_set_union(struct bf_set *dest, const struct bf_set *to_add)
+{
+    int r;
+
+    assert(dest && to_add);
```
qdeslandes: 2 asserts.

src/libbpfilter/set.c
```diff
@@ -416,3 +416,103 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
+
+    r = _bf_set_check_shape(dest, to_add);
+    if (r)
+        return r;
+
+    // This has O(n * m) complexity. We could get to O(n log n + m) by turning
+    // the linked list into an array and sorting it, but we should just replace
+    // underlying bf_list with true hashset and enjoy O(m).
```
qdeslandes: `@todo` would be great here.

src/libbpfilter/set.c
```diff
@@ -416,3 +416,103 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
+            if (memcmp(dest_elem, elem_to_add, dest->elem_size) == 0) {
+                found = true;
+                break;
+            }
+        }
+
+        if (!found) {
+            r = bf_set_add_elem(dest, elem_to_add);
+            if (r)
+                return bf_err_r(r, "failed to add element to set");
```
qdeslandes: This allocates data twice: 
- For the data itself
- For the new list node

Instead, you should steal the node (as we own the set), and insert it in the new list (the list should have operations for that).

src/libbpfilter/set.c
```diff
@@ -416,3 +416,103 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
+    }
+
+    return 0;
+}
+
+int bf_set_difference(struct bf_set *dest, const struct bf_set *to_remove)
+{
+    int r;
+
+    assert(dest && to_remove);
```
qdeslandes: 2 asserts.

tests/unit/libbpfilter/set.c
```diff
@@ -247,6 +247,141 @@ static void new_from_raw_invalid(void **state)
...
+
+    assert_ok(bf_set_add_elem(dest, &elem1));
+    assert_ok(bf_set_add_elem(dest, &elem2));
+
+    assert_ok(bf_set_add_elem(to_add, &elem2));
+    assert_ok(bf_set_add_elem(to_add, &elem3));
+
+    assert_ok(bf_set_union(dest, to_add));
+
+    assert_int_equal(bf_list_size(&dest->elems), 3);
```
qdeslandes, line 272: Check the actual values in the set.

src/bpfilter/xlate/cli.c
```diff
@@ -540,6 +541,87 @@ int _bf_cli_chain_flush(const struct bf_request *request,
...
+        if (r)
+            return r;
+    }
+
+    if (to_add)
+        set_name = to_add->name;
+    else if (to_remove)
+        set_name = to_remove->name;
+    else
+        return bf_err_r(-EINVAL, "at least one of to_add or to_remove must be provided");
```
qdeslandes: "at least one of to_add or to_remove must be provided" -> "missing to_add or to_remove entry". The messages in the daemon target developers not users, so the phrasing can be simplified.

src/bpfilter/xlate/cli.c
```diff
@@ -540,6 +541,87 @@ int _bf_cli_chain_flush(const struct bf_request *request,
...
+
+    cgen = bf_ctx_get_cgen(chain_name);
+    if (!cgen)
+        return bf_err_r(-ENOENT, "chain '%s' does not exist", chain_name);
+
+    dest_set = bf_chain_get_set_by_name(cgen->chain, set_name);
+    if (!dest_set)
+        return bf_err_r(-ENOENT, "set '%s' does not exist", set_name);
+
+    if (to_add) {
```
qdeslandes: Do not update the chain in place, as there are other complications involved. Copy it instead, but none of the primitives are available. A simple solution would be to serialize it, and sererialize in a copy.

src/bfcli/chain.c
```diff
@@ -264,3 +265,61 @@ int bfc_chain_flush(const struct bfc_opts *opts)
...
+            return bf_err_r(r, "failed to create set");
+
+        bf_list_foreach (&opts->set_remove, node) {
+            const char *raw_elem = bf_list_node_get_data(node);
+
+            r = bf_set_parse_elem(to_remove, raw_elem);
+            if (r)
+                return bf_err_r(r, "failed to parse set element '%s'", raw_elem);
+        }
+    }
```
qdeslandes: You don't have to get for `bf_list_is_empty()`, create the list and if there are no elements to add, then it will be serialized empty. The cost is low and it simplifies the code.

tests/e2e/cli/chain_update_set.sh
```diff
@@ -0,0 +1,207 @@
...
+
+set -eux
+set -o pipefail
+
+. "$(dirname "$0")"/../e2e_test_util.sh
+make_sandbox
+
+start_bpfilter
+    # Test set 1: Basic tests
+    ${FROM_NS} bfcli chain set --from-str "chain test_xdp BF_HOOK_XDP{ifindex=${NS_IFINDEX}} ACCEPT
```
qdeslandes: No need to indent

tests/e2e/cli/chain_update_set.sh
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env bash
+
+set -eux
+set -o pipefail
+
+. "$(dirname "$0")"/../e2e_test_util.sh
+make_sandbox
+
+start_bpfilter
+    # Test set 1: Basic tests
```
qdeslandes: Is that a test or setup?

src/bfcli/opts.h
```diff
@@ -71,6 +72,10 @@ struct bfc_opts
     const char *name;
     struct bf_hookopts hookopts;
 
+    const char *set_name;
+    bf_list set_add;
```
qdeslandes, line 78: Add a comment to clarify we do not own the data in the set.

src/bfcli/opts.c
```diff
@@ -354,6 +372,34 @@ static void _bfc_opts_chain_hook_opts_cb(struct argp_state *state,
...
+    opts->set_name = arg;
+};
+
+static void _bfc_opts_set_add_cb(struct argp_state *state,
+                                         const char *arg, struct bfc_opts *opts)
+{
+    if (strlen(arg) == 0)
+        argp_error(state, "--add requires an element");
+
+    bf_list_add_tail(&opts->set_add, (void *)arg);
```
qdeslandes: What if the operation fails?

src/bfcli/opts.c
```diff
@@ -354,6 +372,34 @@ static void _bfc_opts_chain_hook_opts_cb(struct argp_state *state,
...
+};
+
+static void _bfc_opts_set_remove_cb(struct argp_state *state,
+                                             const char *arg,
+                                             struct bfc_opts *opts)
+{
+    if (strlen(arg) == 0)
+        argp_error(state, "--remove requires an element");
+
+    bf_list_add_tail(&opts->set_remove, (void *)arg);
```
qdeslandes: What if the operation fails?

src/bfcli/chain.c
```diff
@@ -264,3 +265,57 @@ int bfc_chain_flush(const struct bfc_opts *opts)
...
     return r;
 }
+
+int bfc_chain_update_set(const struct bfc_opts *opts)
+{
+    _free_bf_set_ struct bf_set *to_add = NULL;
+    _free_bf_set_ struct bf_set *to_remove = NULL;
+    _free_bf_chain_ struct bf_chain *chain = NULL;
+    _free_bf_hookopts_ struct bf_hookopts *_hookopts = NULL;
+    _clean_bf_list_ bf_list _counters = bf_list_default(bf_counter_free, NULL);
```
qdeslandes: No need to prefix with `_`.

src/libbpfilter/cli.c
```diff
@@ -505,6 +506,59 @@ int bf_chain_update(const struct bf_chain *chain)
...
+
+    if (to_add) {
+        bf_wpack_open_object(wpack, "to_add");
+        r = bf_set_pack(to_add, wpack);
+        if (r)
+            return r;
+        bf_wpack_close_object(wpack);
+    } else {
+        bf_wpack_kv_nil(wpack, "to_add");
+    }
```
qdeslandes: `to_add` and `to_remove` are always defined, so instead of sending `nil`, we can pack the chain. If it's empty, then the `to_add` array will be empty, otherwise it will be non-empty. Don't forget to update the doc.

src/libbpfilter/cli.c
```diff
@@ -505,6 +506,59 @@ int bf_chain_update(const struct bf_chain *chain)
...
+    _free_bf_wpack_ bf_wpack_t *wpack = NULL;
+    int r;
+
+    assert(chain_name);
+
+    r = bf_wpack_new(&wpack);
+    if (r)
+        return r;
+
+    bf_wpack_kv_str(wpack, "chain_name", chain_name);
```
qdeslandes: `name` instead of `chain_name`, for consistency with other chain-related functions.

src/libbpfilter/include/bpfilter/chain.h
```diff
@@ -145,3 +145,25 @@ int bf_chain_add_set(struct bf_chain *chain, struct bf_set *set);
...
+
+/**
+ * Copy a chain.
+ *
+ * @param dest The destination chain. It will be allocated during the call.
+ *        Can't be NULL.
+ * @param src The source chain, to copy from. Can't be NULL.
+ * @return 0 on success, negative error code on failure.
+ */
+int bf_chain_copy(struct bf_chain **dest, const struct bf_chain *src);
```
qdeslandes: `bf_chain_new_from_copy` would be more consistent with other functions (and clarify `dest` is allocated and owned by the caller).

src/libbpfilter/set.c
```diff
@@ -425,3 +416,109 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
+
+    for (size_t i = 0; i < first->n_comps; ++i) {
+        if (first->key[i] != second->key[i]) {
+            return bf_err_r(
+                -EINVAL,
+                "set key component %lu type mismatch: expected %s, got %s", i,
+                bf_matcher_type_to_str(first->key[i]),
+                bf_matcher_type_to_str(second->key[i]));
+        }
+    }
```
qdeslandes: `memcmp` would be enough. We don't need to specify which component is invalid.

src/libbpfilter/set.c
```diff
@@ -425,3 +416,109 @@ int bf_set_add_elem(struct bf_set *set, const void *elem)
...
+                break;
+            }
+        }
+
+        if (!found) {
+            // Take ownership of data to stop to_add cleanup from freeing it.
+            elem_to_add = bf_list_node_take_data(elem_node);
+            r = bf_list_add_tail(&dest->elems, elem_to_add);
+            if (r)
+                return bf_err_r(r, "failed to add element to set");
```
qdeslandes: If `bf_list_add_tail` fails, `elem_to_add` will leak. Instead:
- `bf_list_add_tail(bf_list_node_get_data())`
- On success, `bf_list_node_take_data(elem_node)`

src/libbpfilter/include/bpfilter/set.h
```diff
@@ -119,3 +119,43 @@ int bf_set_pack(const struct bf_set *set, bf_wpack_t *pack);
...
+/**
+ * @brief Parse a raw element and insert it into a set.
+ *
+ * The element is parsed according to `set->key`.
+ *
+ * @param set Set to parse the element for. Can't be NULL.
+ * @param raw_elem Raw element to parse. Can't be NULL.
+ * @return 0 on success, or a negative error value on failure.
+ */
+int bf_set_parse_elem(struct bf_set *set, const char *raw_elem);
```
qdeslandes: `bf_set_add_elem_raw()` would be more consistent with `bf_set_add_elem()` now that this logic is public.

qdeslandes (changes requested):
A few more changes, but the PR looks good. Looking forward to using this!

src/bfcli/chain.c
pzmarzly: I wanted to mark that they are never read from, though I wasn't sure if this is a style you use in this project. Will undo.

src/bfcli/chain.c
```diff
@@ -264,3 +265,57 @@ int bfc_chain_flush(const struct bfc_opts *opts)
...
+int bfc_chain_update_set(const struct bfc_opts *opts)
+{
+    _free_bf_set_ struct bf_set *to_add = NULL;
+    _free_bf_set_ struct bf_set *to_remove = NULL;
+    _free_bf_chain_ struct bf_chain *chain = NULL;
+    _free_bf_hookopts_ struct bf_hookopts *hookopts = NULL;
+    _clean_bf_list_ bf_list counters = bf_list_default(bf_counter_free, NULL);
+    struct bf_set *dest_set = NULL;
+    int r;
+
```
qdeslandes, line 278: Ensure `opts->to_add` or `opts->to_remove` is not empty, in which case you should return an error.

src/libbpfilter/set.c
```diff
@@ -432,3 +423,104 @@ bool bf_set_is_empty(const struct bf_set *set)
 
     return bf_list_is_empty(&set->elems);
 }
+
+/**
+ * @brief Check if two sets have the same key format.
+ *
+ * @param first First set. Can't be NULL.
+ * @param second Second set. Can't be NULL.
+ * @return 0 if sets have matching format, or -EINVAL on mismatch.
```
yaakov-stein, line 432: Shouldn't we just return 1 or -1 here like most compare functions? I don't see why we would return -EINVAL if they're not equal - i.e. the lack of equality doesn't seem to be an error if we are looking solely at this method.

PR #408, opened 2026-02-02
Add `REDIRECT` verdict for TC and XDP

qdeslandes:
Add a new `REDIRECT` verdict in addition to `ACCEPT`/`DROP`/`CONTINUE` to redirect a matched packet to a different interface.

The new verdict expects `$iface $dir` with:
- `$iface`: interface to redirect to, name or index
- `$dir`: direction to redirect the packet: in or out of the interface

This new verdict doesn't support Cgroup and Netfilter hooks ([redirection is not supported by BPF on those hooks](https://docs.ebpf.io/linux/helper-function/bpf_redirect/)).


src/bfcli/parser.y
```diff
@@ -176,6 +188,45 @@ verdict         : VERDICT
...
+                }
+                | REDIRECT_TOKEN REDIRECT_IFACE REDIRECT_DIR
+                {
+                    _cleanup_free_ const char *iface = $2;
+                    _cleanup_free_ const char *dir = $3;
+                    long long ifindex;
+                    char *endptr;
+                    enum bf_redirect_dir redirect_dir;
+
+                    // Parse interface: try as number first, then as name
```
pzmarzly: Nit: wouldn't it be nice to also be able to do it in `hookopts`? Maybe extract this into a helper method like `parse_iface`?

qdeslandes: I've merged the logic to parse the interface with the `meta.iface` parser. `ifindex` in hook options is expecting to receive an interface index. Created #409.

PR #411, opened 2026-02-02
build: create a nix flake for development

jordalgo:
This is so we don't have to rely on manual installation of dependencies to develop/build bpfilter, which can be problematic for many reasons (more details on https://nixos.org/). This also allows development on other linux distros.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
 * @qdeslandes
+*.nix @jordalgo @SkohTV
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -1,0 +1,52 @@
+name: Nix Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build_nix:
+    name: "Build: Nix (${{ matrix.runner }})"
+    strategy:
+      matrix:
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
+    runs-on: ${{ matrix.runner }}
+
+    steps:
+    - name: Checkout bpfilter
+      uses: actions/checkout@v4
+    - uses: nixbuild/nix-quick-install-action@v34
+      with:
+        nix_conf: |
+          keep-env-derivations = true
+          keep-outputs = true
+    - name: Restore and save Nix store
+      uses: nix-community/cache-nix-action@v7
+      with:
+        # restore and save a cache using this key
+        primary-key: nix-${{ runner.os }}-${{ hashFiles('**/*.nix', '**/flake.lock') }}
+        # if there's no cache hit, restore a cache by this prefix
+        restore-prefixes-first-match: nix-${{ runner.os }}-
+        # collect garbage until the Nix store size (in bytes) is at most this number
+        # before trying to save a new cache
+        # 1G = 1073741824
+        gc-max-store-size-linux: 1G
+        # do purge caches
+        purge: true
+        # purge all versions of the cache
+        purge-prefixes: nix-${{ runner.os }}-
+        # created more than this number of seconds ago
+        purge-created: 0
+        # or, last accessed more than this number of seconds ago
+        # relative to the start of the `Post Restore and save Nix store` phase
+        purge-last-accessed: 0
+        # except any version with the key that is the same as the `primary-key`
+        purge-primary-key: never
+    - name: Build & check
+      run: |
+        nix fmt -- --check .
+        nix build
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Usual build folder
 /build/
 
+# Nix
+result*
+
 # IDE configuration folder
 .zed
 .vscode
@@ -14,4 +17,3 @@ __pycache__
 
 # Local Claude instructions
 CLAUDE.local.md
-
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/tools/cmake")
 
 include(GitVersion)
 get_version_from_git()
+
 message(NOTICE "bpfilter version ${PROJECT_VERSION}${PROJECT_VERSION_SUFFIX}")
 
 include(GNUInstallDirs)
--- a/derivation.nix
+++ b/derivation.nix
@@ -1,0 +1,92 @@
+{
+  stdenv,
+  lib,
+  cmake,
+  ninja,
+  pkg-config,
+  clang,
+  bison,
+  flex,
+  libbpf,
+  libnl,
+  elfutils,
+  openssl,
+  testers,
+  zlib,
+  zstd,
+  pcre2,
+  xxd,
+  version,
+}:
+
+let
+  fs = lib.fileset;
+
+  # zerocallusedregs is invalid with -target bpf
+  hardeningDisable = [ "zerocallusedregs" ];
+
+in
+{
+  inherit hardeningDisable;
+
+  package = stdenv.mkDerivation (finalAttrs: {
+    pname = "bpfilter";
+    inherit version;
+
+    src = fs.toSource {
+      root = ./.;
+      fileset = fs.unions [
+        ./src
+        ./CMakeLists.txt
+        ./tools/cmake
+      ];
+    };
+
+    inherit hardeningDisable;
+
+    nativeBuildInputs = [
+      cmake
+      ninja
+      pkg-config
+      clang # for building codegen BPF progs
+      bison
+      flex
+    ];
+
+    buildInputs = [
+      libbpf
+      libnl
+      elfutils
+      openssl
+      zlib
+      zstd
+      pcre2
+      xxd
+    ];
+
+    cmakeFlags = [
+      "-DNO_DOCS=1"
+      "-DNO_TESTS=1"
+      "-DNO_CHECKS=1"
+      "-DNO_BENCHMARKS=1"
+    ];
+
+    # We do not run the unit tests because the nix build sandbox doesn't
+    # have access to /sys/kernel/btf/vmlinux.
+    doCheck = false;
+
+    meta.pkgConfigModules = [ "bpfilter" ];
+
+    passthru = {
+      tests.pkg-config = testers.testMetaPkgConfig finalAttrs.finalPackage;
+    };
+
+    preFixup = ''
+      substituteInPlace $out/lib/systemd/system/bpfilter.service \
+        --replace-fail /usr/sbin/bpfilter $out/bin/bpfilter
+
+      # workaround for https://github.com/NixOS/nixpkgs/issues/144170
+      substituteInPlace $out/lib/pkgconfig/bpfilter.pc --replace-fail ''${prefix}/ ""
+    '';
+  });
+}
--- a/doc/developers/build.rst
+++ b/doc/developers/build.rst
@@ -1,7 +1,7 @@
 Build from sources
 ==================
 
-This document describes the process to build ``bpfilter`` from sources. While ``bpfilter`` can be built on most systems, a recent (6.6+) Linux kernel is required with ``libbpf`` 1.2+ to run the ``bpfilter`` daemon. ``bpfilter`` officially supports Fedora 40+, CentOS Stream 9+, and Ubuntu 24.04+.
+This document describes the process to build ``bpfilter`` from sources. While ``bpfilter`` can be built on most systems, a recent (6.6+) Linux kernel is required with ``libbpf`` 1.2+ to run the ``bpfilter`` daemon. ``bpfilter`` officially supports Fedora 40+, CentOS Stream 9+, and Ubuntu 24.04+. There is also a nix flake which supports all the make targets except for ``doc``.
 
 If you want to perform a full build of ``bpfilter`` (including all test tests, code check, benchmarks, and documentation), the following dependencies are required:
 
@@ -82,6 +82,11 @@ If you want to perform a full build of ``bpfilter`` (including all test tests, c
         python3-sphinx \
         sed \
         xxd
+
+    # Nix (this is experimental)
+    nix develop
+    # or
+    nix build
 
 You can then use CMake to generate the build system:
 
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,110 @@
+{
+  description = "bpfilter - eBPF-based packet filtering framework";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
+  outputs =
+    { self, nixpkgs }:
+    let
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+      ];
+      forAllSystems = nixpkgs.lib.genAttrs systems;
+      nixpkgsFor = forAllSystems (system: import nixpkgs { inherit system; });
+
+      version = "0.0.1";
+
+    in
+    {
+      packages = forAllSystems (
+        system:
+        let
+          pkgs = nixpkgsFor.${system};
+          bpfilterLib = pkgs.callPackage ./derivation.nix { inherit version; };
+        in
+        {
+          default = bpfilterLib.package;
+          bpfilter = bpfilterLib.package;
+        }
+      );
+
+      formatter = forAllSystems (system: nixpkgsFor.${system}.nixfmt-rfc-style);
+
+      devShells = forAllSystems (
+        system:
+        let
+          pkgs = nixpkgsFor.${system};
+          bpfilterLib = pkgs.callPackage ./derivation.nix { inherit version; };
+        in
+        {
+          default = pkgs.mkShell {
+            name = "bpfilter-dev";
+
+            inherit (bpfilterLib) hardeningDisable;
+
+            inputsFrom = [ bpfilterLib.package ];
+
+            packages = with pkgs; [
+              gnumake
+              clang-tools # clang-tidy, clang-format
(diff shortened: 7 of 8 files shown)


flake.nix
```diff
@@ -0,0 +1,105 @@
...
+            gawk
+            jq
+            gnused
+            xxd
+            procps
+            lcov
+
+            # Documentation (though this doesn't quite work yet)
+            doxygen
+            pythonEnv
```
fzakaria: I would add: `venvShellHook`
I found moving more to venv when possible is better for more "normal" IDE.

venvShellHook to source a Python 3 venv at the venvDir location. A venv is created if it does not yet exist. postVenvCreation can be used to to run commands only after venv is first created.

flake.nix
```diff
@@ -0,0 +1,105 @@
...
+          breathe
+          sphinx
+          furo
+          python-dateutil
+          gitpython
+          scapy
+        ]);
+
+      in {
+        devShells.default = pkgs.mkShell {
```
fzakaria: do you plan to build it too?

Feel free to look at https://github.com/NixOS/nixpkgs/pull/474393
I did this for another similar project.

I tried to upstream it https://github.com/d-e-s-o/bpflint/pull/70 but @d-e-s-o thought it was better to just upstream simpler to nixpkgs

flake.nix
```diff
@@ -0,0 +1,105 @@
...
+          name = "bpfilter-dev";
+
+          # Disable hardening flags that break BPF compilation
+          hardeningDisable = [ "stackprotector" "fortify" "zerocallusedregs" ];
+
+          nativeBuildInputs = with pkgs; [
```
fzakaria: i dont think there is a distinction for nativeBuildInputs and buildInputs for devshells.
you can just use `packages =` which is a meta one.

flake.nix
```diff
@@ -0,0 +1,105 @@
...
+            cmake
+            ninja
+            gnumake
+            pkg-config
+
+            # Compilers
+            clang
+            clang-tools  # clang-tidy, clang-format
+            include-what-you-use
+            gcc
```
fzakaria: did this work okay having both?
Usually they have conflicts with shared bin symlinks like `cc`.

flake.nix
```diff
@@ -0,0 +1,105 @@
...
+            lcov
+
+            # Documentation (though this doesn't quite work yet)
+            doxygen
+            pythonEnv
+          ];
+
+          shellHook = ''
+            # Add libbpf headers to include path for clang-tidy
+            export CPATH="${pkgs.libbpf}/include''${CPATH:+:$CPATH}"
```
fzakaria: if you add the `.dev` output of a library, it should already be in the NIX_INCLUDE_DIR and NIX_C_FLAGS etc.. via a hook..

flake.nix
jordalgo: As @pzmarzly pointed out, someone already added it as a nixpkg - https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/bp/bpfilter/package.nix

I'll probably copy parts from their flake for the build stuff.

flake.nix
jordalgo: Yeah, it didn't seem to mind. clang is used to compile the codegen BPF progs

flake.nix
jordalgo: I'm still having trouble with `make doc` - perhaps because nix doesn't have all the packages required and we can't install all these with `pip` because cmake still requires things like sphinx.

flake.nix
jordalgo: This is just for clang-tidy. But you're saying I can possibly remove this by using libbpf-dev?

flake.nix
qdeslandes: `-DNO_DOCS=1` will disable the Python dependencies (or at least it should!). Does nix allows for `pip install ...` within the env? That would solve it. We do it for Ubuntu already, as not all the required Python dependencies are packaged.

flake.nix
qdeslandes: You should probably ensure CMake uses GCC until we move to Clang.

.github/workflows/ci.yaml
qdeslandes, line 1: This should probably have its dedicated workflow.

fzakaria: As a Nix advocate, I offered to work with @jordalgo to help curate the Nix recipe as best I can.
A good end state will be:

* The ability to build with Nix `nix build` 

For instance you can have people try out `bpfilter` via `❯ nix run github:facebok/bpflter` on any Nix machine and have it reproducible (not bit-wise) and include it's own libc/loader nicely validated with a reproducible hash in the path.

* The ability to give you a reproducible developer environment `nix develop`. Same strong guarantees as the build but taken 1 edge further to the inputs needed to build the package.

If you need to support traditional filesystem hierarchy codebases like Ubuntu, then probably having a duplicate workflow is unavoidable. The primitives are too orthogonal since Nix eschews all the FHS stuff.

Good news is that Nix the package manager is now a Fedora package though officially :P

Cool to see someone up-streamed it to Nixpkgs though already!

flake.nix
```diff
@@ -0,0 +1,138 @@
+{
+  description = "bpfilter - eBPF-based packet filtering framework";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-linux" ] (system:
```
SkohTV: We *can* replace `flake-utils` with native Nix functions

```nix
systems = ["x86_64-linux" "aarch64-linux"];
forAllSystems = nixpkgs.lib.genAttrs systems;
nixpkgsFor = forAllSystems (system: import nixpkgs { inherit system; });
```

We'd then use `nixpkgsFor.${system}` instead of `nixpkgs`
```nix
devShells = forAllSystems (system: {
  default = nixpkgsFor.${system}.mkShell {
		...
    packages = (with nixpkgsFor.${system}; [
```

ref: https://github.com/Aylur/ags/blob/main/flake.nix#L16-L20
ref: https://github.com/SkohTV/quantum/blob/v6-dev/flake.nix#L27

flake.nix
```diff
@@ -0,0 +1,138 @@
...
+        packages.default = bpfilter;
+
+        devShells.default = pkgs.mkShell {
+          name = "bpfilter-dev";
+
+          inherit hardeningDisable;
+
+          venvDir = ".venv";
+
+          packages = commonNativeBuildInputs ++ commonBuildInputs ++ (with pkgs; [
```
SkohTV: I think we should split this into 2 files:
- `default.nix` or `derivation.nix` or `something-else.nix` which is the package to build
- `flake.nix` which is the flake in use, it imports the derivation

Should make it cleaner and be possible to use the very cool
```nix
pkgs.mkShell {
  inputsFrom = [ pkgs.callPackage ./default.nix { }; ];
};
```

flake.nix
jordalgo: We can do a `pip install` but when I ran cmake it was complaining about missing the sphinx dep. Perhaps if I use `DNO_DOCS=1` it will work but then I assume `make doc` then does not.

SkohTV: @jordalgo 
Mixing `venvShellHook` and `python3Packages` is to avoid, `python3Packages` is the better solution

The main issue here is that `linuxdoc` (and its dependency `fspath`) are not on nixpkgs yet (the other packages installed by pip are already available trough `python3Packages`)

I was able to package them both on my fork of nixpkgs and `make doc` by doing so:

```diff
- nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+ nixpkgs.url = "github:SkohTV/nixpkgs/python-linuxdoc";
```
```diff
- venvDir = ".venv";
```
```diff
  doxygen
  python3
  python3Packages.sphinx
  python3Packages.breathe
  python3Packages.furo
+ python3Packages.linuxdoc
+ python3Packages.gitpython
+ python3Packages.scapy
+ python3Packages.python-dateutil
+ python3Packages.setuptools
- python3Packages.venvShellHook
  glibcLocales
]);

-postVenvCreation = ''
-  pip install linuxdoc python-dateutil gitpython scapy
-'';
```

I've opened a PR on nixpkgs to upstream `linuxdoc` and `fspath`

jordalgo: > I've opened a PR on nixpkgs to upstream linuxdoc and fspath

Awesome. Thanks!

qdeslandes: > We can do a `pip install` but when I ran cmake it was complaining about missing the sphinx dep. Perhaps if I use `DNO_DOCS=1` it will work but then I assume `make doc` then does not.

Correct, with `-DNO_DOCS=1` the doc is disabled, but without it `make doc` will fail. So, if we can't install the required deps for the docs, we might as well disable is explicitly.

.github/workflows/ci.yaml
qdeslandes: I'm fine with a duplicate workflow, both should run on PR and merge to main. The main reason for moving this out of `ci.yaml` is to have a clear distinction between distro workflow and nix workflow :)

qdeslandes:
> > However, as I don't use it (for now), I don't want the burden of maintaining it, so someone needs to take care of this part (there is a .github/CODEOWNERS file for this).
> 
> @qdeslandes I'm fine to be the owner of this for now. I also think (as others have pointed out) there is not much maintenance burden. Claude (and probably other AIs) are pretty good at nix (it helped me create the first draft of this) and adding new dependencies/packages to the nix flake is pretty easy. Additionally, we can mark that nix support is currently experimental.

I feel like I'm the "old man yells at clouds" at the moment. I probably sounded more assertive that I actually felt (that's what 4 days of conferences and lack of sleep does to someone I guess) :p 

Basically, I'm trying to preserve my (limited) time to work on the project itself and not so much the tooling. Given all the people interacting with this PR (5, a new record!), I sense that it shouldn't be too difficult to find a good soul to help me take care of potential issues coming out of this.

Though I have to say the excitement here is contagious!

qdeslandes (changes requested):
Changing the PR state to "Request changes" while the comments are addressed and consensus is built :)

flake.nix
jordalgo: The default that nix uses is GCC but just FYI @fzakaria and I tested and it seems to build just fine with Clang.

.github/workflows/ci.yaml
jordalgo: Done. Moved to a different workflow.

flake.nix
jordalgo: Thanks for the suggestion. Done.

flake.nix
jordalgo: Done. Now we have derivation.nix for the build.

jordalgo:
@fzakaria @SkohTV - I think this is ready for another review 🙏

SkohTV:
LGMT :)
If there's any PR in the future where you need help with Nix, feel free to `@` me !

I'll open a PR when https://github.com/NixOS/nixpkgs/pull/486778 gets merged

.github/workflows/nix.yaml
```diff
@@ -0,0 +1,27 @@
...
+    name: "Build: Nix (${{ matrix.runner }})"
+    strategy:
+      matrix:
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
+    runs-on: ${{ matrix.runner }}
+
+    steps:
+    - name: Checkout bpfilter
+      uses: actions/checkout@v4
+    - name: Install Nix
```
fzakaria: nit: I like using non-Determinate.. maybe I'm a bit of a luddite.

```
      - uses: nixbuild/nix-quick-install-action@v34
        with:
          nix_conf: |
            keep-env-derivations = true
            keep-outputs = true
      - name: Restore and save Nix store
        uses: nix-community/cache-nix-action@v7
        with:
          # restore and save a cache using this key
          primary-key: nix-${{ runner.os }}-${{ hashFiles('**/*.nix', '**/flake.lock') }}
          # if there's no cache hit, restore a cache by this prefix
          restore-prefixes-first-match: nix-${{ runner.os }}-
          # collect garbage until the Nix store size (in bytes) is at most this number
          # before trying to save a new cache
          # 1G = 1073741824
          gc-max-store-size-linux: 1G
          # do purge caches
          purge: true
          # purge all versions of the cache
          purge-prefixes: nix-${{ runner.os }}-
          # created more than this number of seconds ago
          purge-created: 0
          # or, last accessed more than this number of seconds ago
          # relative to the start of the `Post Restore and save Nix store` phase
          purge-last-accessed: 0
          # except any version with the key that is the same as the `primary-key`
          purge-primary-key: never
      - name: Build & check
        run: |
          nix fmt -- --check . --exclude ./node_modules/
          nix build
```

.gitignore
```diff
@@ -1,6 +1,9 @@
 # Usual build folder
 /build/
 
+# Nix
+result
```
fzakaria: It will create result-dev, result-out etc.. if you build multiple at once.

```suggestion
result*
```

derivation.nix
```diff
@@ -0,0 +1,89 @@
...
+    inherit hardeningDisable;
+
+    nativeBuildInputs = [
+      cmake
+      ninja
+      pkg-config
+      clang # for building codegen BPF progs
+      bison
+      flex
+      git
```
fzakaria: i don't think you need when i saw the logs.
You probably want to disable the git cmake module.

derivation.nix
```diff
@@ -0,0 +1,89 @@
...
+
+    # We do not run the unit tests because the nix build sandbox doesn't
+    # have access to /sys/kernel/btf/vmlinux.
+    doCheck = false;
+
+    preFixup = ''
+      substituteInPlace $out/lib/systemd/system/bpfilter.service \
+        --replace-fail /usr/sbin/bpfilter $out/bin/bpfilter
+
+      # workaround for https://github.com/NixOS/nixpkgs/issues/144170
```
fzakaria, line 88: nit: check https://github.com/NixOS/nixpkgs/pull/454243/files
You can see how you add more tests here to validate stuff like pgk-config.

```
  passthru = {
    tests.pkg-config = testers.testMetaPkgConfig finalAttrs.finalPackage;
  };
```

You can also have testers that validate it "runs"

```
  passthru = {
    tests.version = testers.testVersion {
      package = finalAttrs.finalPackage;
      command = "starpls version";
      version = "v${
        lib.strings.substring 0 (builtins.stringLength finalAttrs.version - 6) finalAttrs.version
      }";
    };
  };
```

doc/developers/build.rst
```diff
@@ -83,6 +83,9 @@ If you want to perform a full build of ``bpfilter`` (including all test tests, c
         sed \
         xxd
 
+    # Nix (this is experimental)
+    nix develop
```
fzakaria, line 87: or you can say `nix build` -- that is build from source too! :)

flake.nix
```diff
@@ -0,0 +1,95 @@
...
+              python3
+              python3Packages.sphinx
+              python3Packages.breathe
+              python3Packages.furo
+              # python3Packages.linuxdoc
+              python3Packages.gitpython
+              python3Packages.scapy
+              python3Packages.python-dateutil
+              python3Packages.setuptools
+              glibcLocales
```
fzakaria: I still think lots of the extra packages here aren't needed but that can be refined in subsequent iterations I say.

.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
 * @qdeslandes
+*.nix @jordalgo
```
qdeslandes: @SkohTV wanna be added here?

flake.nix
jordalgo: I can comment them out for now until we get `make doc` working anyway.

CMakeLists.txt
```diff
@@ -15,8 +15,11 @@ project(bpfilter
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/tools/cmake")
 
-include(GitVersion)
-get_version_from_git()
+option(NO_GIT_VERSION "Disable git-based version detection" 0)
+if (NOT NO_GIT_VERSION)
+    include(GitVersion)
+    get_version_from_git()
+endif ()
```
qdeslandes: `get_version_from_git()` starts with:
```
    find_package(Git)
    if (NOT Git_FOUND)
        message(WARNING "Failed to find git, using default version ${DEFAULT_PROJECT_VERSION}")
        return()
    endif ()
```
Isn't it enough?

jordalgo: GitVersion is included if we don't have this. I could add that to the if block in `get_version_from_git`

qdeslandes: I meant: why is `include(GitVersion)` an issue? If git can't be found, it falls back to `DEFAULT_PROJECT_VERSION` and continue.

jordalgo: It's required for building - I can include it in the list of nix packages but it seemed unnecessary just for this.

.github/CODEOWNERS
jordalgo: More folks can be added later as need right? Right now this is just for the ownership of 3 files :)

CMakeLists.txt
qdeslandes: I think it would be useful: the built commit is embedded in the final binary, useful for debugging :)

.github/CODEOWNERS
SkohTV: > @SkohTV wanna be added here?

Yeah, that'd be cool !
I wasn't sure if I could be in `CODEOWNERS` since I'm an external contributor ^^"

jordalgo: Added.

CMakeLists.txt
jordalgo: Done. Removed this part.

PR #412, opened 2026-02-03
lib: use tmpfile when writing out bpfilter state

etsal:
The bpfilter daemon updates persistent state by reopening the existing file, truncating it, then overwriting the state. If the daemon crashes during this process, when it restarts it will find either no data, or malformed data. Turn the state update atomic by instead writing the new persistent data to a temporary file, then moving it to the correct location.

### DIFF
--- a/src/libbpfilter/helper.c
+++ b/src/libbpfilter/helper.c
@@ -8,6 +8,8 @@
 #include <ctype.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <libgen.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/libbpfilter/helper.c
```diff
@@ -98,32 +99,68 @@ int bf_read_file(const char *path, void **buf, size_t *len)
...
+		return bf_err_r(-ENOMEM, "could not duplicate path %s", path);
+
+	dir = dirname(tmp);
+
+	ret = snprintf(tmpname, len, "%s%s", dir, template);
+	free(tmp);
+
+	if (ret >= len) {
+		return bf_err_r(-ENAMETOOLONG, "tmpfile name too long (%lu bytes)", ret);
+	}
```
qdeslandes: No braces for a single statement.

src/libbpfilter/helper.c
```diff
@@ -98,32 +99,68 @@ int bf_read_file(const char *path, void **buf, size_t *len)
     return 0;
 }
 
+
+static int bf_tmp_name(char *tmpname, size_t len, const char *path)
+{
+	char template[] = "/bpfilter_XXXXXX";
+	char *tmp, *dir;
```
qdeslandes: `_cleanup_free_ char *tmp = NULL;` for automatic cleanup.

src/libbpfilter/helper.c
```diff
@@ -98,32 +99,68 @@ int bf_read_file(const char *path, void **buf, size_t *len)
     return 0;
 }
 
+
+static int bf_tmp_name(char *tmpname, size_t len, const char *path)
```
qdeslandes: Static symbols are prefixed with `_`.

src/libbpfilter/helper.c
```diff
@@ -98,32 +99,68 @@ int bf_read_file(const char *path, void **buf, size_t *len)
...
     ssize_t r;
+    char tmpname[PATH_MAX];
 
-    assert(path);
-    assert(buf);
+    bf_assert(path);
+    bf_assert(buf);
 
-    fd = open(path, O_TRUNC | O_CREAT | O_WRONLY, OPEN_MODE_644);
+    err = bf_tmp_name(tmpname, sizeof(tmpname), path);
```
qdeslandes: `bf_tmp_name` does not abide by its contract, it creates a temporary file template name, not a temporary file name.

Additionally, it would be more convenient to have `bf_get_tmpfile_fd(const char *base, char tmpfile_path[static PATH_MAX])`
- `base`: base path to create the tmp file into (after applying `dirname`), so we can pass `path` directly. The final tmpfile will be `dirname(base)/"bpfilter.tmp.XXXXXX"`
- `tmpfile_path`: path to the new tmp file
- Returns a file descriptor to the tmp file

src/libbpfilter/helper.c
```diff
@@ -98,32 +99,68 @@ int bf_read_file(const char *path, void **buf, size_t *len)
...
+    if (err < 0)
+        return bf_err_r(errno, "failed to get temporary file name");
```
qdeslandes: `bf_tmp_name` doesn't set errno, but returns `-errno` on error.

src/libbpfilter/helper.c
```diff
@@ -98,32 +99,68 @@ int bf_read_file(const char *path, void **buf, size_t *len)
     return 0;
 }
 
+
```
qdeslandes: Extra blank line.

etsal:
New version addressing the feedback.

src/libbpfilter/helper.c
```diff
@@ -98,32 +99,66 @@ int bf_read_file(const char *path, void **buf, size_t *len)
     return 0;
 }
 
+static int _bf_get_tmpfile_fd(const char *base, char tmpfile_path[PATH_MAX])
+{
+    char template[] = "/bpfilter.tmp.XXXXXX";
+    _cleanup_free_ char *tmp;
```
qdeslandes: `tmp = NULL` to ensure we don't try to cleanup a garbage point if we exit early.

src/libbpfilter/helper.c
```diff
@@ -98,32 +99,66 @@ int bf_read_file(const char *path, void **buf, size_t *len)
     return 0;
 }
 
+static int _bf_get_tmpfile_fd(const char *base, char tmpfile_path[PATH_MAX])
```
qdeslandes: `[static PATH_MAX]`

src/libbpfilter/helper.c
```diff
@@ -98,32 +99,66 @@ int bf_read_file(const char *path, void **buf, size_t *len)
...
+    int fd;
+
+    /* The dirname call may modify its argument. */
+    tmp = strdup(base);
+    if (!tmp)
+        return bf_err_r(ENOMEM, "could not duplicate path %s", base);
+
+    dir = dirname(tmp);
+
+    path_len = snprintf(tmpfile_path, PATH_MAX, "%s%s", dir, template);
```
qdeslandes: No need to have a variable for template here: `"%s/bpfilter.tmp.XXXXXX"` should do it

PR #414, opened 2026-02-06
lib: matcher: add meta.flow_probability matcher

agelun:
Add a new matcher that randomly selects a percentage of flows based on flow hash. Unlike meta.probability which uses per-packet randomness, meta.flow_probability uses the packet's 5-tuple hash ensuring all packets in the same flow receive the same decision.

- Uses bf_flow_hash elfstub for deterministic per-flow hashing
- Accepts floating-point percentages (e.g., 33.33%, 0.1%, 99.99%)
- Stores probability as float for precision
- Supported on TC and CGROUP hooks (XDP and NF unsupported)

The flow hash is computed from src/dst IP, src/dst port, protocol, and IPv6 flow label, then passed through xxHash32 avalanche finalizer for uniform distribution.


src/bfcli/lexer.l
```diff
@@ -153,6 +155,7 @@ ip6\.nexthdr    { BEGIN(STATE_MATCHER_L4_PROTO); yylval.sval = strdup(yytext); r
 }
 
 meta\.probability  { BEGIN(STATE_MATCHER_META_PROBA); yylval.sval = strdup(yytext); return MATCHER_TYPE; }
+meta\.flow_probability { BEGIN(STATE_MATCHER_META_FLOW_PROBA); yylval.sval = strdup(yytext); return MATCHER_TYPE; }
```
qdeslandes: This should be above `<STATE_MATCHER_META_FLOW_PROBA>{`

src/bpfilter/bpf/flow_hash.bpf.c
qdeslandes, line 1: `flow_hash.bpf.c` -> `flow_proba.bpf.c`

src/bpfilter/cgen/matcher/meta.c
```diff
@@ -202,6 +204,50 @@ static int _bf_matcher_generate_meta_flow_hash(struct bf_program *program,
...
+     * Arguments:
+     * - r1: pointer to bf_runtime context
+     * - r2: L3 protocol ID (from r7, set by prologue)
+     * - r3: L4 protocol ID (from r8, set by prologue)
+     *
+     * Return: hash value in r0
+     */
+
+    // Set up elfstub arguments
+    EMIT(program, BPF_MOV64_REG(BPF_REG_1, BPF_REG_10));
```
qdeslandes, line 237: Before calling the elfstub, you need to ensure the packet contains the protocols you expect, and jump to the next rule if not. I.e., it should be either IPv4 or IPv6, with TCP or UDP.

src/libbpfilter/include/bpfilter/matcher.h
```diff
@@ -64,6 +64,8 @@ enum bf_matcher_type
     BF_MATCHER_META_MARK,
     /// Matches a specific flow hash (source/destination IP and ports).
     BF_MATCHER_META_FLOW_HASH,
+    /// Matches packets based on flow hash probability (consistent per flow).
```
qdeslandes: "flow hash probability" -> "flow probability", otherwise it will quickly become confusive with `meta.flow_hash`.

src/libbpfilter/matcher.c
```diff
@@ -924,6 +958,20 @@ static struct bf_matcher_meta _bf_matcher_metas[_BF_MATCHER_TYPE_MAX] = {
                                    _bf_parse_int_range, _bf_print_int_range),
                 },
         },
+    [BF_MATCHER_META_FLOW_PROBABILITY] =
+        {
+            .layer = BF_MATCHER_NO_LAYER,
+            .unsupported_hooks =
+                BF_FLAGS(BF_HOOK_XDP, BF_HOOK_NF_FORWARD, BF_HOOK_NF_LOCAL_IN,
+                         BF_HOOK_NF_LOCAL_OUT, BF_HOOK_NF_POST_ROUTING,
+                         BF_HOOK_NF_PRE_ROUTING),
```
qdeslandes: Why so?

src/libbpfilter/matcher.c
```diff
@@ -397,6 +397,40 @@ void _bf_print_probability(const void *payload)
...
+static void _bf_print_flow_probability(const void *payload)
+{
+    assert(payload);
+
+    float proba = *(float *)payload;
+    if (proba == (float)(int)proba)
+        (void)fprintf(stdout, "%.0f%%", proba);
+    else
+        (void)fprintf(stdout, "%.2f%%", proba);
+}
```
qdeslandes, line 436: This function is not exercised during the unit tests, even though `bf_matcher_dump()` is used.

tools/benchmarks/main.cpp
```diff
@@ -513,6 +514,107 @@ void single_rule__tcp_sport(::benchmark::State &state)
...
+        if (stats.retval != ::bf::CGROUP_DROP)
+            state.SkipWithError("benchmark run failed");
+
+        state.SetIterationTime((double)stats.duration * stats.repeat);
+    }
+
+    state.counters["nInsn"] = prog.nInsn();
+    state.SetLabel("1 rule, meta.flow_probability, CGROUP");
+}
+BENCHMARK(single_rule__meta_flow_probability_cgroup);
```
qdeslandes: We only need a single benchmark for `meta.flow_hash` and same for `flow_probability`. No need to include the hook name in the name of the benchmark or the label.

src/bpfilter/bpf/flow_hash.bpf.c
agelun: Formally the code in the file still computes hash, not probability

src/bpfilter/cgen/matcher/meta.c
```diff
@@ -202,6 +205,56 @@ static int _bf_matcher_generate_meta_flow_hash(struct bf_program *program,
...
+                                -(int)sizeof(struct bf_runtime))); // r1 = ctx
+    EMIT(program, BPF_MOV64_REG(BPF_REG_2, BPF_REG_7)); // r2 = l3_proto
+    EMIT(program, BPF_MOV64_REG(BPF_REG_3, BPF_REG_8)); // r3 = l4_proto
+
+    // Call the elfstub - result in r0
+    EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_FLOW_HASH);
+
+    // Compare the computed hash with the threshold based on probability
+    // The hash is uniformly distributed across 32 bits, so we compare against
+    // UINT32_MAX * (proba / 100.0) to select the desired percentage of flows
```
qdeslandes: Multiline comment as above.

src/bpfilter/cgen/matcher/meta.c
```diff
@@ -202,6 +205,56 @@ static int _bf_matcher_generate_meta_flow_hash(struct bf_program *program,
...
+     * This ensures all packets in a flow get the same hash value, making
+     * the probability decision consistent per-flow rather than per-packet.
+     *
+     * Arguments:
+     * - r1: pointer to bf_runtime context
+     * - r2: L3 protocol ID (from r7, set by prologue)
+     * - r3: L4 protocol ID (from r8, set by prologue)
+     *
+     * Return: hash value in r0 */
+    // Set up elfstub arguments
```
qdeslandes: Unnecessary comment.

src/bpfilter/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ bf_target_add_elfstubs(bpfilter
         "parse_ipv6_nh"
         "update_counters"
         "log"
+        "flow_hash"
```
qdeslandes, line 66: "flow_probability" is missing

src/libbpfilter/matcher.c
```diff
@@ -397,6 +400,45 @@ void _bf_print_probability(const void *payload)
...
+    assert(raw_payload);
+
+    double proba;
+    char *endptr;
+
+    locale_t c_locale = newlocale(LC_NUMERIC_MASK, "C", (locale_t)0);
+    if (!c_locale)
+        return bf_err_r(-ENOMEM, "failed to create C locale");
+    proba = strtod_l(raw_payload, &endptr, c_locale);
+    freelocale(c_locale);
```
qdeslandes: Did you hit issues with the locale? Dealing with locales is out of the scope of this commit, especially if it only applies to a single matcher. I would suggest either:
- Add a commit prior to this change to support localization for matchers more globally, so we don't call `newlocale` each time we process a single matcher.
- Open an issue, so it can be fixed as a follow-up

src/libbpfilter/matcher.c
```diff
@@ -397,6 +400,45 @@ void _bf_print_probability(const void *payload)
...
+
+    return -EINVAL;
+}
+
+static void _bf_print_flow_probability(const void *payload)
+{
+    assert(payload);
+
+    float proba = *(float *)payload;
+    if (proba == (float)(int)proba)
```
qdeslandes: Use `floorf()` from `math.h`.

src/libbpfilter/matcher.c
```diff
@@ -924,6 +966,20 @@ static struct bf_matcher_meta _bf_matcher_metas[_BF_MATCHER_TYPE_MAX] = {
                                    _bf_parse_int_range, _bf_print_int_range),
                 },
         },
+    [BF_MATCHER_META_FLOW_PROBABILITY] =
+        {
+            .layer = BF_MATCHER_NO_LAYER,
+            .unsupported_hooks =
+                BF_FLAGS(BF_HOOK_NF_FORWARD, BF_HOOK_NF_LOCAL_IN,
+                         BF_HOOK_NF_LOCAL_OUT, BF_HOOK_NF_POST_ROUTING,
+                         BF_HOOK_NF_PRE_ROUTING),
```
qdeslandes: What is the limitation with those hooks?

tests/e2e/matchers/meta_flow_probability.sh
```diff
@@ -0,0 +1,36 @@
...
+bfcli ruleset set --dry-run --from-str "chain test BF_HOOK_TC_INGRESS ACCEPT rule meta.flow_probability eq 100% counter DROP"
+bfcli ruleset set --dry-run --from-str "chain test BF_HOOK_TC_EGRESS ACCEPT rule meta.flow_probability eq 50% counter DROP"
+bfcli ruleset set --dry-run --from-str "chain test BF_HOOK_CGROUP_INGRESS ACCEPT rule meta.flow_probability eq 50% counter DROP"
+bfcli ruleset set --dry-run --from-str "chain test BF_HOOK_CGROUP_EGRESS ACCEPT rule meta.flow_probability eq 50% counter DROP"
+
+# Floating-point percentages
+bfcli ruleset set --dry-run --from-str "chain test BF_HOOK_TC_INGRESS ACCEPT rule meta.flow_probability eq 33.33% counter DROP"
+bfcli ruleset set --dry-run --from-str "chain test BF_HOOK_TC_INGRESS ACCEPT rule meta.flow_probability eq 0.1% counter DROP"
+bfcli ruleset set --dry-run --from-str "chain test BF_HOOK_TC_INGRESS ACCEPT rule meta.flow_probability eq 99.99% counter DROP"
+bfcli ruleset set --dry-run --from-str "chain test BF_HOOK_TC_INGRESS ACCEPT rule meta.flow_probability eq 50.0% counter DROP"
```
qdeslandes, line 28: Add validate for `0%`, `0.00%`, `100%`, and `100.00%` (only one test each, no need to validate every hook).

src/bpfilter/bpf/flow_hash.bpf.c
```diff
@@ -0,0 +1,116 @@
...
+        for (int i = 0; i < IPV6_ADDR_WORDS; ++i)
+            hash ^= daddr[i];
+
+        // Include flow label (20 bits from version/traffic class/flow label)
+        // First 4 bytes contain: version (4) + traffic class (8) + flow label (20)
+        // After ntohl, flow label is in the lower 20 bits
+        hash ^= bpf_ntohl(*(__u32 *)ip6) & IPV6_FLOW_LABEL_MASK;
+
+        hash ^= (__u32)l4_proto << 16;
+    }
```
qdeslandes, line 107: Consider updating the matcher documentation (`bfcli.rst`) to mention that:
- This matcher will match the same connection for both direction (the XOR used is commutative)
- This matcher only applies when the packet is IPv4/IPv6 on L3, and TCP/UDP on L4

src/bpfilter/CMakeLists.txt
agelun: elfstub is "flow hash" (because it computes hash), there's no "flow_probability" one

src/bpfilter/bpf/flow_hash.bpf.c
agelun: Actually src/dst ports are different, so the hash will also be different. But it's a good point, I should make IPv4/6 address accounted differently as well.

src/libbpfilter/matcher.c
agelun: Didn't test this, and don't know if that will work at all. In theory it should.

src/bpfilter/bpf/flow_hash.bpf.c
agelun: Updated the doc to highlight ip+tcp/udp

src/libbpfilter/matcher.c
agelun: Ok, "simplified" this part, let's fix it separately

src/libbpfilter/matcher.c
agelun: Not tested, in theory should work

src/bpfilter/CMakeLists.txt
qdeslandes: My bad on the misread, please update https://github.com/facebook/bpfilter/blob/main/tests/fuzz/keywords.dict to add the new matcher.

src/libbpfilter/matcher.c
```diff
@@ -3,6 +3,8 @@
  * Copyright (c) 2022 Meta Platforms, Inc. and affiliates.
  */
 
+#define _GNU_SOURCE
```
qdeslandes: Not needed anymore.

src/bpfilter/CMakeLists.txt
agelun: done

Issue #389, opened 2026-01-22
Restore rules counters on chain update


qdeslandes:
When updating a chain, bpfilter will discard the previous packet counters and create new ones. We should be able to restore the counters to their previous values.

We will have to match rules from the old chain to the new one. Design is required.

PR #416, opened 2026-02-11
Restore counter values on chain update

yaakov-stein:
As mentioned in #389, when `bf_cgen_update()` replaces a chain, the new BPF program starts with a zeroed counter map. The behavior should be that when updating a specific entity within a chain (set, rule, etc.), the counters should be preserved (see @qdeslandes comment [below](https://github.com/facebook/bpfilter/pull/416#pullrequestreview-3835437432) for more details and rationale).

This PR adds that behavior specifically for `chain update-set`, while leaving the existing behavior for `chain update`.

Notes/Thoughts:
- With the current implementation, counters are copied from the old map to the new map after the new program is loaded but before the link swap. This means that it is possible for some packets to be unaccounted for in the time between the data being transferred and when the new chain begins filtering. While it is possible to keep the same maps (and thus the exact number) since we know the number of rules will be the same, doing that would require a more invasive approach (the flag would need to propagate into more methods) and will not work for `add-rule` or `remove-rule`.
- A `bf_cgen_update_flags` enum is added for flexibility in case we need future arguments as opposed to a boolean (which is simpler but harder to work with if we need to use or update this in the future). However, the goal either way should be for this to be temporary (in line with the consolidated behavior mentioned [here](https://github.com/facebook/bpfilter/pull/416#pullrequestreview-3835437432)).

~## Design~
~The implementation assumes:~
- ~Chains may have large numbers of rules, sets, and set elements, but rules won't have exceedingly large numbers of matchers~
- ~Update overhead should remain negligible relative to program generation and loading~
- ~Simplicity is preferred over hyper-optimizations where possible~

~**Set mapping.** Because set matchers reference sets by position and positions can change between chains, a set index mapping is built first. To efficiently compare large sets, old sets are hashed (FNV-1a with commutative element sum for order-independence), sorted by hash, and binary-searched for each new set. Hash matches are verified deterministically with `bf_set_equal()`, which copies each set's elements into a flat array, sorts them, and compares with memcmp.~

~**Rule matching.** Rules are hashed the same way - FNV-1a of scalar fields plus a commutative sum of per-matcher hashes. A new rules set matchers hash their mapped old index so equivalent rules produce identical hashes regardless of set position (this abstracts away the whole idea of the set so we don't need to worry about it). Old rules are sorted by hash and binary-searched for each new rule, with full order-independent matcher comparison to verify matches. Once again, hash matches are verified deterministically using `_bf_rule_equal`.~

~**Counter transfer.** For each matched pair, the old counter is read and written to the new program's counter map. Policy and error counters are always preserved. This happens after `bf_program_load` but before `bf_link_update`, so there is no race.~

~## Performance
Tested with up to 10k rules and 100 sets of 50k elements each. The mapping and transfer overhead was negligible relative to program generation and loading.~

## Test plan 
- Added tests to `chain_update.sh` and `chain_update_set.sh`
- Manual testing 
- ~Unit tests for `bf_fnv1a` and `bf_set_equal`~
- ~E2E test covering rule reordering and set index shuffling~

### DIFF
--- a/src/bpfilter/cgen/cgen.c
+++ b/src/bpfilter/cgen/cgen.c
@@ -27,6 +27,7 @@
 #include "cgen/dump.h"
 #include "cgen/handle.h"
 #include "cgen/prog/link.h"
+#include "cgen/prog/map.h"
 #include "cgen/program.h"
 #include "ctx.h"
 #include "opts.h"
@@ -334,7 +335,51 @@ int bf_cgen_attach(struct bf_cgen *cgen, const struct bf_ns *ns,
     return r;
 }
 
-int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain)
+/**
+ * @brief Transfer all counters from old handle to new handle.
+ *
+ * Copies counter values 1:1 for all rule counters plus policy and error
+ * counters. The old and new chains must have the same number of rules.
+ * Both handles must be loaded.
+ *
+ * @param old_handle Handle with the source counter map. Can't be NULL.
+ * @param new_handle Handle with the destination counter map. Can't be NULL.
+ * @param n_rules Number of rules in the chain.
+ * @return 0 on success, or a negative errno value on failure.
+ */
+static int _bf_cgen_transfer_counters(const struct bf_handle *old_handle,
+                                      struct bf_handle *new_handle,
+                                      size_t n_rules)
+{
+    int r;
+
+    assert(old_handle);
+    assert(new_handle);
+
+    if (!old_handle->cmap || !new_handle->cmap)
+        return bf_err_r(-ENOENT, "missing counter map for counter transfer");
+
+    // n_rules entries for rules, +1 for policy, +1 for errors.
+    for (uint32_t i = 0; i < n_rules + 2; ++i) {
+        struct bf_counter counter;
+
+        r = bf_handle_get_counter(old_handle, i, &counter);
+        if (r)
+            return bf_err_r(r, "failed to read counter %u", i);
+
+        if (!counter.count && !counter.size)
+            continue;
+
+        r = bf_map_set_elem(new_handle->cmap, &i, &counter);
+        if (r)
+            return bf_err_r(r, "failed to write counter %u", i);
+    }
+
+    return 0;
+}
+
+int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain,
+                   uint32_t flags)
 {
     _free_bf_program_ struct bf_program *new_prog = NULL;
     _free_bf_handle_ struct bf_handle *new_handle = NULL;
@@ -344,6 +389,9 @@ int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain)
 
     assert(cgen);
     assert(new_chain);
+
+    if (flags & ~BF_FLAGS_MASK(_BF_CGEN_UPDATE_MAX))
+        return bf_err_r(-EINVAL, "unknown update flags: 0x%x", flags);
 
     old_handle = cgen->handle;
 
@@ -370,6 +418,19 @@ int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain)
     r = bf_program_load(new_prog);
     if (r)
         return bf_err_r(r, "failed to load new program");
+
+    if (flags & BF_FLAG(BF_CGEN_UPDATE_PRESERVE_COUNTERS)) {
+        if (bf_list_size(&cgen->chain->rules) !=
+            bf_list_size(&(*new_chain)->rules)) {
+            return bf_err_r(-EINVAL,
+                            "rule count mismatch for counter transfer");
+        }
+
+        r = _bf_cgen_transfer_counters(old_handle, new_handle,
+                                       bf_list_size(&(*new_chain)->rules));
+        if (r)
+            return bf_err_r(r, "failed to transfer counters");
+    }
 
     if (bf_opts_persist())
         bf_handle_unpin(old_handle, pindir_fd);
--- a/src/bpfilter/cgen/cgen.h
+++ b/src/bpfilter/cgen/cgen.h
@@ -130,6 +130,17 @@ int bf_cgen_attach(struct bf_cgen *cgen, const struct bf_ns *ns,
                    struct bf_hookopts **hookopts);
 
 /**
+ * Flags to control the behavior of `bf_cgen_update`.
+ */
+enum bf_cgen_update_flags
+{
+    /** Preserve counter values from the old program. */
+    BF_CGEN_UPDATE_PRESERVE_COUNTERS,
+
+    _BF_CGEN_UPDATE_MAX,
+};
+
+/**
  * Update the program attached to the hook.
  *
  * A new program will be generated based on `new_chain`, before it is loaded
@@ -142,9 +153,11 @@ int bf_cgen_attach(struct bf_cgen *cgen, const struct bf_ns *ns,
  * @param cgen Codegen to update. Can't be NULL.
  * @param new_chain Chain containing the new rules, sets, and policy.
  *        Can't be NULL.
+ * @param flags Flags to control update behavior. 0 if no flags.
  * @return 0 on success, or negative errno value on failure.
  */
-int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain);
+int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain,
+                   uint32_t flags);
 
 /**
  * Detach a program from the kernel.
--- a/src/libbpfilter/include/bpfilter/helper.h
+++ b/src/libbpfilter/include/bpfilter/helper.h
@@ -116,6 +116,13 @@ extern const char *strerrordesc_np(int errnum);
  */
 #define BF_FLAG(n) (1 << (n))
 
+/**
+ * @brief Generate a bitmask covering all valid flags below `max_value`.
+ *
+ * @return Mask with bits 0 through max_value-1 set.
+ */
+#define BF_FLAGS_MASK(max_value) ((1ULL << (max_value)) - 1)
+
 #define bf_packed __attribute__((packed))
 #define bf_aligned(x) __attribute__((aligned(x)))
 #define bf_unused __attribute__((unused))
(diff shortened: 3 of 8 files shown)


tests/e2e/cli/chain_update.sh
```diff
@@ -23,3 +23,17 @@ ${FROM_NS} bfcli chain update --from-str "chain chain_load_xdp_3 BF_HOOK_XDP ACC
 ${FROM_NS} bfcli chain update --name chain_load_xdp_3 --from-str "chain chain_load_xdp_3 BF_HOOK_XDP{ifindex=${NS_IFINDEX}} ACCEPT"
 ping -c 1 -W 0.1 ${NS_IP_ADDR}
 ${FROM_NS} bfcli chain flush --name chain_load_xdp_3
+
+# Counters are reset after chain update
+${FROM_NS} bfcli chain set --from-str "chain chain_load_xdp_4 BF_HOOK_XDP{ifindex=${NS_IFINDEX}} ACCEPT
+    rule ip4.proto icmp counter DROP
+"
+ping -c 1 -W 0.1 ${NS_IP_ADDR} || true
+counter=$(${FROM_NS} bfcli chain get --name chain_load_xdp_4 | awk '/^        counters [0-9]+ packets/{print $2}')
```
qdeslandes: Use `bpftool map dump pinned $PATH | jq` to get the counter value.

tests/e2e/cli/chain_update.sh
```diff
@@ -23,3 +23,17 @@ ${FROM_NS} bfcli chain update --from-str "chain chain_load_xdp_3 BF_HOOK_XDP ACC
...
+test "$counter" = "1"
+${FROM_NS} bfcli chain update --from-str "chain chain_load_xdp_4 BF_HOOK_XDP ACCEPT
+    rule ip4.proto icmp counter DROP
+"
+counter=$(${FROM_NS} bfcli chain get --name chain_load_xdp_4 | awk '/^        counters [0-9]+ packets/{print $2}')
```
qdeslandes: Use `bpftool map dump pinned $PATH | jq` to get the counter value.

tests/e2e/cli/chain_update_set.sh
```diff
@@ -202,3 +202,25 @@ chain_output=$(${FROM_NS} bfcli chain get --name test_xdp)
...
+    set blocked_ips (ip4.saddr) in {
+        10.0.0.1
+    }
+    rule
+        ip4.proto icmp
+        counter
+        DROP
+"
+ping -c 1 -W 0.1 ${NS_IP_ADDR} || true
+counter=$(${FROM_NS} bfcli chain get --name test_xdp | awk '/^        counters [0-9]+ packets/{print $2}')
```
qdeslandes: Use `bpftool map dump pinned $PATH | jq` to get the counter value.

tests/e2e/cli/chain_update_set.sh
```diff
@@ -202,3 +202,25 @@ chain_output=$(${FROM_NS} bfcli chain get --name test_xdp)
...
+test "$counter" = "1"
+
+${FROM_NS} bfcli chain update-set \
+    --name test_xdp \
+    --set-name blocked_ips \
+    --add 10.0.0.2
+
+counter=$(${FROM_NS} bfcli chain get --name test_xdp | awk '/^        counters [0-9]+ packets/{print $2}')
```
qdeslandes: Use `bpftool map dump pinned $PATH | jq` to get the counter value.

src/bpfilter/cgen/cgen.c
```diff
@@ -334,7 +336,50 @@ int bf_cgen_attach(struct bf_cgen *cgen, const struct bf_ns *ns,
...
 
-int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain)
+/**
+ * @brief Transfer all counters from old handle to new handle.
+ *
+ * Copies counter values 1:1 for all rule counters plus policy and error
+ * counters. The old and new chains must have the same number of rules.
+ *
+ * @param old_handle Handle with the source counter map. Can't be NULL.
+ * @param new_handle Handle with the destination counter map. Can't be NULL.
```
pzmarzly, line 346: Nit: Maybe add a comment that both old and new handle must be loaded (to have bf_maps initialized).

tests/e2e/cli/chain_update_set.sh
```diff
@@ -202,3 +202,25 @@ chain_output=$(${FROM_NS} bfcli chain get --name test_xdp)
...
+    set blocked_ips (ip4.saddr) in {
+        10.0.0.1
+    }
+    rule
+        ip4.proto icmp
+        counter
+        DROP
+"
+ping -c 1 -W 0.1 ${NS_IP_ADDR} || true
+counter=$(${FROM_NS} bpftool map dump pinned ${WORKDIR}/bpf/bpfilter/test_xdp/bf_cmap | jq '.[0].value.packets')
```
yaakov-stein: Note: this does now rely on BTF

src/bpfilter/cgen/cgen.c
```diff
@@ -334,7 +336,51 @@ int bf_cgen_attach(struct bf_cgen *cgen, const struct bf_ns *ns,
...
+    assert(old_handle);
+    assert(new_handle);
+
+    if (!old_handle->cmap || !new_handle->cmap)
+        return bf_err_r(-ENOENT, "missing counter map for counter transfer");
+
+    // n_rules entries for rules, +1 for policy, +1 for errors.
+    for (uint32_t i = 0; i < n_rules + 2; ++i) {
+        struct bf_counter counter;
+
```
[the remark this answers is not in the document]
yaakov-stein: Given that `bf_handle_set_counter` would only be used for this case as of now, and this use case may change soon, I think it would make sense to include this change when there is a stronger or more permanent use case.

src/bpfilter/cgen/cgen.c
```diff
@@ -371,6 +417,17 @@ int bf_cgen_update(struct bf_cgen *cgen, struct bf_chain **new_chain)
...
 
+    if (flags & BF_FLAG(BF_CGEN_UPDATE_PRESERVE_COUNTERS)) {
+        if (bf_list_size(&cgen->chain->rules) !=
+            bf_list_size(&(*new_chain)->rules))
+            return bf_err_r(-EINVAL, "rule count mismatch for counter transfer");
+
+        r = _bf_cgen_transfer_counters(old_handle, new_handle,
+                                       bf_list_size(&(*new_chain)->rules));
+        if (r)
+            return bf_err_r(r, "failed to transfer counters");
```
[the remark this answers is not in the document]
yaakov-stein: I don't think this is necessary. This case should be impossible and the additional check adds a bit of clutter. The check is more (but not exclusively) so we can easily catch code changes that break this assumption.

src/bpfilter/cgen/cgen.c
```diff
@@ -334,7 +335,51 @@ int bf_cgen_attach(struct bf_cgen *cgen, const struct bf_ns *ns,
...
+    assert(old_handle);
+    assert(new_handle);
+
+    if (!old_handle->cmap || !new_handle->cmap)
+        return bf_err_r(-ENOENT, "missing counter map for counter transfer");
+
+    // n_rules entries for rules, +1 for policy, +1 for errors.
+    for (uint32_t i = 0; i < n_rules + 2; ++i) {
+        struct bf_counter counter;
+
```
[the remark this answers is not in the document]
yaakov-stein: Needing to add a narrowing cast doesn't seem like a better approach here. The current approach seems fine to me

src/bpfilter/cgen/cgen.h
```diff
@@ -129,6 +129,17 @@ int bf_cgen_load(struct bf_cgen *cgen);
 int bf_cgen_attach(struct bf_cgen *cgen, const struct bf_ns *ns,
                    struct bf_hookopts **hookopts);
 
+/**
+ * Flags to control the behavior of @ref bf_cgen_update .
```
qdeslandes: `@ref bf_cgen_update .` ->
```
`bf_cgen_update`.
```

src/bpfilter/cgen/cgen.c
qdeslandes, line 392: Ensure only existing flags are used:

Probably in `helper.c`:
```
#define BF_FLAGS_MASK(max_value) (1U << (max_value)) - 1)
```

Then:
```
if (flags & ~BF_FLAGS_MASK(_BF_CGEN_UPDATE_MAX))
    return bf_err_r(-EINVAL, ...);
```

tests/e2e/cli/chain_update.sh
```diff
@@ -23,3 +23,17 @@ ${FROM_NS} bfcli chain update --from-str "chain chain_load_xdp_3 BF_HOOK_XDP ACC
 ${FROM_NS} bfcli chain update --name chain_load_xdp_3 --from-str "chain chain_load_xdp_3 BF_HOOK_XDP{ifindex=${NS_IFINDEX}} ACCEPT"
 ping -c 1 -W 0.1 ${NS_IP_ADDR}
 ${FROM_NS} bfcli chain flush --name chain_load_xdp_3
+
+# Counters are reset after chain update
+${FROM_NS} bfcli chain set --from-str "chain chain_load_xdp_4 BF_HOOK_XDP{ifindex=${NS_IFINDEX}} ACCEPT
+    rule ip4.proto icmp counter DROP
+"
+ping -c 1 -W 0.1 ${NS_IP_ADDR} || true
```
qdeslandes: `(! ping -c 1 -W 0.1 ${NS_IP_ADDR})`, that way we ensure the ping fails, instead of blindly accepting anything.

tests/e2e/cli/chain_update_set.sh
```diff
@@ -202,3 +202,25 @@ chain_output=$(${FROM_NS} bfcli chain get --name test_xdp)
...
+${FROM_NS} bfcli chain set --from-str "chain test_xdp BF_HOOK_XDP{ifindex=${NS_IFINDEX}} ACCEPT
+    set blocked_ips (ip4.saddr) in {
+        10.0.0.1
+    }
+    rule
+        ip4.proto icmp
+        counter
+        DROP
+"
+ping -c 1 -W 0.1 ${NS_IP_ADDR} || true
```
qdeslandes: Same here.

tests/e2e/cli/chain_update_set.sh
qdeslandes, line 1: It would be worth also ensure counters are preserved for non-set rules.

src/libbpfilter/include/bpfilter/helper.h
```diff
@@ -116,6 +116,13 @@ extern const char *strerrordesc_np(int errnum);
  */
 #define BF_FLAG(n) (1 << (n))
 
+/**
+ * @brief Generate a bitmask covering all valid flags below `max_value`.
+ *
+ * @return Mask with bits 0 through max_value-1 set.
```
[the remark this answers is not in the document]
yaakov-stein: This is fixed in #437. If we want to be extra careful, we can wait until that gets merged before merging this.

src/libbpfilter/include/bpfilter/helper.h
```diff
@@ -116,6 +116,13 @@ extern const char *strerrordesc_np(int errnum);
...
+ */
+#define BF_FLAGS_MASK(max_value) ((1ULL << (max_value)) - 1)
```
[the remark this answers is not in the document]
yaakov-stein: See [above comment](https://github.com/facebook/bpfilter/pull/416#discussion_r2861832376)

PR #419, opened 2026-02-14
Prevent serialization of `bf_program` object

qdeslandes:
`bf_program` is used to generate the BPF bytecode for every chain: when a chain needs to be created, a new `bf_program` object is created and is used to contain the generation context. Once the chain is loaded, this object becomes irrelevant.

This change moves the references to the BPF objects into a new `bf_handle` object: when the program is loaded, all the maps and the program FD are stored in `bf_program.handle`, `bf_cgen` can then steal the pointer to `bf_handle` and discard the `bf_program` object.

### DIFF
--- a/src/bpfilter/cgen/prog/map.c
+++ b/src/bpfilter/cgen/prog/map.c
@@ -73,7 +73,7 @@ static int _bf_btf_load(struct bf_btf *btf)
     if (!raw)
         return bf_err_r(errno, "failed to request BTF raw data");
 
-    r = bf_bpf_btf_load(raw, bf_ctx_token());
+    r = bf_bpf_btf_load(raw, attr.btf_size, bf_ctx_token());
     if (r < 0)
         return r;
 
--- a/src/libbpfilter/bpf.c
+++ b/src/libbpfilter/bpf.c
@@ -176,7 +176,7 @@ int bf_bpf_token_create(int bpffs_fd)
     return bf_bpf(BF_BPF_TOKEN_CREATE, &attr);
 }
 
-int bf_bpf_btf_load(const void *btf_data, int token_fd)
+int bf_bpf_btf_load(const void *btf_data, size_t btf_data_len, int token_fd)
 {
     assert(btf_data);
 
@@ -185,6 +185,7 @@ int bf_bpf_btf_load(const void *btf_data, int token_fd)
     memset(&attr, 0, sizeof(attr));
 
     attr.btf = bf_ptr_to_u64(btf_data);
+    attr.btf_size = btf_data_len;
 
     if (token_fd != -1) {
         attr.btf_token_fd = token_fd;
--- a/src/libbpfilter/include/bpfilter/bpf.h
+++ b/src/libbpfilter/include/bpfilter/bpf.h
@@ -111,12 +111,13 @@ int bf_bpf_token_create(int bpffs_fd);
  * @brief Load BTF data into the kernel.
  *
  * @param btf_data Raw BTF data to send to the kernel. Can't be NULL.
+ * @param btf_data_len Length of `btf_data`.
  * @param token_fd File descriptor of the BPF token to use, or -1 if no token
  *        should be used.
  * @return A valid BTF file descriptor on success (owned by the caller),
  *         or a negative error value on failure.
  */
-int bf_bpf_btf_load(const void *btf_data, int token_fd);
+int bf_bpf_btf_load(const void *btf_data, size_t btf_data_len, int token_fd);
 
 /**
  * @brief Create a new BPF map.
--- a/tests/unit/libbpfilter/bpf.c
+++ b/tests/unit/libbpfilter/bpf.c
@@ -289,7 +289,7 @@ static void bpf_btf_load_success(void **state)
     (void)mock;
 
     bft_mock_syscall_set_retval(25);
-    r = bf_bpf_btf_load(btf_data, -1);
+    r = bf_bpf_btf_load(btf_data, 1, -1);
     assert_int_equal(r, 25);
 }
 
@@ -304,7 +304,7 @@ static void bpf_btf_load_with_token(void **state)
     (void)mock;
 
     bft_mock_syscall_set_retval(26);
-    r = bf_bpf_btf_load(btf_data, 5);
+    r = bf_bpf_btf_load(btf_data, 1, 5);
     assert_int_equal(r, 26);
 }
 


src/bpfilter/cgen/prog/link.c
```diff
@@ -24,23 +24,82 @@
...
 
     assert(link);
-    assert(name);
-    assert(name[0] != '\0');
+    assert(hookopts);
 
-    _link = malloc(sizeof(*_link));
+    if (!*hookopts) {
+        return bf_err_r(-EINVAL,
+                        "hookopts are required when created a new bf_link");
```
yaakov-stein, line 44: nit: "when creating"

src/bpfilter/cgen/prog/link.h
```diff
@@ -17,10 +17,8 @@
...
 struct bf_link
 {
-    /** Name of the link. From the kernel perspective, BPF link objects don't
-     * have a name like the programs, but this name will be used to pin the
-     * link. */
-    char name[BPF_OBJ_NAME_LEN];
+    /** Hook to which the link is attached. */
+    enum bf_hook hook;
 
     /** Hook options used for the link. Only valid if the link is materialized */
```
yaakov-stein, line 23: nit: I believe this comment should be updated now that hookopts should always be valid

src/bpfilter/cgen/prog/link.h
```diff
@@ -44,10 +42,14 @@ struct bf_link
  *
  * @param link `bf_link` object to allocate and initialize. On failure,
  *        this parameter is unchanged. Can't be NULL.
- * @param name Name of the link. Can't be empty or NULL.
+ * @param hook Hook to attach the link to.
```
yaakov-stein, line 45: Not sure how to let github let me comment on unchanged lines above, but I believe line 40/41: 
```
@note This function won't create a new BPF link, but a bpfilter-specific
object used to keep track of a BPF link on the system.
 ```
is no longer accurate (this does now create the new BPF link as well as the bpfilter-specific link).

File: src/bpfilter/cgen/prog/link.c [hunk 4]
```diff
@@ -51,39 +110,34 @@ int bf_link_new_from_pack(struct bf_link **link, int dir_fd,
...
-    }
+    r = bf_hookopts_new_from_pack(&_link->hookopts, child);
+    if (r)
+        return r;
 
-    if (dir_fd != -1) {
-        r = bf_bpf_obj_get("bf_link", dir_fd, &_link->fd);
-        if (r)
-            return bf_err_r(r, "failed to open pinned BPF link 'bf_link'");
+    r = bf_bpf_obj_get("bf_link", dir_fd, &_link->fd);
```
yaakov-stein, line 133: nit: this removes the `dir_fd != -1` check (which make sense given the changes) but doesn't update the method's docs (link.h 59/60)

src/bpfilter/cgen/prog/link.c
```diff
@@ -24,23 +24,82 @@
 #include <bpfilter/logger.h>
 #include <bpfilter/pack.h>
 
-int bf_link_new(struct bf_link **link, const char *name)
+static int _bf_link_try_attach_xdp(int prog_fd, int ifindex, enum bf_hook hook);
```
yaakov-stein, line 27: Would it be simpler and less duplicate code if we just move this entire method here?

tests/e2e/daemon/restore_with_sets.sh
```diff
@@ -0,0 +1,19 @@
...
+
+start_bpfilter
+    ${FROM_NS} bfcli chain set --from-str "chain test BF_HOOK_XDP{ifindex=${NS_IFINDEX}} ACCEPT
+        set myset (ip4.saddr) in { 192.168.1.1; 192.168.1.2 }
+        rule (ip4.saddr) in myset counter DROP"
+stop_bpfilter --skip-cleanup
+
+start_bpfilter
+    # Verify chain with sets is restored properly
+    ${FROM_NS} bfcli chain get --name test
```
yaakov-stein, line 18: Should we verify the set is still there and with the same elements? It seems that all we're verifying is that the chain still exists. (ie we can add a ping and ensure it's dropped or something like that.)

tests/e2e/daemon/restore_with_sets.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -eux
+set -o pipefail
```
yaakov-stein, line 4: I believe these are no longer needed as of #417

src/bpfilter/cgen/prog/map.c
```diff
@@ -24,20 +24,141 @@
 
 #include "ctx.h"
 
+#define _free_bf_btf_ __attribute__((__cleanup__(_bf_btf_free)))
+
+static void _bf_btf_free(struct bf_btf **btf);
```
yaakov-stein, line 29: Similar to the forward declaration comment in `bf_link` - is the forward declaration beneficial as opposed to just moving the method here?

src/bpfilter/cgen/prog/map.c
```diff
@@ -127,14 +264,9 @@ int bf_map_new_from_pack(struct bf_map **map, int dir_fd, bf_rpack_node_t node)
...
 
-    if (dir_fd != -1) {
-        r = bf_bpf_obj_get(_map->name, dir_fd, &_map->fd);
-        if (r < 0) {
-            return bf_err_r(r, "failed to open pinned BPF map '%s'",
-                            _map->name);
-        }
-    } else {
-        _map->fd = -1;
+    r = bf_bpf_obj_get(_map->name, dir_fd, &_map->fd);
```
yaakov-stein, line 267: Same comment regarding stale docs as `dir_fd == -1` in `bf_link` (`bf_map_new_from_pack` docs in `bf_map.h` L84/85 still say `dir_fd` can equal -1).

src/bpfilter/cgen/prog/map.c
```diff
@@ -127,14 +264,9 @@ int bf_map_new_from_pack(struct bf_map **map, int dir_fd, bf_rpack_node_t node)
...
+    if (r < 0) {
```
yaakov-stein, line 268: nit: unnecessary brackets

src/bpfilter/cgen/prog/map.c
```diff
@@ -24,20 +24,141 @@
...
+        btf__add_int(kbtf, "u64", 8, 0);
+        btf->key_type_id = btf__add_int(kbtf, "u32", 4, 0);
+        btf->value_type_id = btf__add_struct(kbtf, "bf_counters", 16);
+        btf__add_field(kbtf, "packets", 1, 0, 0);
+        btf__add_field(kbtf, "bytes", 1, 64, 0);
+        break;
+    case BF_MAP_TYPE_PRINTER:
+    case BF_MAP_TYPE_SET:
+    case BF_MAP_TYPE_LOG:
+        // No BTF data available for this map types
```
yaakov-stein, line 118: Just wondering, are we going to have an issue recovering these types now that we are more heavily relying on BTF? Asking specifically regarding when we move to a daemonless model

src/bpfilter/cgen/prog/map.c
```diff
@@ -52,6 +173,15 @@ static int _bf_map_new(struct bf_map **map, const char *name,
 
     bf_strncpy(_map->name, BPF_OBJ_NAME_LEN, name);
 
+    btf = _bf_map_make_btf(_map);
+
+    fd =
+        bf_bpf_map_create(_map->name, _map->bpf_type, _map->key_size,
```
yaakov-stein, line 179: Previously, there was a note mentioning that btf was not mandatory. It seems like that behavior is still the same (L127 uses `bf_warn_r`), but we've removed the note. Just double checking that was intentional

src/bpfilter/cgen/prog/map.c
```diff
@@ -210,176 +342,12 @@ void bf_map_dump(const struct bf_map *map, prefix_t *prefix)
     DUMP(prefix, "key_size: %lu", map->key_size);
     DUMP(prefix, "value_size: %lu", map->value_size);
 
-    if (map->n_elems == BF_MAP_N_ELEMS_UNKNOWN) {
-        DUMP(prefix, "n_elems: unknown");
-    } else {
-        DUMP(prefix, "n_elems: %lu", map->n_elems);
```
yaakov-stein, line 216: While we don't need the check anymore, should we keep `DUMP(prefix, "n_elems: %lu", map->n_elems);` as part of the dump? That part still seems useful.

src/bpfilter/cgen/program.c
```diff
@@ -1063,19 +1022,19 @@ int bf_program_load(struct bf_program *prog)
 
     r = _bf_program_load_sets_maps(prog);
     if (r)
-        return r;
+        return bf_err_r(r, "failed to load the sets map");
 
     r = _bf_program_load_counters_map(prog);
     if (r)
-        return r;
+        return bf_err_r(r, "failed to load the counter map");
```
yaakov-stein, line 888: nit: Now that we add these `bf_err_r`, we're logging the error both in `bf_program_load` as well as in the `_bf_program_load_*_map` methods. The error logging granularity could be helpful, but I could also see that being unnecessary. Just wanted to flag, don't have an opinion either way.

src/bpfilter/cgen/program.c
```diff
@@ -1127,18 +1086,16 @@ void bf_program_detach(struct bf_program *prog)
 
 void bf_program_unload(struct bf_program *prog)
 {
+    _clean_bf_list_ bf_list list = bf_list_default_from(prog->sets);
```
yaakov-stein: `prog->sets` accesses prog before `assert(prog)` (L1091)

src/bpfilter/cgen/program.c
```diff
@@ -1028,28 +857,19 @@ static int _bf_program_load_sets_maps(struct bf_program *new_prog)
...
+        if (r)
+            return bf_err_r(r, "failed to add set elements to the map");
 
-        set_node = bf_list_node_next(set_node);
-        map_node = bf_list_node_next(map_node);
-    }
+        r = bf_list_push(&new_prog->handle->sets, (void **)&map);
+        if (r)
+            return r;
+    };
```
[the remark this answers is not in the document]
yaakov-stein: +1 to this copilot comment

src/bpfilter/cgen/program.c
```diff
@@ -288,17 +253,29 @@ int bf_program_pack(const struct bf_program *program, bf_wpack_t *pack)
     assert(program);
     assert(pack);
 
-    bf_wpack_open_object(pack, "cmap");
-    bf_map_pack(program->cmap, pack);
-    bf_wpack_close_object(pack);
+    if (program->cmap) {
```
yaakov-stein: For L256, L264, L311, L320, L1140, L1148, L1196, L1198 - while an optimization was added for the log map and it is possible it may not be created, it appears that the counter map and printer map _are_ always created. Do we need these null checks?

src/bpfilter/cgen/program.c
```diff
@@ -1127,18 +1086,16 @@ void bf_program_detach(struct bf_program *prog)
...
-    bf_map_destroy(prog->lmap);
-    bf_list_foreach (&prog->sets, map_node) {
-        struct bf_map *map = bf_list_node_get_data(map_node);
-        if (map)
-            bf_map_destroy(map);
-    }
+    bf_map_free(&prog->cmap);
+    bf_map_free(&prog->pmap);
+    bf_map_free(&prog->lmap);
+    bf_swap(list, prog->sets);
```
yaakov-stein: This is a pretty cool trick!

src/bpfilter/cgen/program.c
```diff
@@ -105,14 +106,15 @@ int bf_program_new(struct bf_program **program, const struct bf_chain *chain)
     if (!_program)
         return -ENOMEM;
 
-    (void)snprintf(_program->prog_name, BPF_OBJ_NAME_LEN, "%s", _BF_PROG_NAME);
```
yaakov-stein: We keep `prog_name` in the `bf_program` struct but remove the initialization. `program->prog_name` is still used (for example in `bf_program_dump` on L217 and in `bf_program_load` on 910) and is UB.

src/bpfilter/cgen/handle.c
```diff
@@ -0,0 +1,384 @@
...
+    assert(handle);
+
+    bf_link_free(&handle->link);
+}
+
+void bf_handle_unload(struct bf_handle *handle)
+{
+    _clean_bf_list_ bf_list list = bf_list_default_from(handle->sets);
+
+    assert(handle);
```
yaakov-stein, line 376: `asset(handle)` after using handle on L374

src/bpfilter/cgen/handle.c
```diff
@@ -0,0 +1,384 @@
...
+            return bf_err_r(r, "failed to unpack bf_map into bf_handle.sets");
+    }
+
+    *handle = TAKE_PTR(_handle);
+
+    return 0;
+}
+
+void bf_handle_free(struct bf_handle **handle)
+{
```
yaakov-stein, line 121: Add an `assert(handle)` here?

src/bpfilter/cgen/program.c
```diff
@@ -230,18 +179,7 @@ void bf_program_free(struct bf_program **program)
     bf_list_clean(&(*program)->fixups);
     free((*program)->img);
 
-    /* Close the file descriptors if they are still open. If --transient is
-     * used, then the file descriptors are already closed (as
-     * bf_program_unload() has been called). Otherwise, bf_program_unload()
-     * won't be called, but the programs are pinned, so they can be closed
-     * safely. */
```
yaakov-stein, line 278: Should we keep this comment but move it to `bf_handle_free`? (IIUC the behavior is still the same)

src/bpfilter/cgen/handle.h
```diff
@@ -0,0 +1,176 @@
...
+ * Closes all file descriptors and frees all owned objects. If the BPF objects
+ * are pinned, they will survive the close.
+ *
+ * @param handle `bf_handle` object to free. If `*handle` is NULL, this function
+ *        has no effect. Can't be NULL.
+ */
+void bf_handle_free(struct bf_handle **handle);
+
+/**
+ * Serialize a bf_handle.
```
yaakov-stein, line 94: `@brief `?

src/bpfilter/cgen/handle.c
```diff
@@ -0,0 +1,384 @@
...
+
+    if (handle->link) {
+        bf_wpack_open_object(pack, "link");
+        bf_link_pack(handle->link, pack);
+        bf_wpack_close_object(pack);
+    } else {
+        bf_wpack_kv_nil(pack, "link");
+    }
+
+    if (handle->cmap) {
```
yaakov-stein, line 152: Similar to a comment from the previous review - `handle->cmap` and `handle->pmap` are guaranteed to be non-null IIUC. Should we remove the checks on L152, L160, L202, L211, L265, L273, L316, L318?

src/bpfilter/cgen/cgen.c
```diff
@@ -288,16 +290,18 @@ int bf_cgen_load(struct bf_cgen *cgen)
...
+        r = bf_handle_pin(handle, pindir_fd);
         if (r)
             return r;
     }
 
     bf_info("load %s", cgen->chain->name);
     bf_cgen_dump(cgen, EMPTY_PREFIX);
 
-    cgen->program = TAKE_PTR(prog);
+    cgen->handle = TAKE_PTR(handle);
```
yaakov-stein, line 304: Because we're only taking the pointer here, the `bf_cgen_dump` on L302 will have stale/null data.

qdeslandes:
Closing in favor of #426

PR #421, opened 2026-02-15
tests: check: fail checkstyle on formatting violations

yaakov-stein:
Relies on commit fc16f93 - previously, .clang-format would run but would return 0 (success) even if there were formatting violations. With this change, the command properly fails when formatting violations exist. (This CI _should_ fail for this PR until fc16f93 is merged.)

Update: This has the CI run clang-format on all source files (`tests/ `+ `src/`) and clang-tidy on `src/` files, failing if any errors come from those. e0450b2 and 294515f are results from running `make -C build fixstyle`.

### DIFF
--- a/doc/developers/contributing.rst
+++ b/doc/developers/contributing.rst
@@ -11,7 +11,6 @@ You are welcome to reach out to qde@naccy.de if you need help, or have any quest
 To do
 -----
 
-* Enforce ``clang-format`` and ``clang-tidy`` on all source files.
 * Remove the RPM ``x86_64`` macro `workaround <https://pagure.io/epel/issue/325>`_ from the Fedora ``bpfilter.spec``.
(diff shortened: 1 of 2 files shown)


doc/developers/contributing.rst
```diff
@@ -11,7 +11,6 @@ You are welcome to reach out to qde@naccy.de if you need help, or have any quest
 To do
 -----
 
-* Enforce ``clang-format`` and ``clang-tidy`` on all source files.
```
qdeslandes, line 14: This is also for test sources. Adding `-Werror` is required, but also running clang-format on test sources.

yaakov-stein: See update in original PR message^

qdeslandes (changes requested):
LGTM, thanks!

PR #432, opened 2026-02-20
tools: update Claude PR reviewer prompt

pzmarzly:
Combining the best parts of https://github.com/anthropics/claude-code-action/blob/0cf5eeec4f908121edd03a81411b55485994f8d3/docs/solutions.md and https://github.com/anthropics/claude-code/blob/main/plugins/code-review/commands/code-review.md to create a better prompt, that we can keep iterating on.

Demo: https://github.com/pzmarzly/demo--claude-bot-reviews/pull/8 (comments 2 and 3).

### DIFF
--- a/.github/workflows/claude-pr-review.yaml
+++ b/.github/workflows/claude-pr-review.yaml
@@ -23,6 +23,7 @@ jobs:
               uses: actions/checkout@v6
 
             - uses: anthropics/claude-code-action@v1
+              if: github.event.pull_request.draft == false
               with:
                   # Anthropic auth
                   # anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-pr-review.yaml
```diff
@@ -21,27 +21,66 @@ jobs:
...
 
                   # Prompt
-                  plugin_marketplaces: "https://github.com/anthropics/claude-code.git"
-                  plugins: "code-review@claude-code-plugins"
                   prompt: |
-                      /code-review:code-review --comment
+                      REPO: ${{ github.repository }}
+                      PR NUMBER: ${{ github.event.pull_request.number }}
+
+                      Please review this pull request.
```
qdeslandes: Is the AI more efficient if we're nice to it?

.github/workflows/claude-pr-review.yaml
```diff
@@ -21,27 +21,66 @@ jobs:
...
                   # anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
                   claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
                   github_token: ${{ secrets.GITHUB_TOKEN }}
 
                   # Prompt
-                  plugin_marketplaces: "https://github.com/anthropics/claude-code.git"
-                  plugins: "code-review@claude-code-plugins"
                   prompt: |
-                      /code-review:code-review --comment
+                      REPO: ${{ github.repository }}
```
qdeslandes, line 35: Shouldn't we leverage `.claude/commands/review-pr.md` here?

qdeslandes (changes requested):

.github/workflows/claude-pr-review.yaml
pzmarzly: Nobody knows. That trick definitely worked in the past, and seems to be still recommended in Claude examples, so why not?

.github/workflows/claude-pr-review.yaml
pzmarzly: Good point! We shouldn't use that command as-is, since it also builds and tests the PR, and we probably don't want to duplicate other CI workflows (also, I'm not sure if we can mix Claude action with Fedora image, or if it assumes Ubuntu). But we can at least tell Claude to read that file. Let me update the PR.

PR #435, opened 2026-02-22
daemon: cgen: move asserts before pointer dereferences

yaakov-stein:
Fixes a few locations where we dereferenced a variable before the assert.


qdeslandes (approved):
Perfect, thanks!

PR #441, opened 2026-02-24
tools: claude: remove restriction for draft PRs

pzmarzly:
I took this `github.event.pull_request.draft` check from the AI review workflow we were using before (`/code-review:code-review`).

But it may be useful to iterate on your PR with the help of Claude review bot, while keeping it hidden from review queue at the same time. AI review can easily add a few iteration loops when making a PR.

Let's remove this one line.

### DIFF
--- a/.github/workflows/claude-pr-review.yaml
+++ b/.github/workflows/claude-pr-review.yaml
@@ -28,7 +28,6 @@ jobs:
               uses: actions/checkout@v6
 
             - uses: anthropics/claude-code-action@v1
-              if: github.event.pull_request.draft == false
               env:
                   ANTHROPIC_BASE_URL: https://openrouter.ai/api
               with:


qdeslandes:
> Claude, can you advise us here? EDIT: apparently the GH Action cannot :/ EDIT2: I ran the question locally:
> 
> ```
> A few things to consider:
>   1. The concurrency group with cancel-in-progress: true already limits cost — only one review runs per PR at a time. Rapid pushes just cancel previous runs.
>   2. The real abuse vector is external contributors, not project members. But the workflow already checks out upstream code (not the PR branch), so the security model is fine. The
>   cost is just API credits for Claude reviews on junk PRs.
>   3. Practical mitigations if abuse becomes a problem:
>     - Add a label-based gate (e.g., only review if a needs-review label is present)
>     - Restrict to known collaborators with if: github.event.pull_request.author_association != 'NONE'
>     - Rate-limit per author
> ```

Alright, also the CI runs on drafts, so why not!

PR #454, opened 2026-03-04
lib,daemon,cli: use float for meta.probability matcher

agelun:
Change the meta.probability matcher payload from uint8_t to float, enabling fractional percentage values (e.g., "33.33%", "0.1%"). This aligns meta.probability with meta.flow_probability which already used float. Consolidate the now-identical parse and print functions between the two matchers.


src/bfcli/lexer.l
qdeslandes: You can merge those together and get rid of `STATE_MATCHER_META_FLOW_PROBA`.

qdeslandes (changes requested):
Small change, otherwise LGTM.

PR #456, opened 2026-03-05
[0/3] Enable flavor specific matching

yaakov-stein:
Stacked PRs:
- -> #456 
- #464 
- #459 
- #465 

Currently, matcher codegen is hardcoded in a flavor-agnostic switch. With this refactor, each flavor controls its own matcher codegen through a callback, allowing a new flavor to intercept matchers and generate context-appropriate bytecode.

The `gen_inline_matcher` callback generalizes the ad-hoc `gen_inline_get_mark` and `gen_inline_get_skb` callbacks. Instead of having dedicated callbacks for each flavor-sensitive value, every flavor controls its full matcher codegen through a single dispatch point. Packet-based flavors delegate most matchers to a shared `bf_matcher_generate_packet()` function (the extracted switch), overriding only the matchers that need flavor-specific context access.

Here is an overview of the architectural changes:

### Current Architecture

```
  bf_program_generate()
  |
  |-- ops->gen_inline_prologue()             // Flavor-specific: parse raw
  |       |                                  // packet into l3_hdr/l4_hdr
  |       |                                  // scratch buffers, set R7/R8
  |       +-- bf_stub_parse_l3_hdr()
  |       +-- bf_stub_parse_l4_hdr()
  |
  |-- for each rule:
  |     _bf_program_generate_rule()
  |     |
  |     |-- for each matcher:
  |     |     switch(matcher_type)           // Hardcoded switch - same for
  |     |       |                            // ALL flavors (flavor-agnostic)
  |     |       +-- META_* --> bf_matcher_generate_meta()
  |     |       |                 |
  |     |       |                 +-- META_MARK:
  |     |       |                 |     ops->gen_inline_get_mark()  <-- Ad-hoc
  |     |       |                 |                                     flavor callback
  |     |       |                 +-- META_FLOW_HASH:
  |     |       |                       ops->gen_inline_get_skb()   <-- Ad-hoc
  |     |       |                                                       flavor callback
  |     |       +-- IP4_*  --> bf_matcher_generate_ip4()    // Reads l3_hdr
  |     |       +-- IP6_*  --> bf_matcher_generate_ip6()    // Reads l3_hdr
  |     |       +-- TCP_*  --> bf_matcher_generate_tcp()    // Reads l4_hdr
  |     |       +-- UDP_*  --> bf_matcher_generate_udp()    // Reads l4_hdr
  |     |       +-- ICMP_* --> bf_matcher_generate_icmp()   // Reads l4_hdr
  |     |       +-- SET    --> bf_matcher_generate_set()    // Reads l3/l4_hdr
  |     |
  |     |-- ops->gen_inline_set_mark()       // Rule action (if mark set)
  |     +-- ops->get_verdict()               // Emit verdict
  |
  |-- ops->gen_inline_epilogue()             // No-op in all 4 flavors
  +-- ops->get_verdict(chain->policy)        // Default policy
```

**bf\_flavor\_ops (7 callbacks):**

```
  gen_inline_prologue   -- Flavor-specific header parsing
  gen_inline_epilogue   -- No-op in all 4 flavors (dead code)
  gen_inline_set_mark   -- Write mark to sk_buff (rule action)
  gen_inline_get_mark   -- Read mark from sk_buff (used by meta.mark matcher only)
  gen_inline_get_skb    -- Get sk_buff pointer (used by meta.flow_hash matcher only)
  gen_inline_redirect   -- Redirect packet to another interface
  get_verdict           -- Convert ACCEPT/DROP to BPF return code
```

**Problem:** There is no simple mechanism for a flavor to provide alternative bytecode for the same matcher type. The only two flavor-aware matchers (`meta.mark` and `meta.flow_hash`) use ad-hoc callbacks rather than a general dispatch mechanism.

### New Architecture (After)

```
  bf_program_generate()
  |
  |-- ops->gen_inline_prologue()             // Unchanged
  |
  |-- for each rule:
  |     _bf_program_generate_rule()
  |     |
  |     |-- for each matcher:
  |     |     ops->gen_inline_matcher(program, matcher)    // NEW: flavor dispatch
  |     |       |
  |     |       +-- [XDP]  bf_matcher_generate_packet() directly
  |     |       |            (no mark/flow_hash support)
  |     |       |
  |     |       +-- [TC]   _bf_tc_gen_inline_matcher()
  |     |       |     +-- META_MARK, META_FLOW_HASH: flavor-specific loading
  |     |       |     +-- default:        bf_matcher_generate_packet()
  |     |       |
  |     |       +-- [NF]   _bf_nf_gen_inline_matcher()
  |     |       |     +-- META_MARK, META_FLOW_HASH: flavor-specific loading
  |     |       |     +-- default:        bf_matcher_generate_packet()
  |     |       |
  |     |       +-- [CGROUP_SKB] _bf_cgroup_gen_inline_matcher()
  |     |       |     +-- META_MARK, META_FLOW_HASH: flavor-specific loading
  |     |       |     +-- default:        bf_matcher_generate_packet()
  |     |
  |     |-- ops->gen_inline_set_mark()       // Unchanged
  |     +-- ops->get_verdict()               // Unchanged
  |
  |-- ops->gen_inline_epilogue()             // Unchanged
  +-- ops->get_verdict(chain->policy)        // Unchanged
```

**bf\_flavor\_ops (6 callbacks):**

```
  gen_inline_prologue   -- KEPT (unchanged)
  gen_inline_epilogue   -- KEPT (unchanged)
  gen_inline_set_mark   -- KEPT (rule action, unchanged)
  gen_inline_redirect   -- KEPT (unchanged)
  get_verdict           -- KEPT (unchanged)
  gen_inline_matcher    -- NEW: flavor-specific matcher dispatch

  gen_inline_get_mark   -- REMOVED (subsumed by gen_inline_matcher)
  gen_inline_get_skb    -- REMOVED (subsumed by gen_inline_matcher)
```

Notes:
- Commit a65aca1 adds a missing load instruction that is moved as part of this PR.
- While this enables flavor-specific matching in a more intuitive way, an even better approach would be a callback-centric approach that let's flavors define their own table of callbacks per matcher. The plan is to add this eventually.

### DIFF
--- a/src/bpfilter/cgen/cgroup_skb.c
+++ b/src/bpfilter/cgen/cgroup_skb.c
@@ -15,9 +15,12 @@
 #include <bpfilter/btf.h>
 #include <bpfilter/flavor.h>
 #include <bpfilter/helper.h>
+#include <bpfilter/matcher.h>
 #include <bpfilter/verdict.h>
 
 #include "cgen/cgen.h"
+#include "cgen/matcher/meta.h"
+#include "cgen/matcher/packet.h"
 #include "cgen/program.h"
 #include "cgen/stub.h"
 #include "cgen/swich.h"
@@ -114,30 +117,35 @@ static int _bf_cgroup_skb_gen_inline_set_mark(struct bf_program *program,
     return 0;
 }
 
-static int _bf_cgroup_skb_gen_inline_get_mark(struct bf_program *program,
-                                              int reg)
+static int _bf_cgroup_skb_gen_inline_matcher(struct bf_program *program,
+                                             const struct bf_matcher *matcher)
 {
-    EMIT(program,
-         BPF_LDX_MEM(BPF_DW, BPF_REG_1, BPF_REG_10, BF_PROG_CTX_OFF(arg)));
-    EMIT(program,
-         BPF_LDX_MEM(BPF_W, reg, BPF_REG_1, offsetof(struct __sk_buff, mark)));
+    assert(program);
+    assert(matcher);
 
-    return 0;
-}
+    switch (bf_matcher_get_type(matcher)) {
+    case BF_MATCHER_META_MARK:
+        EMIT(program,
+             BPF_LDX_MEM(BPF_DW, BPF_REG_1, BPF_REG_10, BF_PROG_CTX_OFF(arg)));
+        EMIT(program, BPF_LDX_MEM(BPF_W, BPF_REG_1, BPF_REG_1,
+                                  offsetof(struct __sk_buff, mark)));
 
-static int _bf_cgroup_skb_gen_inline_get_skb(struct bf_program *program,
-                                             int reg)
-{
-    EMIT(program, BPF_LDX_MEM(BPF_DW, reg, BPF_REG_10, BF_PROG_CTX_OFF(arg)));
+        return bf_matcher_generate_meta_mark_cmp(program, matcher);
+    case BF_MATCHER_META_FLOW_HASH:
+        EMIT(program,
+             BPF_LDX_MEM(BPF_DW, BPF_REG_1, BPF_REG_10, BF_PROG_CTX_OFF(arg)));
 
-    return 0;
+        return bf_matcher_generate_meta_flow_hash_cmp(program, matcher);
+    default:
+        return bf_matcher_generate_packet(program, matcher);
+    }
 }
 
 /**
  * Convert a standard verdict into a return value.
  *
  * @param verdict Verdict to convert. Must be valid.
- * @return TC return code corresponding to the verdict, as an integer.
+ * @return Cgroup return code corresponding to the verdict, as an integer.
  */
 static int _bf_cgroup_skb_get_verdict(enum bf_verdict verdict)
 {
@@ -155,7 +163,6 @@ const struct bf_flavor_ops bf_flavor_ops_cgroup_skb = {
     .gen_inline_prologue = _bf_cgroup_skb_gen_inline_prologue,
     .gen_inline_epilogue = _bf_cgroup_skb_gen_inline_epilogue,
     .gen_inline_set_mark = _bf_cgroup_skb_gen_inline_set_mark,
-    .gen_inline_get_mark = _bf_cgroup_skb_gen_inline_get_mark,
-    .gen_inline_get_skb = _bf_cgroup_skb_gen_inline_get_skb,
     .get_verdict = _bf_cgroup_skb_get_verdict,
+    .gen_inline_matcher = _bf_cgroup_skb_gen_inline_matcher,
 };
--- a/src/bpfilter/cgen/matcher/meta.h
+++ b/src/bpfilter/cgen/matcher/meta.h
@@ -10,3 +10,29 @@ struct bf_program;
 
 int bf_matcher_generate_meta(struct bf_program *program,
                              const struct bf_matcher *matcher);
+
+/**
+ * @brief Generate bytecode to compare a packet mark.
+ *
+ * The mark value must already be loaded into `BPF_REG_1`.
+ *
+ * @param program Program being generated. Can't be NULL.
+ * @param matcher Matcher to generate comparison for. Can't be NULL.
+ * @return 0 on success, negative errno on error.
+ */
+int bf_matcher_generate_meta_mark_cmp(struct bf_program *program,
+                                      const struct bf_matcher *matcher);
+
+/**
+ * @brief Generate bytecode to compute and compare a flow hash.
+ *
+ * The skb pointer must already be loaded into `BPF_REG_1`. Calls
+ * `bpf_get_hash_recalc` on the skb, then compares the result against the
+ * matcher's payload.
+ *
+ * @param program Program being generated. Can't be NULL.
+ * @param matcher Matcher to generate comparison for. Can't be NULL.
+ * @return 0 on success, negative errno on error.
+ */
+int bf_matcher_generate_meta_flow_hash_cmp(struct bf_program *program,
+                                           const struct bf_matcher *matcher);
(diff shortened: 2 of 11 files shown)


src/bpfilter/cgen/matcher/meta.h
```diff
@@ -10,3 +10,31 @@ struct bf_program;
...
+ * against the matcher's payload.
+ *
+ * The caller is responsible for loading the mark value into @c BPF_REG_1
+ * before calling this function.
+ *
+ * @param program Program being generated. Can't be NULL.
+ * @param matcher Matcher to generate comparison for. Can't be NULL.
+ * @return 0 on success, negative errno on error.
+ */
+int bf_matcher_generate_meta_mark(struct bf_program *program,
```
yaakov-stein: I am slightly torn between two approaches here (but I don't think it is too consequential, especially as `bf_matcher_generate_meta_mark` is no longer needed after #464):
1) The approach currently in this PR - make `bf_matcher_generate_meta_mark/flow_hash` public and have the various `gen_inline_matcher` functions (i.e. `_bf_cgroup_skb_gen_inline_matcher`) call them directly. Have `bf_matcher_generate_packet` error out on those meta matchers.
2) We can keep `bf_matcher_generate_meta_mark/flow_hash` private and the `gen_inline_matcher` callbacks `EMIT` the necessary instructions and then call `bf_matcher_generate_packet` -> `bf_matcher_generate_meta` -> `_bf_matcher_generate_meta_mark/flow_hash`. 

Approach 2 would add less to the public API and tries to stay as similar to the current architecture, while approach 1 is more explicit in how it handles flavor-specific matchers. I'm currently leaning towards approach 1, especially given future plans to move to something along the lines of a table for per-matcher callbacks, but am open to thoughts. Thanks!

src/bpfilter/cgen/cgroup_skb.c
```diff
@@ -114,30 +117,33 @@ static int _bf_cgroup_skb_gen_inline_set_mark(struct bf_program *program,
...
+        return bf_matcher_generate_packet(program, matcher);
+    }
 }
 
 /**
  * Convert a standard verdict into a return value.
  *
  * @param verdict Verdict to convert. Must be valid.
- * @return TC return code corresponding to the verdict, as an integer.
+ * @return Cgroup return code corresponding to the verdict, as an integer.
```
yaakov-stein, line 148: Drive-by fix here and same thing in `nf.c`

src/libbpfilter/include/bpfilter/flavor.h
```diff
@@ -117,6 +116,16 @@ struct bf_flavor_ops
...
+    /**
+     * @brief Generate bytecode for a matcher. Required for all flavors.
+     *
+     * @param program Program being generated. Can't be NULL.
+     * @param matcher Matcher to generate code for. Can't be NULL.
+     * @return 0 on success, negative errno on error.
+     */
+    int (*gen_inline_matcher)(struct bf_program *program,
+                              const struct bf_matcher *matcher);
 };
```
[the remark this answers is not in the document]
yaakov-stein: Beyond the scope of this PR

src/bpfilter/cgen/matcher/meta.h
qdeslandes: The issue I have with the current suggestion is that `bf_matcher_generate_meta_mark` and `bf_matcher_generate_meta_flow_hash` expects the packet's mark to be in `r1` already. It should be self contained and not expect specific register values (except for r6 to r9 registers which are callee saved).

src/bpfilter/cgen/matcher/meta.h
```diff
@@ -10,3 +10,29 @@ struct bf_program;
 
 int bf_matcher_generate_meta(struct bf_program *program,
                              const struct bf_matcher *matcher);
+
+/**
+ * @brief Generate bytecode to compare a packet mark.
+ *
+ * The mark value must already be loaded into @c BPF_REG_1.
```
qdeslandes: `@c BPF_REG_1` ->
```
`BPF_REG_1`
```

Doxygen supports markdown, so we should use it when possible.

src/bpfilter/cgen/matcher/meta.h
```diff
@@ -10,3 +10,29 @@ struct bf_program;
...
+ * @param matcher Matcher to generate comparison for. Can't be NULL.
+ * @return 0 on success, negative errno on error.
+ */
+int bf_matcher_generate_meta_mark(struct bf_program *program,
+                                  const struct bf_matcher *matcher);
+
+/**
+ * @brief Generate bytecode to compute and compare a flow hash.
+ *
+ * The skb pointer must already be loaded into @c BPF_REG_1. Calls
```
qdeslandes: Backticks

src/bpfilter/cgen/matcher/meta.h
```diff
@@ -10,3 +10,29 @@ struct bf_program;
...
+ * @c bpf_get_hash_recalc on the skb, then compares the result against the
```
qdeslandes: Backticks

src/bpfilter/cgen/matcher/meta.h
yaakov-stein: This probably can use some further discussion - my current thoughts, as described in #464, are that matchers have three stages of codegen, two of which are important here:
1) data loading - each flavor will/can have it's own way to load the data into the proper registers
2) comparison - once the values are in the right registers, the comparisons are flavor-agnostic

Given this, my thought is it would make sense to generalize the comparison stage when we can and then have each flavor define it's own data loading step. The idea here is similar, but as mentioned, it does require that we assume previous instructions loaded the correct value into specific registers. 

For now, I renamed the functions to be more explicit that they are only generating the comparison bytecode, but that still runs into the problem you've mentioned (the expectation that data is loaded into certain registers).

qdeslandes: The design is sound, my comment is mostly about the (private) API we define than the actual logic. I would like the code to be more expressive about the fact `r1` is expected to contain the processed value. However, it should be fine in this specific PR and is probably something to refine in the next one (#1/3). I'll add a comment there for more details.

Issue #355, opened 2026-01-12
Generate `BPF_PROG_TYPE_CGROUP_SOCK_ADDR` BPF programs


qdeslandes:
Implement support for `BPF_PROG_TYPE_CGROUP_SOCK_ADDR` BPF programs which can be attached to e.g. `cgroup/connect4` and `cgroup/sendmsg6` to filter at the socket level.

This flavour would be very different from the program types currently available, as it doesn't operate at the packet level. Hence, specific-matchers would have to be written. Luckily, those programs only allow a limited set of socket fields to be accessed, limiting the number of matchers to support.

PR #464, opened 2026-03-08
[1/3] Separate comparison logic from data loading in matchers

yaakov-stein:
Stacked PRs:
- #456 
- -> #464 
- #459 
- #465 

This is helpful for #355 (but makes sense as a standalone refactor as well). Changes are non-functional - all bytecode should remain exactly the same, no new features, etc.

## Summary

Decomposes matcher codegen into independent concerns to enable flavor-specific data loading while sharing comparison logic across flavors.

Every matcher generator performs three operations:
A) Protocol guard : Is this the right L3/L4 protocol?  (R7/R8 check)
B) Data loading (flavor-specific): Get the field value into R1 (R1+R2 for 128-bit)
C) Comparison: Does the value match the rule payload?  (EQ/NE/RANGE/mask)

Previously, concerns B and C were tangled in per-protocol files (`ip4.c`, `ip6.c`, `tcp.c`, `udp.c`, `icmp.c`). Each file hardcoded both packet header loading *and* comparison, making it impossible to reuse comparison logic for non-packet flavors and hard to reuse the logic in general.

### Before

```
bf_matcher_generate_packet()
     │
     ├── ip4.c ── OUTER: guard + load l3_hdr → INNER: offset + 32-bit compare
     ├── ip6.c ── OUTER: guard + load l3_hdr → INNER: offset + 128-bit compare
     ├── tcp.c ── OUTER: guard + load l4_hdr → INNER: offset + port compare
     ├── udp.c ── OUTER: guard + load l4_hdr → INNER: identical to tcp port
     └── icmp.c ── guard + load l4_hdr + 8-bit compare
```

Problems: B+C tangled, tcp/udp port comparison duplicated, adding a new flavor means either duplicating C or refactoring B+C apart.

### After

```
bf_matcher_generate_packet()         future: _bf_cgroup_sock_addr_gen_inline_matcher()
     │                                    │
     ├── A: bf_stub_rule_check_protocol() │  A: bf_stub_rule_check_protocol()  ← SHARED
     ├── B: load from l3_hdr/l4_hdr       │  B: load from bpf_sock_addr        ← FLAVOR-SPECIFIC
     └── C: bf_cmp_*()                    └  C: bf_cmp_*()                     ← SHARED
                    │
          ┌─────────────────────┐
          │  cgen/matcher/cmp.c │
          │  bf_cmp_value()     │  EQ/NE at sizes 1, 2, 4, 16
          │  bf_cmp_masked_value│  masked EQ/NE for network prefixes
          │  bf_cmp_range()     │  16-bit port RANGE
          │  bf_cmp_bitfield()  │  TCP flag ANY/ALL
          └─────────────────────┘
```

Per-protocol files eliminated (`ip4.c/h`, `ip6.c/h`, `tcp.c/h`, `udp.c/h`, `icmp.c/h`). Loading logic absorbed into `packet.c` static helpers; comparison logic extracted into cmp.c.

## Test plan

- unit + e2e testing
- BPF bytecode output is identical for all existing matchers (pure refactor)


src/libbpfilter/include/bpfilter/flavor.h
```diff
@@ -117,6 +116,16 @@ struct bf_flavor_ops
...
     int (*get_verdict)(enum bf_verdict verdict);
+
+    /**
+     * @brief Generate bytecode for a matcher. Required for all flavors.
+     *
+     * @param program Program being generated. Can't be NULL.
+     * @param matcher Matcher to generate code for. Can't be NULL.
+     * @return 0 on success, negative errno on error.
+     */
+    int (*gen_inline_matcher)(struct bf_program *program,
```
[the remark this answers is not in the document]
yaakov-stein: Beyond scope of PR. The breakage is fine for now

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+}
+
+/**
+ * @brief Compute a network prefix mask.
+ *
+ * @param prefixlen Prefix length in bits.
+ * @param mask Output buffer.
+ * @param mask_len Size of mask buffer in bytes (4 or 16).
+ */
+static void _bf_prefix_to_mask(unsigned int prefixlen, uint8_t *mask,
```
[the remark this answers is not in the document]
yaakov-stein: Let's keep this as is for now. This method was moved from where it was previously located and was largely kept the same (combined ip4/ip6). We can fix it separately if necessary

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+#define _bf_make32(a, b, c, d)                                                 \
+    (((uint32_t)(a) << 24) | ((uint32_t)(b) << 16) | ((uint32_t)(c) << 8) |    \
+     (uint32_t)(d))
+#define _BF_MASK_LAST_BYTE 15
+
+/**
+ * @brief Emit a 4-instruction sequence to build a 64-bit immediate from 8 bytes.
+ *
+ * Produces: MOV32_IMM(dst, high32) -> LSH(dst, 32) -> MOV32_IMM(scratch,
+ * low32) -> OR(dst, scratch)
```
qdeslandes: Use `@code` block here.

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+ *
+ * @param program Program to emit into.
```
qdeslandes: Can't be NULL.

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+ * @param dst_reg Destination register for the 64-bit value.
+ * @param scratch_reg Scratch register (clobbered).
+ * @param data Pointer to 8 bytes of data.
+ */
+static int _bf_cmp_build_imm64(struct bf_program *program, int dst_reg,
+                               int scratch_reg, const uint8_t *data)
```
qdeslandes: Expect `uint64_t` as 4th argument to prevent UB/out-of-bounds, then use it as `uint8_t` in the function.

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+
+/**
+ * @brief Compute a network prefix mask.
+ *
+ * @param prefixlen Prefix length in bits.
+ * @param mask Output buffer.
+ * @param mask_len Size of mask buffer in bytes (4 or 16).
+ */
+static void _bf_prefix_to_mask(unsigned int prefixlen, uint8_t *mask,
+                               unsigned int mask_len)
```
qdeslandes: `unsigned int` -> `size_t`

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+        /* 32-bit values: may exceed signed 32-bit immediate range, so
+         * use MOV32_IMM into R2 + JMP_REG. */
+        uint32_t val = *(const uint32_t *)ref;
+
+        EMIT(program, BPF_MOV32_IMM(BPF_REG_2, val));
+        EMIT_FIXUP_JMP_NEXT_RULE(
+            program, BPF_JMP_REG(op == BF_MATCHER_EQ ? BPF_JNE : BPF_JEQ,
+                                 BPF_REG_1, BPF_REG_2, 0));
+        break;
+    }
```
qdeslandes, line 106: Add a 64-bits case for completeness.

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+        if (r)
+            return r;
+
+        if (op == BF_MATCHER_EQ) {
+            EMIT_FIXUP_JMP_NEXT_RULE(
+                program, BPF_JMP_REG(BPF_JNE, BPF_REG_1, BPF_REG_3, 0));
+
+            r = _bf_cmp_build_imm64(program, BPF_REG_3, BPF_REG_4, addr + 8);
+            if (r)
+                return r;
```
qdeslandes: You can build both low and high bits halves before the `if/else` to prevent duplication.

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+            j1 = bf_jmpctx_get(program,
+                               BPF_JMP_REG(BPF_JNE, BPF_REG_2, BPF_REG_3, 0));
+
+            EMIT_FIXUP_JMP_NEXT_RULE(program, BPF_JMP_A(0));
+
+            bf_jmpctx_cleanup(&j0);
+            bf_jmpctx_cleanup(&j1);
```
qdeslandes: Use `_clean_bf_jmpctx_` instead

src/bpfilter/cgen/matcher/cmp.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include <bpfilter/matcher.h>
+
+struct bf_program;
+
+/**
+ * @brief Compare R1 (R1+R2 for 128-bit) against a reference value.
+ *
+ * Internally selects BPF emission strategy based on size:
+ * - size 1-2: BPF_JMP_IMM (value fits in signed 32-bit immediate)
```
qdeslandes: Backticks around keywords.

src/bpfilter/cgen/matcher/cmp.h
```diff
@@ -0,0 +1,69 @@
...
+ * - size 4:   BPF_MOV32_IMM + BPF_JMP_REG (value may exceed signed 32-bit)
```
qdeslandes: Backticks around keywords.

src/bpfilter/cgen/matcher/cmp.h
```diff
@@ -0,0 +1,69 @@
...
+ * - size 16:  multi-word comparison with bf_jmpctx for NE
+ *
+ * @param program Program to emit into. Can't be NULL.
+ * @param op BF_MATCHER_EQ or BF_MATCHER_NE.
```
qdeslandes: Backticks around keywords.

src/bpfilter/cgen/matcher/cmp.h
```diff
@@ -0,0 +1,69 @@
...
+int bf_cmp_value(struct bf_program *program, enum bf_matcher_op op,
+                 const void *ref, unsigned int size);
+
+/**
+ * @brief Mask R1 (R1+R2 for 128-bit), then compare against masked reference.
+ *
+ * Computes mask from prefixlen, emits AND instructions, then compares.
+ *
+ * @param program Program to emit into. Can't be NULL.
+ * @param op BF_MATCHER_EQ or BF_MATCHER_NE.
```
qdeslandes: Backticks around keywords.

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+    case 2: {
+        /* Small values: compare R1 directly via JMP_IMM.
+         * For size 1, ref is uint8_t; for size 2, ref is uint16_t.
+         * Both fit in a signed 32-bit immediate. */
+        uint32_t val =
+            (size == 1) ? *(const uint8_t *)ref : *(const uint16_t *)ref;
+
+        EMIT_FIXUP_JMP_NEXT_RULE(
+            program, BPF_JMP_IMM(op == BF_MATCHER_EQ ? BPF_JNE : BPF_JEQ,
+                                 BPF_REG_1, val, 0));
```
qdeslandes: I'm not a fan of having `r1` loaded as a requirement for this function as it's easy to misuse, it brings no value and prevents this function from using the registers the way it wants.

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+        if (r)
+            return r;
+
+        if (op == BF_MATCHER_EQ) {
+            EMIT_FIXUP_JMP_NEXT_RULE(
+                program, BPF_JMP_REG(BPF_JNE, BPF_REG_1, BPF_REG_3, 0));
+
+            r = _bf_cmp_build_imm64(program, BPF_REG_3, BPF_REG_4, masked_hi);
+            if (r)
+                return r;
```
qdeslandes, line 243: Same as above, this can be done before the `if/else` to prevent duplication.

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+        break;
+    }
+    default:
+        return bf_err_r(-EINVAL, "unsupported masked comparison size %u", size);
+    }
+
+    return 0;
+}
+
+int bf_cmp_range(struct bf_program *program, uint16_t min, uint16_t max)
```
qdeslandes: `bf_cmp_range_16` to clarify this only supports 16-bits payloads.

src/bpfilter/cgen/matcher/packet.c
```diff
@@ -0,0 +1,331 @@
...
+                                          const struct bf_matcher_meta *meta)
+{
+    int r;
+
+    r = bf_stub_rule_check_protocol(program, meta);
+    if (r)
+        return r;
+
+    return bf_stub_load_header(program, meta, BPF_REG_6);
+}
```
qdeslandes: This could be done earlier:
- Check for the protocols for each matcher before `bf_list_foreach (&rule->matchers, matcher_node)` in `_bf_program_generate_rule`. This way you can jump to the next rule early if needed, and you can deduplicate checks (e.g. `icmp.code 8 icmp.type 9` don't need to check for ICMP twice).
- Load the header (`bf_stub_load_header`) within the `bf_list_foreach (&rule->matchers, matcher_node)` loop before `gen_inline_matcher`.

This way it won't mix with the actual flavor-specific loading and comparison operations.

src/bpfilter/cgen/matcher/packet.c
```diff
@@ -0,0 +1,331 @@
...
+                                     const struct bf_matcher *matcher,
+                                     const struct bf_matcher_meta *meta)
+{
+    const uint8_t ehdr = *(uint8_t *)bf_matcher_payload(matcher);
+    uint8_t eh_mask;
+    int r;
+
+    if ((bf_matcher_get_op(matcher) != BF_MATCHER_EQ) &&
+        (bf_matcher_get_op(matcher) != BF_MATCHER_NE))
+        return -EINVAL;
```
qdeslandes: Rely on `meta->ops` for the check: if there is no ops for a specific operator, it's not supported.

src/bpfilter/cgen/matcher/packet.c
```diff
@@ -0,0 +1,331 @@
...
+        if (r)
+            return r;
+
+        EMIT(program, BPF_LDX_MEM(BPF_B, BPF_REG_1, BPF_REG_6,
+                                  meta->hdr_payload_offset));
+        return bf_cmp_bitfield(program, bf_matcher_get_op(matcher),
+                               *(uint8_t *)bf_matcher_payload(matcher));
+    }
+
+    return _bf_matcher_pkt_load_and_cmp(program, matcher, meta);
```
qdeslandes: A switch for this kind of logic would be clearer, what do you think?

src/bpfilter/cgen/matcher/cmp.c
```diff
@@ -0,0 +1,276 @@
...
+     * is a JLT/JGT comparison, we need to have the MSB where the machine
+     * expects then. */
+    EMIT(program, BPF_BSWAP(BPF_REG_1, 16));
+    EMIT_FIXUP_JMP_NEXT_RULE(program, BPF_JMP_IMM(BPF_JLT, BPF_REG_1, min, 0));
+    EMIT_FIXUP_JMP_NEXT_RULE(program, BPF_JMP_IMM(BPF_JGT, BPF_REG_1, max, 0));
+
+    return 0;
+}
+
+int bf_cmp_bitfield(struct bf_program *program, enum bf_matcher_op op,
```
qdeslandes, line 282: Same as `bf_cmp_port`: this function only supports 8-bits flags, either make it clear in its name (`bf_cmp_bitfield_8`), or make it generic (with support for at least 8-bits, and error on unsupported sizes).

src/bpfilter/cgen/matcher/packet.c
```diff
@@ -0,0 +1,331 @@
...
+
+    r = _bf_matcher_pkt_prepare_header(program, meta);
+    if (r)
+        return r;
+
+    EMIT(program,
+         BPF_LDX_MEM(BPF_DW, BPF_REG_1, BPF_REG_6, meta->hdr_payload_offset));
+    EMIT(program, BPF_LDX_MEM(BPF_DW, BPF_REG_2, BPF_REG_6,
+                              meta->hdr_payload_offset + 8));
+    return bf_cmp_value(program, bf_matcher_get_op(matcher),
```
qdeslandes, line 130: Let's reuse BPF's payload size macros (e.g. `BPF_H`) instead of `8`, `16`, ...

src/bpfilter/cgen/matcher/packet.c
```diff
@@ -0,0 +1,331 @@
...
+    EMIT_FIXUP_JMP_NEXT_RULE(
+        program,
+        BPF_JMP_IMM(bf_matcher_get_op(matcher) == BF_MATCHER_EQ ? BPF_JNE :
+                                                                  BPF_JEQ,
+                    BPF_REG_1, (uint16_t)dscp << 4, 0));
+
+    return 0;
+}
+
+int bf_matcher_generate_packet(struct bf_program *program,
```
qdeslandes, line 273: I think I'm starting to put my finger on what exactly I'm not 100% in line with (c.f. https://github.com/facebook/bpfilter/pull/456#discussion_r2887019599). 

While matchers require multiple steps (check/load header, load value, then compare), both the load and compare steps are mixed together. It would be worth decoupling those steps, this way:
- It's easier to figure out how a specific matcher's reference data is loaded from the packet, and what the comparison actually is. We could define `cmp.c` and `load.c`, 
- We can rely more on `bf_matcher_meta`: `_bf_matcher_pkt_generate_tcp_flags` is loading the reference data "manually" while the meta contains all the info we need to do it in a generic fashion
- We can have more generic comparison functions (e.g. `bf_cmp_masked`...)

Doing so, with a well documented behaviour at the beginning of this file (including e.g. "matchers expect the packet's reference data to compare against to be loaded into r1 and r2 (if necessary)") would resolve my initial concern. 

Let me know what you think!

yaakov-stein: Thanks for the clarification - I definitely agree that this is a cleaner approach! To make sure I understood, here's what I did - let me know if this is in line with what you were thinking:

1) Added register arguments to the `bf_cmp_*` functions - they now are even more general and don't make any assumptions of what register needs to be compared (and this now works for `flow_hash`, which allows us to theoretically get rid of that method). Importantly, this also makes the contract fully explicit (no necessitating the a certain register must have certain data).
2) Made better use of the data in `_bf_matcher_meta` for loading. This allows the matcher code generation to be more generic (we just load with `meta->hdr_payload_offset`). At the same time, generalizing the loading right now and making a `load.c` file may be premature - the way packets are loaded (`bf_stub_load_header` + loading `meta->hdr_payload_offset`) and the way `CGROUP_SOCK_ADDR` are loaded (`runtime.arg` and something like `offsetof(struct bpf_sock_addr, user_ip4)`) are different enough that it doesn't feel trivial to make a generic loading change. I'm happy to spend some time thinking about this (I'd imagine we can use `_bf_matcher_meta` in some way to make this work), but I think this can be left for future work. As of now, the loading is generic over packets and is placed in loading methods in `packet.c` (e.g. `_bf_matcher_pkt_load_field`).

Does this solve the concerns you're having?

Note: I think some nits remain and can use a bit of cleanup/refinement, but it's mostly there. Mainly want to align first on the above^.

src/bpfilter/cgen/matcher/cmp.c
yaakov-stein: Are we sure we want to do that? I'd imagine it would negatively affect performance as we're now loading in all of the data before checking each half individually

src/bpfilter/cgen/matcher/cmp.c
yaakov-stein: Same question as above

src/bpfilter/cgen/matcher/packet.c
yaakov-stein: Agreed for the protocols. However, `bf_stub_load_header` is packet specific, so we'd need to wait to do that until we enter `gen_inline_matcher`, right?

src/bpfilter/cgen/matcher/packet.c
yaakov-stein: Not quite sure I understand this one. Are you saying to change the parameters that `bf_cmp_value` gets? I guess, now that we use `meta->hdr_payload_size`, is this still relevant?

src/bpfilter/cgen/matcher/cmp.c
yaakov-stein: Added support up to 32 bits for both and documented that. Should we name it `bf_cmp_*_32`?

src/bpfilter/cgen/matcher/cmp.c
yaakov-stein: Agreed - fixed that. Also commented at length in [here](https://github.com/facebook/bpfilter/pull/464#discussion_r2921552241)

src/bpfilter/cgen/matcher/cmp.c
yaakov-stein: See https://github.com/facebook/bpfilter/pull/464#discussion_r2921828068

src/bpfilter/cgen/program.c
```diff
@@ -285,15 +284,39 @@ static int _bf_program_fixup(struct bf_program *program,
...
     assert(rule);
+    assert(program->runtime.ops->gen_inline_matcher);
 
     if (rule->disabled)
         return 0;
 
-    assert(program->runtime.ops->gen_inline_matcher);
+    /* Emit protocol checks before matchers, deduplicating consecutive
+     * matchers that share the same protocol at a given layer. */
+    bf_list_foreach (&rule->matchers, matcher_node) {
```
yaakov-stein, line 297: This does allow for multiple protocol checks in strange cases. If matchers have ip4 -> ip6 -> ip4 -> ip6, then we'd add a check for each one. I think this is fine for now for a few reasons:
- It's better than it was before where we always checked on every protocol
- Having different non-consecutive `hdr_id`'s from the same layer within a rule should be rare
- We can use a hashmap once the repo supports one to do this more efficiently
- Doing something fancy here is probably beyond the scope of this PR that's focusing on a different point

src/bpfilter/cgen/program.c
```diff
@@ -291,81 +284,45 @@ static int _bf_program_fixup(struct bf_program *program,
...
-            break;
-        case BF_MATCHER_ICMP_TYPE:
-        case BF_MATCHER_ICMP_CODE:
-        case BF_MATCHER_ICMPV6_TYPE:
-        case BF_MATCHER_ICMPV6_CODE:
-            r = bf_matcher_generate_icmp(program, matcher);
+        if (!meta)
+            continue;
+
+        if (meta->layer == BF_MATCHER_LAYER_3 && !l3_checked) {
```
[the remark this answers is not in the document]
yaakov-stein: See https://github.com/facebook/bpfilter/pull/464#discussion_r2921851151

src/bpfilter/cgen/matcher/cmp.c
qdeslandes: It supports 8, 16, and 32 now, so I would keep the current name, like `bf_cmp_value`.

src/bpfilter/cgen/matcher/cmp.h
```diff
@@ -0,0 +1,82 @@
...
+#pragma once
+
+#include <stdint.h>
+
+#include <bpfilter/matcher.h>
+
+struct bf_program;
+
+/**
+ * @brief Compare the value in @p reg against a reference value.
```
qdeslandes: `@p` ->
```
`p`
```

qdeslandes: Same for other instances of `@SOMETHING`.

src/bpfilter/cgen/matcher/cmp.h
```diff
@@ -0,0 +1,82 @@
...
+ * @param reg BPF register holding the value to compare.
+ * @return 0 on success, negative errno on error.
+ */
+int bf_cmp_value(struct bf_program *program, enum bf_matcher_op op,
+                 const void *ref, unsigned int size, int reg);
+
+/**
+ * @brief Mask the value in @p reg, then compare against masked reference.
+ *
+ * For 128-bit comparisons (size=16), the value spans @p reg and @p reg + 1.
```
qdeslandes: This is an example of why backticks are more suited: Doxygen will only consider `reg` as an expression, not `reg + 1`.

src/bpfilter/cgen/matcher/cmp.h
```diff
@@ -0,0 +1,82 @@
...
+ * @param size Comparison width in bytes: 4 or 16.
+ * @param reg BPF register holding the value to compare.
+ * @return 0 on success, negative errno on error.
+ */
+int bf_cmp_masked_value(struct bf_program *program, enum bf_matcher_op op,
+                        const void *ref, unsigned int prefixlen,
+                        unsigned int size, int reg);
+
+/**
+ * @brief Check min <= reg <= max, for values up to 32 bits.
```
qdeslandes: "min <= reg <= max" -> `min <= reg <= max`

src/bpfilter/cgen/matcher/cmp.h
```diff
@@ -0,0 +1,82 @@
...
+ * @param reg BPF register holding the value to compare (host byte order).
+ * @return 0 on success, negative errno on error.
+ */
+int bf_cmp_range(struct bf_program *program, uint32_t min, uint32_t max,
+                 int reg);
+
+/**
+ * @brief Check @p reg against a bitmask, for values up to 32 bits.
+ *
+ * ANY: (reg & flags) != 0
```
qdeslandes: Backticks around the expression.

src/bpfilter/cgen/matcher/packet.c
```diff
@@ -0,0 +1,330 @@
...
+#include <bpfilter/matcher.h>
+
+#include "cgen/matcher/cmp.h"
+#include "cgen/matcher/meta.h"
+#include "cgen/matcher/set.h"
+#include "cgen/program.h"
+#include "cgen/stub.h"
+#include "filter.h"
+
+/* Packet matcher codegen follows a three-stage pipeline:
```
qdeslandes: This would be worth being promoted to a Doxygen comment: `/*` -> `/**`.

src/bpfilter/cgen/matcher/packet.c
qdeslandes: LGTM.

src/bpfilter/cgen/program.c
```diff
@@ -291,81 +284,46 @@ static int _bf_program_fixup(struct bf_program *program,
...
-            if (r)
-                return r;
-            break;
-        case BF_MATCHER_ICMP_TYPE:
-        case BF_MATCHER_ICMP_CODE:
-        case BF_MATCHER_ICMPV6_TYPE:
-        case BF_MATCHER_ICMPV6_CODE:
-            r = bf_matcher_generate_icmp(program, matcher);
+        if (!meta)
+            continue;
```
qdeslandes: `if (bf_matcher_get_type(matcher) == BF_MATCHER_SET) continue;`
then fail if `!meta`.

src/bpfilter/cgen/program.c
```diff
@@ -291,81 +284,46 @@ static int _bf_program_fixup(struct bf_program *program,
...
             if (r)
                 return r;
-            break;
-        default:
-            return bf_err_r(-EINVAL, "unknown matcher type %d",
-                            bf_matcher_get_type(matcher));
-        };
+            l4_hdr_id = meta->hdr_id;
+        }
+    }
```
qdeslandes: See #471. Once merged, you can simplify this by using a bitfield (`field |= BF_FLAG(meta->layer)`) and gate on this. This way we'll only check once for each layer and rules that can't match (2 different protocols checked on the same layer) will be disabled.

Only one small nit though: `BF_MATCHER_SET` is not handled and neither is `META_(D|S)PORT`.

src/bpfilter/cgen/matcher/packet.c
qdeslandes: That's perfectly in line with my comment, and your argument for `load.c` makes sense. LGTM! :)

src/bpfilter/cgen/program.c
```diff
@@ -285,15 +284,40 @@ static int _bf_program_fixup(struct bf_program *program,
...
+        if (checked_layers & BF_FLAG(meta->layer))
+            continue;
+
+        if (meta->layer == BF_MATCHER_LAYER_3 ||
+            meta->layer == BF_MATCHER_LAYER_4) {
+            r = bf_stub_rule_check_protocol(program, meta);
+            if (r)
+                return r;
+            checked_layers |= BF_FLAG(meta->layer);
+        }
```
qdeslandes: We don't have any matcher on L2 yet, but the enum value is defined, so it should be supported here.

src/bpfilter/cgen/matcher/packet.c
```diff
@@ -5,27 +5,281 @@
...
+    EMIT(program,
+         BPF_LDX_MEM(bpf_size, reg, BPF_REG_6, meta->hdr_payload_offset));
+
+    if (meta->hdr_payload_size == 16) {
+        EMIT(program, BPF_LDX_MEM(BPF_DW, reg + 1, BPF_REG_6,
+                                  meta->hdr_payload_offset + 8));
+    }
+
+    return 0;
+}
```
qdeslandes, line 95: I would insert the `EMIT` instruction in each case here, it would clarify the behaviour instead of handling 64-bits compare as a special case. This leads to very low code duplicate (in the source code, no duplication in the BPF bytecode).

src/bpfilter/cgen/matcher/cmp.c
qdeslandes: Not sure which question you're refering to :/

yaakov-stein: This one - https://github.com/facebook/bpfilter/pull/464#discussion_r2921630335

qdeslandes: Yeah, I thought about long after the comment. Let's keep it this way.

src/bpfilter/cgen/cgroup_skb.c
```diff
@@ -130,12 +130,20 @@ static int _bf_cgroup_skb_gen_inline_matcher(struct bf_program *program,
...
+        EMIT(program, BPF_EMIT_CALL(BPF_FUNC_get_hash_recalc));
 
-        return bf_matcher_generate_meta_flow_hash_cmp(program, matcher);
+        if (bf_matcher_get_op(matcher) == BF_MATCHER_RANGE) {
+            uint32_t *hash = (uint32_t *)bf_matcher_payload(matcher);
+            return bf_cmp_range(program, hash[0], hash[1], BPF_REG_0);
+        }
+
+        return bf_cmp_value(program, bf_matcher_get_op(matcher),
+                            bf_matcher_payload(matcher), 4, BPF_REG_0);
```
qdeslandes: I missed that, but `bpf_get_hash_recalc` is not supported for `CGROUP_SKB` programs. See `bf_matcher_meta` for `BF_MATCHER_META_FLOW_HASH`.

yaakov-stein: Nice catch - I also removed the case from the `NF` hooks

PR #470, opened 2026-03-12
lib: implement hashset

### LINKED ISSUE #418: Use hashset data structure in bf_set
Currently bf_set uses linked lists under the hood, and the operations are much slower than they could be.

Let's write new bf_set implementation, that will use simple hashsets. API can stay the same. Let's try avoiding pulling in any dependencies, `hsearch` from glibc should be enough.

Initially let's have it live alongside bf_set, maybe call it bf_hashset. Once we have it tested, we can try swapping the old implementation out.

pzmarzly:
Currently, `bf_set` uses `bf_list` under the hood. This is suboptimal performance-wise. Here I am changing `bf_set` to be backed by `bf_hashset`.

- `lib: helper: add FNV-1a hash function` - imported from another PR, a simple hash function.
- `lib: add bf_hashset` - main commit. Implements a simple hashset (tombstones on deletion, grows 2x when load factor exceeds 50%, never shrinks, linear probing on collision).  See https://en.wikipedia.org/wiki/Linear_probing (with "special flag value" strategy in "Deletion" section).
- `lib: set: use bf_hashset for elems field` - changes `bf_set`'s inner structure

See https://github.com/facebook/bpfilter/pull/460 for previous attempt.

Fixes https://github.com/facebook/bpfilter/issues/418


src/libbpfilter/include/bpfilter/helper.h
```diff
@@ -394,3 +395,22 @@ int bf_read_file(const char *path, void **buf, size_t *len);
...
+/// FNV-1a 64-bit offset basis.
+/// @see https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function
+#define BF_FNV1A_INIT 0xcbf29ce484222325ULL
+/// FNV-1a 64-bit prime.
+#define BF_FNV1A_PRIME 0x100000001b3ULL
+
+/**
+ * @brief Compute a FNV-1a 64-bit hash.
+ *
+ * Pass @ref BF_FNV1A_INIT as @p hash for the initial call. To hash
```
qdeslandes: `@ref` should be replaced with backticks.

src/libbpfilter/helper.c
```diff
@@ -195,3 +195,17 @@ char *bf_trim(char *str)
 
     return bf_rtrim(bf_ltrim(str));
 }
+
+uint64_t bf_fnv1a(const void *data, size_t len, uint64_t hash)
+{
+    assert(data);
+
+    const uint8_t *bytes = data;
```
qdeslandes:
```
const uint8_t *bytes = data;

assert(data);
```

src/bfcli/chain.c
qdeslandes, line 1: Not fan of `bpfilter/data_structures/...` path. As we're moving to daemon-less, I would suggest `bpfilter/core/...`, which will later on contain other core structures of the project (e.g. `bf_chain`).

src/libbpfilter/core/vector.c
qdeslandes, line 1: Check `dynbuf.c`, it's the same logic but `dynbuf` uses `elem_size = 1`. I would suggest implementing `bf_vector_take` to take ownership of the data (so `bf_vector` would re-allocate memory on the next `add`). Then (in a subsequent commit), replace `bf_dynbuf` with `bf_vector`. 

Could you also add a `@todo` in `bf_program` to replace `img` with a `bf_vector`?

src/libbpfilter/data_structures/vector.c
```diff
@@ -0,0 +1,148 @@
...
+#include "bpfilter/helper.h"
+
+#define _BF_VECTOR_INIT_CAP 8
+
+int bf_vector_new(struct bf_vector **vec, size_t elem_size)
+{
+    _free_bf_vector_ struct bf_vector *_vec = NULL;
+
+    assert(vec);
+    assert(elem_size);
```
qdeslandes: Only assert pointers for NULL. `elem_size` should be checked and return an error on 0.

src/libbpfilter/data_structures/vector.c
```diff
@@ -0,0 +1,148 @@
...
+
+void bf_vector_free(struct bf_vector **vec)
+{
+    assert(vec);
+
+    if (!*vec)
+        return;
+
+    bf_vector_clean(*vec);
+    free(*vec);
```
qdeslandes: `freep((void *)vec)`

src/libbpfilter/data_structures/vector.c
```diff
@@ -0,0 +1,148 @@
...
+}
+
+void *bf_vector_get(const struct bf_vector *vec, size_t index)
+{
+    assert(vec);
+
+    if (index >= vec->len)
+        return NULL;
+
+    return (char *)vec->data + (index * vec->elem_size);
```
qdeslandes: No need for `char *`.

src/libbpfilter/data_structures/vector.c
```diff
@@ -0,0 +1,148 @@
...
+    int r;
+
+    assert(vec);
+    assert(elem);
+
+    if (vec->len == vec->cap) {
+        if (vec->cap > SIZE_MAX / 2)
+            return -ENOMEM;
+
+        size_t new_cap = vec->cap ? vec->cap * 2 : _BF_VECTOR_INIT_CAP;
```
qdeslandes: Variable should be defined at the beginning of the scope.

src/libbpfilter/data_structures/vector.c
```diff
@@ -0,0 +1,148 @@
...
+
+        r = bf_vector_resize(vec, new_cap);
+        if (r)
+            return r;
+    }
+
+    memcpy((char *)vec->data + (vec->len * vec->elem_size), elem,
```
qdeslandes: Same as above, `char *` is not necessary here.

src/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,148 @@
...
+    assert(vec);
+
+    if (new_cap < vec->len)
+        return -EINVAL;
+
+    if (new_cap == 0) {
+        freep((void *)&vec->data);
+        vec->cap = 0;
+        return 0;
+    }
```
qdeslandes: Is this necessary?

src/libbpfilter/data_structures/vector.c
```diff
@@ -0,0 +1,148 @@
...
+    if (new_cap < vec->len)
+        return -EINVAL;
+
+    if (new_cap == 0) {
+        freep((void *)&vec->data);
+        vec->cap = 0;
+        return 0;
+    }
+
+    size_t alloc_size;
```
qdeslandes: Define variables at the beginning of the scope.

src/libbpfilter/include/bpfilter/data_structures/vector.h
```diff
@@ -0,0 +1,167 @@
...
+ * @struct bf_vector
+ *
+ * @var bf_vector::data
+ *  Backing buffer. NULL when the vector is empty and has never been allocated.
+ * @var bf_vector::len
+ *  Number of elements currently stored.
+ * @var bf_vector::cap
+ *  Number of elements that can be stored before a reallocation is needed.
+ * @var bf_vector::elem_size
+ *  Size of a single element in bytes.
```
qdeslandes: Use `///` comments on top of the line defining the field instead, because this Doxygen expression is ugly. We still have some laying around, but let's avoid it for new code.

src/libbpfilter/include/bpfilter/data_structures/vector.h
```diff
@@ -0,0 +1,167 @@
...
+struct bf_vector
+{
+    void *data;
+    size_t len;
+    size_t cap;
+    size_t elem_size;
+};
+
+/**
+ * @brief Returns a zero-initialized @ref bf_vector for elements of size @p esz.
```
qdeslandes: Backticks for parameters.

src/libbpfilter/include/bpfilter/data_structures/vector.h
```diff
@@ -0,0 +1,167 @@
...
+ * @param vec Initialised vector. Must be non-NULL.
+ * @return Number of elements that fit without reallocation.
+ */
+size_t bf_vector_cap(const struct bf_vector *vec);
+
+/**
+ * @brief Get a pointer to the n-th element.
+ *
+ * @param vec Initialised vector. Must be non-NULL.
+ * @param index Index of the element. Must be < @ref bf_vector_len.
```
qdeslandes: `< bf_vector_lem`

src/libbpfilter/include/bpfilter/data_structures/vector.h
```diff
@@ -0,0 +1,167 @@
...
+ * element in turn. Safe to break out of but not to remove elements during
+ * iteration.
+ *
+ * @param vec Pointer to the vector. Must be non-NULL.
+ * @param elem Name of the iteration variable. Will be declared as a
+ *        `void *` and cast by the caller.
+ */
+#define bf_vector_foreach(vec, elem)                                           \
+    for (void *(elem) = (vec)->data;                                           \
+         (elem) && (elem) < (void *)((char *)(vec)->data +                     \
```
qdeslandes: Same as above, let's not cast to `char *`.

src/libbpfilter/set.c
```diff
@@ -421,6 +421,35 @@ bool bf_set_is_empty(const struct bf_set *set)
...
+{
+    assert(set);
+
+    return set->n_comps;
+}
+
+enum bf_matcher_type bf_set_get_key_comp(const struct bf_set *set, size_t index)
+{
+    assert(set);
+    assert(index < set->n_comps);
```
qdeslandes: No assert except for non pointer types. Either return the type as a return parameters (and `int` as the function return type), or return a `const enum bf_matcher_type *` array.

src/libbpfilter/include/bpfilter/set.h
```diff
@@ -65,6 +65,24 @@ struct bf_set
...
 };
 
+/**
+ * @brief Iterate over the elements of a set.
+ *
+ * @param set Pointer to the set to iterate over. Must be non-NULL.
+ * @param elem_var Name of the variable containing the current element data
+ *        (as `void *`). This variable will be created automatically.
+ */
+#define bf_set_foreach(set, elem_var)                                          \
```
qdeslandes: Why not
```
#define bf_set_foreach(set, elem_var) bf_list_foreach (&(set)->elems, (elem_var)) 
```

src/libbpfilter/include/bpfilter/data_structures/hashset.h
```diff
@@ -0,0 +1,213 @@
...
+    void *ctx;
+} bf_hashset;
+
+#define _free_bf_hashset_ __attribute__((cleanup(bf_hashset_free)))
+#define _clean_bf_hashset_ __attribute__((cleanup(bf_hashset_clean)))
+
+static inline bool _bf_hashset_slot_is_occupied(void *slot)
+{
+    return slot != NULL && slot != (void *)1;
+}
```
qdeslandes: No private functions in headers.

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,213 @@
...
+    return slot != NULL && slot != (void *)1;
+}
+
+/**
+ * @brief Iterate over all occupied elements in a hashset.
+ *
+ * @param set Pointer to the hashset. Must be non-NULL.
+ * @param elem_var Name of the @c void* variable to hold each element.
+ */
+#define bf_hashset_foreach(set, elem_var)                                      \
```
qdeslandes, line 110: Can't you leverage `bf_vector_foreach` in here to make it simpler?

tests/unit/libbpfilter/data_structures/hashset.c
```diff
@@ -0,0 +1,370 @@
...
+ */
+
+#include <errno.h>
+
+#include <bpfilter/data_structures/hashset.h>
+#include <bpfilter/helper.h>
+
+#include "test.h"
+
+static uint64_t uint32_hash(const void *data, void *ctx)
```
qdeslandes: Static symbols are prefixed with `_bf_`

src/libbpfilter/include/bpfilter/core/hashset.h
qdeslandes, line 1: Does it need to be part of the public headers?

src/libbpfilter/core/hashset.c
qdeslandes, line 1: It might be worth having a `bf_hashset_take`.

qdeslandes (changes requested):
Alright, first pass of review and a few things to fix. I'll do a second, deeper pass, when those are solved. That being said, it's a very welcome addition! :D

src/libbpfilter/include/bpfilter/set.h
pzmarzly: I wanted to have `void* elem_var`, not `bf_list_node* elem_var` . That complicated things a lot: since I need to define 2 variables, and in C's `for(A;B;C)` A can only define variables of the same type, so I need 2 `for`s. But when I have 2 `for`s, suddenly `break` statement doesn't work, so I need to check how the inner loop exited.

I don't like it, but I find the alternative worse. Wdyt?

src/libbpfilter/include/bpfilter/core/hashset.h
pzmarzly: Good point, I took the `bf_set_foreach` from `set.h`, but I could do `bf_vector_foreach (set, elem_var) if (_bf_hashset_slot_is_occupied(elem_var))`

src/libbpfilter/include/bpfilter/core/hashset.h
pzmarzly: Only because `set.h` has its fields public.

Once you effectively merge bpfilter and libbpfilter together, I'm happy to revisit this.

src/libbpfilter/core/vector.c
pzmarzly: Oh, I haven't thought to check that file out. Filed #473

src/libbpfilter/set.c
```diff
@@ -346,8 +367,8 @@ int bf_set_pack(const struct bf_set *set, bf_wpack_t *pack)
     bf_wpack_close_array(pack);
 
     bf_wpack_open_array(pack, "elements");
-    bf_list_foreach (&set->elems, elem_node)
-        bf_wpack_bin(pack, bf_list_node_get_data(elem_node), set->elem_size);
+    bf_set_foreach (set, elem)
```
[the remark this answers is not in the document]
pzmarzly: Isn't the point of a set to not have well defined order? We shouldn't check their equality by comparing byte streams, and we don't.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,289 @@
...
+}
+
+static inline bool _bf_slot_is_live(void *slot)
+{
+    return slot && slot != _BF_HASHSET_TOMBSTONE;
+}
+
+static size_t _bf_hashset_index(const bf_hashset *set, const void *data)
+{
+    return set->ops.hash(data, set->ctx) % _bf_n_slots(set);
```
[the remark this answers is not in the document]
pzmarzly: This breaks badly if we stop using power of 2. I'd rather save future headache and not optimize this much.

src/libbpfilter/include/bpfilter/core/hashset.h
pzmarzly: Actually no, I cannot, because that'd give me void** instead of void*

src/libbpfilter/core/vector.c
pzmarzly: It's unlikely this function will ever be called with 0 (at least intentionally), but why not have this?

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+ * iteration.
+ *
+ * @param vec Pointer to the vector. Must be non-NULL.
+ * @param elem Name of the iteration variable. Will be declared as a
+ *        `void *` and cast by the caller.
+ */
+#define bf_vector_foreach(vec, elem)                                           \
+    for (void *(elem) = (vec)->data;                                           \
+         (elem) && (elem) < (vec)->data + ((vec)->len * (vec)->elem_size);     \
+         (elem) = (elem) + (vec)->elem_size)
```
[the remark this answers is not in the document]
pzmarzly: That's actually an interesting problem. Claude says:

```
# Why tests pass despite the void pointer arithmetic?

Look at what pedantic_c.sh actually compiles (line 24-26):

SRC="${INCLUDES}\nint main(void) { return 0; }"
echo -e ${SRC} | gcc ... -pedantic-errors -std=c17 -Werror ... -c -o /dev/null -

It #includes every public header, then does int main(void) { return 0; }.
That's it — nobody ever expands the macro.
Since bf_vector_foreach is a #define, the compiler only sees it as text. The void pointer
arithmetic inside never gets compiled, so -pedantic-errors never fires.

The codebase is deeply reliant on GNU extensions — and the pedantic test only passes because it never expands any macros.

# Findings by category

## In public headers (would break pedantic_c.sh if macros were expanded)

┌─────────────────────────────────────────────────┬────────────────┬───────────────────────────────────────────────────────────────────────────────────┐
│                    Extension                    │     Count      │                                     Key locations                                 │
├─────────────────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ Statement expressions ({ ... })                 │ ~10 macros     │ helper.h (TAKE_PTR, bf_min, bf_max, bf_swap), logger.h (bf_abort, bf_err_r etc.), │
│                                                 │                │ list.h (bf_list_move, bf_list_emplace), ns.h, pack.h                              │
├─────────────────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ typeof / __typeof__                             │ 7 uses         │ helper.h — TAKE_PTR, TAKE_STRUCT, bf_swap, bf_min, bf_max                         │
├─────────────────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ ##__VA_ARGS__ (comma elision)                   │ 11+ macros     │ logger.h (all log macros), helper.h (_BF_NUM_ARGS), dump.h, list.h                │
├─────────────────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ __attribute__ (cleanup, packed, aligned,        │ 25+            │ Nearly every public header — foundational to the RAII pattern                     │
│ unused)                                         │ definitions    │                                                                                   │
├─────────────────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ Void pointer arithmetic                         │ 1 macro        │ vector.h (bf_vector_foreach) — the one the reviewer flagged                       │
└─────────────────────────────────────────────────┴────────────────┴───────────────────────────────────────────────────────────────────────────────────┘

## In non-public code (won't affect pedantic_c.sh but still non-portable)

- Void pointer arithmetic in dump.c, vector.c, print.c
- __builtin_clzl in bfcli/helper.c
- __builtin_mul_overflow in core/vector.c
- strerrordesc_np declared in public helper.h (glibc-only, link-time issue)

# The real situation

The pedantic_c.sh test is essentially a no-op for macro hygiene. It proves the headers can be included under -pedantic-errors, but since no macro is ever expanded, the compiler never sees any of the GNU
extensions inside them. __attribute__ in particular is interesting — GCC silently accepts __attribute__((...)) even under -pedantic-errors because it uses the double-underscore reserved namespace. So the
cleanup macros and bf_packed etc. are actually fine at include time.
```

@qdeslandes I guess it would be a bigger task to make this codebase actually work with pedantic C17. Want to make a task?

src/libbpfilter/include/bpfilter/helper.h
```diff
@@ -394,3 +395,22 @@ int bf_read_file(const char *path, void **buf, size_t *len);
...
+/// FNV-1a 64-bit offset basis.
+/// @see https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function
+#define BF_FNV1A_INIT 0xcbf29ce484222325ULL
+/// FNV-1a 64-bit prime.
+#define BF_FNV1A_PRIME 0x100000001b3ULL
+
+/**
+ * @brief Compute a FNV-1a 64-bit hash.
+ *
+ * Pass `BF_FNV1A_INIT` as @p hash for the initial call. To hash
```
yaakov-stein: nit: @p hash -> \`hash\`

src/libbpfilter/helper.c
```diff
@@ -195,3 +195,17 @@ char *bf_trim(char *str)
 
     return bf_rtrim(bf_ltrim(str));
 }
+
+uint64_t bf_fnv1a(const void *data, size_t len, uint64_t hash)
+{
+    const uint8_t *bytes = data;
```
yaakov-stein: nit: declaration -> assert -> initialization

src/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * Copyright (c) 2026 Meta Platforms, Inc. and affiliates.
```
yaakov-stein: tiny nit: we don't need the year anymore pre Michel's guidance

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+#define bf_vector_default(esz)                                                 \
+    (bf_vector)                                                                \
+    {                                                                          \
+        .data = NULL, .len = 0, .cap = 0, .elem_size = (esz)                   \
+    }
+
+/**
+ * @brief Iterate over every element of a @ref bf_vector.
+ *
+ * @p elem is declared as a pointer to the element type and will point to each
```
yaakov-stein: nit: backticks here and for @ref bf_vector above

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * @param vec Pointer to the vector pointer. Must be non-NULL.
+ */
+void bf_vector_free(bf_vector **vec);
+
+/**
+ * @brief Clean up a vector, freeing its backing buffer.
+ *
+ * After this call the vector can be reused (e.g. by re-assigning via
+ * @ref bf_vector_default) or discarded.
```
yaakov-stein: nit: backticks (also in other locations, won't leave a comment by each one to avoid clutter)

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+ */
+void bf_vector_clean(bf_vector *vec);
+
+/**
+ * @brief Get the number of elements in the vector.
+ *
+ * @param vec Initialised vector. Must be non-NULL.
+ * @return Number of elements stored.
+ */
+size_t bf_vector_len(const bf_vector *vec);
```
yaakov-stein: It looks like there is some inconsistency in the codebase, but `bf_vector_get_len/cap/data` etc. seems clearer to me when accessing the structs variables (inconsistency: `bf_list_get_head` vs `bf_list_size`). This is true here but also throughout the PR (in `vector`, `set`, and `hashset`)

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+ * @brief Append an element to the end of the vector, growing it if necessary.
+ *
+ * The element is copied from @p elem into the vector's backing buffer.
+ *
+ * @param vec Initialised vector. Must be non-NULL.
+ * @param elem Pointer to the element to copy in. Must be non-NULL and point
+ *        to at least @c vec->elem_size bytes.
+ * @return 0 on success, or a negative errno value on failure.
+ */
+int bf_vector_add(bf_vector *vec, const void *elem);
```
yaakov-stein, line 135: Given that this is a general-purpose vector, I'd imagine there should be some remove functionality (ie `bf_vector_remove`).

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+ * @brief Resize the vector's backing buffer to hold exactly @p new_cap elements.
+ *
+ * @p new_cap must be >= the current length. If @p new_cap is 0 the backing
+ * buffer is freed.
+ *
+ * @param vec Initialised vector. Must be non-NULL.
+ * @param new_cap New capacity (in number of elements).
+ * @return 0 on success, or a negative errno value on failure.
+ */
+int bf_vector_resize(bf_vector *vec, size_t new_cap);
```
yaakov-stein: Should this be exposed as a public API? This seems like something which should be internal - ie why should a user have access to resizing their own vector when that is part of the vector implementation.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+ * @brief Set the number of live elements.
+ *
+ * @p len must be <= the current capacity. No initialization of the new
+ * elements is performed.
+ *
+ * @param vec Initialised vector. Must be non-NULL.
+ * @param len New element count.
+ * @return 0 on success, or -EINVAL if @p len exceeds the capacity.
+ */
+int bf_vector_set_len(bf_vector *vec, size_t len);
```
yaakov-stein: Same comment as `bf_vector_resize` - I'm not sure we should be exposing this as a public API. While there is some type of precedence as a `Vec` in rust does have `resize` and `set_len` (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.set_len), `set_len` is unsafe and `resize` has a different API. I'll discuss more in my main comment, but I think if we're adding a vector class, the point should be to have a basic vector and the API shouldn't be built around what is needed to build a hashset.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+#define _free_bf_vector_ __attribute__((cleanup(bf_vector_free)))
+#define _clean_bf_vector_ __attribute__((cleanup(bf_vector_clean)))
+
+/**
+ * @brief Returns a zero-initialized `bf_vector` for elements of size `esz`.
+ *
+ * @param esz Size of a single element in bytes.
+ * @return A zero-initialized `bf_vector`.
+ */
+#define bf_vector_default(esz)                                                 \
```
yaakov-stein: We check that `elem_size > 0` or return `-EINVAL` in `bf_vector_new`, we should make sure we ensure that as well for stack-allocated initialization.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,168 @@
...
+ * @return Number of elements that fit without reallocation.
+ */
+size_t bf_vector_cap(const bf_vector *vec);
+
+/**
+ * @brief Get a pointer to the n-th element.
+ *
+ * @param vec Initialised vector. Must be non-NULL.
+ * @param index Index of the element. Must be < `bf_vector_len`.
+ * @return Pointer to the element, or NULL if @p index is out of bounds.
```
yaakov-stein: I'm a bit conflicted on returning NULL here. In general, I'd imagine asking for an index out of bounds should be UB (and we should add asserts on it). Returning NULL can allow for silent failure and bugs. If not, it is more of a `bf_vector_try_get`, but I'm not 100% sure how I feel about this.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,300 @@
...
+{
+    return slot == _BF_HASHSET_TOMBSTONE;
+}
+
+static inline bool _bf_hashset_slot_is_live(void *slot)
+{
+    return slot && !bf_hashset_slot_is_tombstone(slot);
+}
+
+static size_t _bf_hashset_index(const bf_hashset *set, const void *data)
```
yaakov-stein, line 49: Almost all of the methods in this file need `assert`'s

yaakov-stein (changes requested):
I think this is much closer, thanks for all of the work! I did not delve deep into all of the implementation details of `bf_hashset` yet, but I have two high-level concerns that I'm interested to hear your thoughts on:
1) Are we sure that a vector is the right underlying data structure for the hashset? While the vector may be a good standalone data structure to include in the repo, it seems to be (a) we're abusing it to make it work for the hashset (with`resize`, `set_len`, while `bf_vector_add` isn't called once), and (b) it isn't the correct data structure. The main reason I think this is the case is that the whole idea of the vector is that it is meant for occasional resizing and lots of appends. This is fundamentally not the case by a hashset, where any append means all elements need to be re-shuffled (and this can be seen by the lack of `bf_vector_add` calls in the `bf_hashset` implementation). The allocation patterns are just not the same. On top of that, using something like a `void **` would be simpler and likely less lines of code.
2) We are silently changing the behavior of `bf_set`, going from being ordered to unordered. I'd lean towards the direction of saying this is a regression as this effects users and is unanticipated behavior (defining a set, and then when you call any get API, the set can be ordered differently), but I'm not sure. I believe there are ways to give ordered API's in the hashset, and this may be worth looking in to (ex. Java's [LinkedHashSet](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashSet.html), Python [dict](https://docs.python.org/3/tutorial/datastructures.html#dictionaries), Rust [IndexMap](https://docs.rs/indexmap/latest/indexmap/)). Maybe this should be a follow-up?

Let me know what you think! (And please disagree with me if you think I'm missing something or if my reasoning isn't quite right :))

src/libbpfilter/include/bpfilter/core/vector.h
pzmarzly: Yup, `get_` looks better to me as well. Thanks for bringing this up.

src/libbpfilter/include/bpfilter/core/vector.h
pzmarzly: Good point. Let's just have `assert` crash the program. We can make a safe version later if we need it.

src/libbpfilter/include/bpfilter/core/vector.h
pzmarzly: Good point. I want to keep using bf_vector, but indeed I rely on internals too much. Changed the code to use `bf_vector_reserve` (which reallocs underlying buffer but doesn't change len) and `bf_vector_add`.

C++ has `vector<T>::reserve` that works similar.

src/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,162 @@
...
+}
+
+void *bf_vector_get(const bf_vector *vec, size_t index)
+{
+    assert(vec);
+    assert(index < vec->len);
+
+    return vec->data + (index * vec->elem_size);
+}
+
```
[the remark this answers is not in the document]
pzmarzly: Nah, I think C17 compat is a bigger topic for the future, we can keep using GCC17 for now.

yaakov-stein:
> Nit: Not all appends cause reshuffling, only some. Most appends just pick an empty slot in the memory and insert into it.

Yes, sorry, I meant whenever we do an append to the underlying vector we need to re-shuffle.

> You are right that we're not using most of vector's properties here. We only really care about it being a resizable array. A `void**` could work. But I think vector could be useful for other things, and I like the `bf_vector_add` / `bf_vector_get` API instead of `void**` management in `hashset.c`.

100% agree - I certainly think that we should be adding `bf_vector` to the repo, especially given #473 and the @todo you've added. That being said, just because we want to add it to the repo doesn't mean that it's the right data structure to use to store `bf_hashset`'s slots. Here are a few reasons/smells for why I think that it probably isn't the right tool here:
1) The previous issue I mentioned. They have fundamentally different allocation patterns. A vector is designed for incremental appends, you push elements one at a time and it grows behind the scenes. A hashset never appends (to the underlying data it holds); when it grows, it allocates a fresh buffer and rehashes every element into new positions. That's just calloc + free, not a resizable array. While we could force a vector to work for us in this way, it doesn't seem like the right use case. 
2) We have two resizing algorithms running at the same time. `bf_hashset` checks if `_bf_hashset_needs_grow` and then calls the grow method. However, `bf_vector` _also_ has it's own determination for when we should grow - why do we need to do this twice, both in terms of unnecessary complexity and performance? 
3) `_bf_hashset_grow` now calls `bf_vector_add` in a loop to fill `new_n_slots` slots with `NULL`. This is n function calls (each with growth-logic checks that can never fire) replacing what would be a single `calloc` with a raw `void**`.  These are symptoms of the same root cause mentioned in (1), as the vector's API is designed for append-and-grow, but the hashset needs allocate-and-fill.

> I'm talking to Quentin about it. Should be fine.

Agreed. I think it should be fine as well, but let's just double check with Quentin.

Based on the above, here is what I propose - as opposed to having this as one large PR, let's break it up into two pieces: 1) the `bf_vector`, which should be added regardless, and 2) the `bf_hashset`. This will make both PR's easier to review and also will allow us to better evaluate whether the two should be linked. What do you think of that?

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,278 @@
...
+{
+    assert(set);
+    assert(data);
+
+    return set->ops.hash(data, set->ctx) % set->cap;
+}
```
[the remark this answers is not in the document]
pzmarzly: Nah. Also, asserts should only be used for NULL-ness checks.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,278 @@
...
+
+    assert(set);
+
+    old_cap = set->cap;
+    if (old_cap > SIZE_MAX / 2)
+        return -ENOMEM;
+
+    new_cap = old_cap ? old_cap * 2 : _BF_HASHSET_INIT_CAP;
+
+    void **new_slots = (void **)calloc(new_cap, sizeof(void *));
```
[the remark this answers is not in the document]
pzmarzly: Re 2: quoting Claude Code:

> Kept the `(void **)calloc` cast — removing it fails `check.lint` (`bugprone-multi-level-implicit-pointer-conversion` is enforced as error). There are no other `void **` allocations in the codebase to compare against; this is a unique case.

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,202 @@
...
+#define _free_bf_hashset_ __attribute__((cleanup(bf_hashset_free)))
+#define _clean_bf_hashset_ __attribute__((cleanup(bf_hashset_clean)))
+
+/**
+ * @brief Check whether a slot holds the tombstone sentinel.
+ *
+ * @param slot Slot value to test.
+ * @return True if `slot` is the tombstone marker.
+ */
+bool bf_hashset_slot_is_tombstone(const void *slot);
```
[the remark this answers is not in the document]
pzmarzly: Nah, let's be consistent with how we define public functions.

src/libbpfilter/include/bpfilter/helper.h
```diff
@@ -394,3 +395,22 @@ int bf_read_file(const char *path, void **buf, size_t *len);
  * @return 0 on success, negative errno value on error.
  */
 int bf_write_file(const char *path, const void *buf, size_t len);
+
+/// FNV-1a 64-bit offset basis.
+/// @see https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function
+#define BF_FNV1A_INIT 0xcbf29ce484222325ULL
+/// FNV-1a 64-bit prime.
+#define BF_FNV1A_PRIME 0x100000001b3ULL
```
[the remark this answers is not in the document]
pzmarzly: Nah, let's keep 2 constants together.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,286 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
pzmarzly: Don't need the year anymore.

src/bpfilter/cgen/program.c
```diff
@@ -711,6 +711,13 @@ static int _bf_program_load_log_map(struct bf_program *program)
     return 0;
 }
 
+static size_t _bf_set_elem_size;
```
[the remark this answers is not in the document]
pzmarzly: Removed this qsort

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,198 @@
...
+ * @param elem_var Name of the `void *` variable to hold each element.
+ */
+#define bf_hashset_foreach(set, elem_var)                                      \
+    for (struct {                                                              \
+             bf_hashset_elem *cur;                                             \
+             int brk;                                                          \
+         } _bf_hset = {.cur = (set)->head};                                    \
+         !_bf_hset.brk && _bf_hset.cur; _bf_hset.cur = _bf_hset.cur->next)     \
+        for (void *(elem_var) = (_bf_hset.brk = 1, _bf_hset.cur->data);        \
+             _bf_hset.brk; _bf_hset.brk = 0)
```
pzmarzly: Claude really cooked here, I didn't know this is possible in C. However, I think this is an overkill. If we give up on `void*` and expose `bf_hashset_elem*`, this becomes 10x simpler.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+}
+
+static size_t _bf_hashset_index(const bf_hashset *set, const void *data)
+{
+    assert(set);
+    assert(data);
+
+    return set->ops.hash(data, set->ctx) % set->cap;
+}
+
```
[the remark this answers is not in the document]
pzmarzly: But asserts are only to be used for checking NULL-ness. Ignoring.

src/libbpfilter/include/bpfilter/helper.h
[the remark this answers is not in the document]
pzmarzly: I want consts together.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
pzmarzly: Yes but we're moving to yearless model.

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,201 @@
...
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/**
+ * @file hashset.h
+ *
+ * Open-addressing hashset with linear probing that preserves insertion order.
+ * Backed by an array of @ref bf_hashset_elem slots; each slot holds a data
```
yaakov-stein: nit: backticks, also below in this paragraph

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,201 @@
...
+    /// First element in insertion order, or NULL if empty.
+    bf_hashset_elem *head;
+    /// Last element in insertion order, or NULL if empty.
+    bf_hashset_elem *tail;
+} bf_hashset;
+
+/**
+ * @brief Compound literal for stack-initialising an empty hashset.
+ *
+ * @param ops_ptr Pointer to a @ref bf_hashset_ops struct. Must be non-NULL.
```
yaakov-stein: nit: backticks, also below

src/libbpfilter/include/bpfilter/core/hashset.h
yaakov-stein, line 1: Now that we keep insertion order, should we add a simple `bf_hashset_take`?

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+{
+    bf_hashset_elem *slot;
+    size_t idx;
+    bool was_tombstone;
+    int r;
+
+    assert(set);
+    assert(data);
+
+    if (data == _BF_HASHSET_TOMBSTONE)
```
yaakov-stein: nit: parentheses since it's multiple lines

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+    set->head = NULL;
+    set->tail = NULL;
+}
+
+void bf_hashset_clean(bf_hashset *set)
+{
+    assert(set);
+
+    if (set->ops.free) {
+        for (bf_hashset_elem *cur = set->head; cur; cur = cur->next)
```
yaakov-stein: nit: `bf_hashset_foreach` in case we ever change implementation details, need to update less code

src/libbpfilter/core/hashset.c
yaakov-stein: `_bf_hashset_slot_is_tombstone`?

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+    assert(data);
+
+    if (data == _BF_HASHSET_TOMBSTONE)
+        return bf_err_r(
+            -EINVAL, "data pointer collides with internal tombstone sentinel");
+
+    if (set->len >= _BF_HASHSET_MAX_ELEMS)
+        return bf_err_r(-ENOMEM, "hashset reached maximum element count");
+
+    if (_bf_hashset_find(set, data, NULL))
```
yaakov-stein: nit: `bf_hashset_contains`, same reason as above

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+
+    freep((void *)&old_slots);
+    return 0;
+}
+
+static bool _bf_hashset_needs_grow(const bf_hashset *set)
+{
+    assert(set);
+
+    if (set->cap == 0)
```
yaakov-stein: Do we need this? If `set->cap == 0`, then we'll return true anyways no matter what the other values are

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+#include "bpfilter/helper.h"
+#include "bpfilter/logger.h"
+
+#define _BF_HASHSET_TOMBSTONE ((void *)1)
+#define _BF_HASHSET_INIT_CAP 16
+#define _BF_HASHSET_MAX_LOAD_NUM 7
+#define _BF_HASHSET_MAX_LOAD_DEN 10
+#define _BF_HASHSET_MAX_ELEMS ((size_t)100000000)
+
+static inline bool _bf_hashset_slot_is_tombstone(const bf_hashset_elem *slot)
```
yaakov-stein: Do we need both `_bf_hashset_slot_is_tombstone` and `_bf_hashset_slot_is_live`, we can just have `!_bf_hashset_slot_is_live` instead of `_bf_hashset_slot_is_tombstone`, right?

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+    assert(slot);
+    return slot->data && !_bf_hashset_slot_is_tombstone(slot);
+}
+
+static size_t _bf_hashset_index(const bf_hashset *set, const void *data)
+{
+    assert(set);
+    assert(data);
+
+    return set->ops.hash(data, set->ctx) % set->cap;
```
yaakov-stein, line 54: Should probably give some indication here that the caller needs to check if `set->cap == 0`?

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+}
+
+static int _bf_hashset_grow(bf_hashset *set)
```
yaakov-stein: I think we can simplify this by not mutating in-place, and instead make a new hashset with the larger capacity - thoughts? (can also use `__clean_bf_hashset__` then and it should be much cleaner)

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,315 @@
...
+}
+
+void bf_hashset_init(bf_hashset *set, const bf_hashset_ops *ops, void *ctx)
+{
+    assert(set);
+    assert(ops);
+    assert(ops->hash);
+    assert(ops->equal);
+
+    set->slots = NULL;
```
yaakov-stein: `bf_hashset_default` instead of manually initializing everything?

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,201 @@
...
+#define bf_hashset_default(ops_ptr, ctx_ptr)                                   \
+    ((bf_hashset) {.ops = *(ops_ptr), .ctx = (ctx_ptr)})
+
+#define _free_bf_hashset_ __attribute__((cleanup(bf_hashset_free)))
+#define _clean_bf_hashset_ __attribute__((cleanup(bf_hashset_clean)))
+
+/**
+ * @brief Iterate over all elements in a hashset in insertion order.
+ *
+ * Do not add or remove elements during iteration.
```
yaakov-stein: Any reason that we can't remove the deletion limitation? Can't we just cache next and do something like this:
```
#define bf_hashset_foreach(set, elem_var)                                      \                             
      for (bf_hashset_elem *(elem_var) = (set)->head,                            \                             
                            *__next = (elem_var) ? (elem_var)->next : NULL;       \                            
           (elem_var);                                                           \
           (elem_var) = __next,                                                  \
           __next = (elem_var) ? (elem_var)->next : NULL)
```
(Inspiration from `list.h` [foreach](https://github.com/facebook/bpfilter/blob/1ef730cd4ca0b922a93ef8455cac5b561ac7b63b/src/libbpfilter/include/bpfilter/list.h#L106))

yaakov-stein (dismissed):
Super neat implementation! Haven't seen an insertion order hashset like this implemented before. First two commits and last commit LGTM. Have a few nits and some questions on `hashset.h/c` but I think this is pretty much ready. (Haven't reviewed the tests yet, will do that tomorrow)

src/libbpfilter/include/bpfilter/core/hashset.h
pzmarzly: What should it give us, `bf_hashset_elem*`? I'm not sure how useful that is.

yaakov-stein: We have `bf_hashset_remove`, which already takes ownership of a single element of data. I was thinking it would be more along the lines of `bf_dynbuf_take` (where it takes all of the data), but I think we can wait to see if Quentin thinks that'd be helpful

src/libbpfilter/core/hashset.c
pzmarzly: Huh, haven't thought of that. There will be some lifetime dance involved, since I want to leave hashset in some reasonable state if growing fails. But should be possible to do safely.

src/libbpfilter/include/bpfilter/core/hashset.h
pzmarzly: That's amazing. Yes, that should work.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,298 @@
...
+    bf_hashset_elem *slot;
+    size_t idx;
+
+    assert(set);
+    assert(data);
+
+    if (!_bf_hashset_find(set, data, &idx))
+        return -ENOENT;
+
+    slot = &set->slots[idx];
```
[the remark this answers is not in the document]
pzmarzly: `bf_list_foreach(list, elem) bf_list_delete(elem)` has the same problem. I think that after you call a function called `*_delete`, you should expect the element to be unusable.

src/libbpfilter/set.c
```diff
@@ -194,9 +217,12 @@ int bf_set_add_elem_raw(struct bf_set *set, const char *raw_elem)
                         raw_elem);
     }
 
-    r = bf_list_add_tail(&set->elems, elem);
+    r = bf_hashset_add(&set->elems, elem);
+    if (r == -EEXIST)
```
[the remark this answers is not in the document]
pzmarzly: Not sure if this needs docs, this behavior was already present

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * @param set Initialised hashset. Must be non-NULL.
+ * @return True if the hashset has no elements.
+ */
+bool bf_hashset_is_empty(const bf_hashset *set);
+
+/**
+ * @brief Pre-allocate capacity for at least @p count elements.
+ *
+ * Ensures the backing array is large enough to hold @p count elements without
```
yaakov-stein: nit: backticks a few times in this doxygen

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,217 @@
...
+ * The slot is marked as a tombstone and the element is unlinked from the
+ * iteration order. The element is freed using the `free` callback if one
+ * was provided.
+ *
+ * @param set Initialised hashset. Must be non-NULL.
+ * @param data Element to delete. Must be non-NULL.
+ * @return 0 on success, -ENOENT if the element is not found, or a negative
+ *  errno value on failure.
+ */
+int bf_hashset_delete(bf_hashset *set, const void *data);
```
yaakov-stein, line 221: Just wondering - why delete instead of remove?

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,313 @@
...
+// Caller must ensure set->cap > 0 to avoid division by zero.
+static size_t _bf_hashset_index(const bf_hashset *set, const void *data)
+{
+    assert(set);
+    assert(data);
+
+    return set->ops.hash(data, set->ctx) % set->cap;
+}
+
+static int _bf_hashset_resize(bf_hashset *set, size_t new_cap)
```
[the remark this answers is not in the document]
yaakov-stein: +1 on this

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,313 @@
...
+{
+    bf_hashset_elem **new_slots;
+
+    new_slots = (bf_hashset_elem **)calloc(new_cap, sizeof(*new_slots));
```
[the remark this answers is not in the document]
yaakov-stein: +1 to this as well

tests/unit/libbpfilter/core/hashset.c
yaakov-stein, line 1: Regarding the tests, I think there's a couple areas that are especially complex that we'd want to test (some of which you already have covered):
- Ensure insertion order is maintained across a grow operation (already covered)
- Validate that this works for several elements that collide, i.e. their hashes are equal. On top of this, it'd be worth testing that removing some of those colliding elements and creating tombstones within that linear probe doesn't cause any buggy behavior. (I don't believe these cases are covered)

yaakov-stein (commented):
Claude has a few valid nits/suggestions and I have one last comment on the tests. Overall LGTM once those points are taken care of!

src/libbpfilter/include/bpfilter/core/hashset.h
pzmarzly: For consistency with bf_list_delete

src/libbpfilter/core/hashset.c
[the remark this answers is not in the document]
pzmarzly: clang-tidy complains without it:
`Multilevel pointer conversion from 'void *' to 'bf_hashset_elem **' (aka 'struct bf_hashset_elem **'), please use explicit cast (clang-tidy bugprone-multi-level-implicit-pointer-conversion)`

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,317 @@
...
+        new_cap *= 2;
+    }
+
+    return _bf_hashset_resize(set, new_cap);
+}
+
+int bf_hashset_add(bf_hashset *set, void *data)
+{
+    bf_hashset_elem *elem;
+    size_t free_idx;
```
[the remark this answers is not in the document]
pzmarzly: Value of 0 is invalid as well, and in the scenario you describe would cause silent data corruption.

Adding -EEXIST is OK though.

src/libbpfilter/include/bpfilter/set.h
```diff
@@ -54,7 +54,7 @@ struct bf_set
     size_t n_comps;
 
     /** Elements of the set. */
-    bf_list elems;
+    bf_hashset elems;
```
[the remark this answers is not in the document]
pzmarzly: We don't guarantee ABI stability.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,317 @@
...
+static int _bf_hashset_resize(bf_hashset *set, size_t new_cap)
+{
+    bf_hashset_elem **new_slots;
+
+    assert(set);
+
+    new_slots = (bf_hashset_elem **)calloc(new_cap, sizeof(*new_slots));
+    if (!new_slots)
+        return -ENOMEM;
+
```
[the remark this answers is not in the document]
pzmarzly: Required by clang-tidy.

`Multilevel pointer conversion from 'void *' to 'bf_hashset_elem **' (aka 'struct bf_hashset_elem **'), please use explicit cast (clang-tidy bugprone-multi-level-implicit-pointer-conversion)`

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,331 @@
...
+#define _BF_HASHSET_MAX_CAP (SIZE_MAX / 16 + 1)
+
+static inline size_t _bf_round_next_power_of_2(size_t value)
+{
+    value--;
+    value |= value >> 1;
+    value |= value >> 2;
+    value |= value >> 4;
+    value |= value >> 8;
+    value |= value >> 16;
```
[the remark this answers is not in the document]
pzmarzly: Fixed, and opened https://github.com/facebook/bpfilter/pull/494 for cgen.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,334 @@
...
+/* Maximum load factor before growing. Lowering this number reduces collisions
+ * but causes higher memory usage. */
+#define _BF_HASHSET_MAX_LOAD_NUM 5
+#define _BF_HASHSET_MAX_LOAD_DEN 10
+/* Largest power-of-two element count (2^60) that still leaves headroom for
+ * load-factor arithmetic (slots_in_use * 10, cap * 5) without overflowing
+ * size_t. */
+#define _BF_HASHSET_MAX_CAP (SIZE_MAX / 16 + 1)
+
+static inline size_t _bf_round_next_power_of_2(size_t value)
```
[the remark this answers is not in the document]
pzmarzly: Fixing in #494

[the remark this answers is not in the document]
pzmarzly: fixed in https://github.com/facebook/bpfilter/pull/494

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,334 @@
...
+    new_slots = (bf_hashset_elem **)calloc(new_cap, sizeof(*new_slots));
+    if (!new_slots)
+        return -ENOMEM;
+
+    bf_hashset_foreach (set, elem) {
+        size_t idx = set->ops.hash(elem->data, set->ctx) % new_cap;
+        while (new_slots[idx])
+            idx = (idx + 1) % new_cap;
+        new_slots[idx] = elem;
+    }
```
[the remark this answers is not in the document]
pzmarzly: But then we could silently skip over some elements. If someone makes a serious implementation bug like that, it's better to crash or busyloop, than to silently lose data.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,334 @@
...
+}
+
+void bf_hashset_clean(bf_hashset *set)
+{
+    assert(set);
+
+    bf_hashset_foreach (set, elem) {
+        if (set->ops.free)
+            set->ops.free(&elem->data, set->ctx);
+        freep((void *)&elem);
```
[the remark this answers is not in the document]
pzmarzly: I think it's pretty obvious way to do cleanup?

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,218 @@
...
+} bf_hashset_elem;
+
+/**
+ * @brief Open-addressing hashset with linear probing. Preserves insertion
+ *        order.
+ */
+typedef struct bf_hashset
+{
+    /// Backing array of slot pointers.
+    bf_hashset_elem **slots;
```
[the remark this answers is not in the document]
pzmarzly: I'll add a comment. The `bf_hashset` and the `bf_hashset_elem` fields need to be visible for `bf_hashset_foreach`.

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,226 @@
...
+#define bf_hashset_default(ops_ptr, ctx_ptr)                                   \
+    ((bf_hashset) {.ops = *(ops_ptr), .ctx = (ctx_ptr)})
+
+#define _free_bf_hashset_ __attribute__((cleanup(bf_hashset_free)))
+#define _clean_bf_hashset_ __attribute__((cleanup(bf_hashset_clean)))
+
+/**
+ * @brief Iterate over all elements in a hashset in insertion order.
+ *
+ * Unsafe to add elements during iteration (can cause rehashing).
```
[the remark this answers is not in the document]
pzmarzly: I think it's obvious?

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,226 @@
...
+ * @brief Element node stored in a hashset, maintaining an insertion-order
+ *        linked list.
+ *
+ * @warning From the user's perspective, the inside of this structure should
+ * not be directly accessed. Directly modifying any of the fields should be
+ * considered undefined behavior.
+ */
+typedef struct bf_hashset_elem
+{
+    /// User-provided data pointer.
```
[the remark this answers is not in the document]
pzmarzly: Same pattern in bf_list_foreach. Macros need access to struct internals.

tests/harness/test.c
```diff
@@ -251,21 +251,17 @@ bool bft_counter_eq(const struct bf_counter *lhs, const struct bf_counter *rhs)
...
 
     if (lhs->elem_size != rhs->elem_size)
         return false;
 
-    n0 = bf_list_get_head(&lhs->elems);
-    n1 = bf_list_get_head(&rhs->elems);
-    for (; n0 || n1; n0 = bf_list_node_next(n0), n1 = bf_list_node_next(n1)) {
-        if (0 != memcmp(bf_list_node_get_data(n0), bf_list_node_get_data(n1),
-                        lhs->elem_size))
+    for (bf_hashset_elem *n0 = lhs->elems.head, *n1 = rhs->elems.head; n0 || n1;
```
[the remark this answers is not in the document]
pzmarzly: Test code, should be fine.

src/libbpfilter/set.c
```diff
@@ -459,30 +495,10 @@ int bf_set_add_many(struct bf_set *dest, struct bf_set **to_add)
...
-
-        if (!found) {
-            r = bf_list_add_tail(&dest->elems,
-                                 bf_list_node_get_data(elem_node));
-            if (r)
-                return bf_err_r(r, "failed to add element to set");
-            // Take ownership of data to stop to_add cleanup from freeing it.
-            bf_list_node_take_data(elem_node);
-        }
+    bf_hashset_foreach (&(*to_add)->elems, elem) {
```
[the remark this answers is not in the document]
pzmarzly: Fair point, something like `bf_hashset_take_node` and `bf_hashset_add_node` (that rewires existing `bf_hashset_elem*`) would work, though I'm not sure if it's worth it?

In my intuition, a `free(<bf_hashset_elem*>)` followed by `malloc(sizeof(bf_hashset_elem))` should be pretty cheap, since we have a perfect gap of unallocated memory to reuse.

src/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,351 @@
...
+/**
+ * @brief Checks if an element is present in the hashset.
+ *
+ * If it is, also sets `found_index` to the index of the found element.
+ * If it is not, also sets `free_index` to the index of the first free slot,
+ *               if there is any free slot.
+ *
+ * @param set Hashset to search. Can't be NULL.
+ * @param data Element to search for. Can't be NULL.
+ * @param found_index If non-NULL, will be set to the index of the found element.
```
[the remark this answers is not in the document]
pzmarzly: Will fix it in next iteration, for now I'll wait for human review.

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,226 @@
...
+ * @param ops Callbacks for hashing, comparing, and freeing elements.
+ *        `hash` and `equal` must be non-NULL. Can't be NULL.
+ * @param ctx Opaque context pointer passed to every callback. Can be NULL.
+ */
+void bf_hashset_init(bf_hashset *set, const bf_hashset_ops *ops, void *ctx);
+
+/**
+ * @brief Clean up a hashset, freeing all elements and the backing buffer.
+ *
+ * After this call the hashset is empty and can be reused or discarded.
```
qdeslandes: This is a bit confusing: the hashset should not be discarded if allocated with `bf_hashset_new()`.

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,226 @@
...
+ * returned; ownership of `data` is not transferred in that case.
+ * The hashset grows automatically when the load factor is exceeded.
+ * On successful insertion, the hashset takes ownership of `data` and
+ * appends it to the end of the iteration order.
+ *
+ * @param set Initialised hashset. Can't be NULL.
+ * @param data Pointer to store. Can't be NULL.
+ * @return 0 on success, or a negative errno value on failure.
+ */
+int bf_hashset_add(bf_hashset *set, void *data);
```
qdeslandes: If the hashset takes ownership of the data on success, we should pass `void **data` instead, so `bf_hashset_add()` sets it to NULL.

src/libbpfilter/include/bpfilter/core/hashset.h
```diff
@@ -0,0 +1,226 @@
...
+ * @brief Delete an element from the hashset.
+ *
+ * The slot is marked as a tombstone and the element is unlinked from the
+ * iteration order. The element is freed using the `free` callback if one
+ * was provided.
+ *
+ * @param set Initialised hashset. Must be non-NULL.
+ * @param data Element to delete. Must be non-NULL.
+ * @return 0 on success, -ENOENT if the element is not found, or a negative
+ *  errno value on failure.
```
qdeslandes:
```
 * @return 0 on success, -ENOENT if the element is not found, or a negative
 *  errno value on failure.
```

->

```
 * @return 0 on success, `-ENOENT` if the element is not found, or a negative
 *         errno value on failure.
```

src/libbpfilter/set.c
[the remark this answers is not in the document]
qdeslandes: All good for me, keep it this way.

src/libbpfilter/set.c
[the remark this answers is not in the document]
qdeslandes: > In my intuition, a free(<bf_hashset_elem*>) followed by malloc(sizeof(bf_hashset_elem)) should be pretty cheap, since we have a perfect gap of unallocated memory to reuse.

I would argue that memory allocators are not intuitive :p 

Furthermore: we could have `bf_hashset_take()` in combination with `bf_hashset_add()`, assuming `bf_hashset_add()` expected `void **ptr`.

tests/unit/libbpfilter/set.c
```diff
@@ -122,7 +122,44 @@ static void add_multiple_elems(void **state)
...
-    assert_int_equal(bf_list_size(&set->elems), 5);
+    assert_int_equal(bf_hashset_size(&set->elems), 5);
+}
+
+static void add_duplicate(void **state)
+{
+    _free_bf_set_ struct bf_set *set = NULL;
+
+    enum bf_matcher_type key[] = {BF_MATCHER_IP4_SADDR};
+
```
qdeslandes, line 133: No need for newlines here.

tests/unit/libbpfilter/set.c
```diff
@@ -122,7 +122,44 @@ static void add_multiple_elems(void **state)
...
+    assert_ok(bf_set_new(&set, "test", key, ARRAY_SIZE(key)));
+    assert_ok(bf_set_add_elem(set, &elem));
+    assert_ok(bf_set_add_elem(set, &elem));
+    assert_int_equal(bf_hashset_size(&set->elems), 1);
+}
+
+static void reserve(void **state)
+{
+    _free_bf_set_ struct bf_set *set = NULL;
+
```
qdeslandes, line 147: No need for newline.

src/libbpfilter/include/bpfilter/core/hashset.h
qdeslandes: I'm all for it, as suggested by another comment (somewhere in this PR).

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+    bf_hashset_init(&set, &_bf_uint32_ops, NULL);
+    assert_true(bf_hashset_is_empty(&set));
+    assert_int_equal(bf_hashset_size(&set), 0);
+    assert_int_equal(bf_hashset_cap(&set), 0);
+}
+
+static void add_and_contains(void **state)
+{
+    _clean_bf_hashset_ bf_hashset set;
+    uint32_t key;
```
qdeslandes: `key = 42`

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+
+    (void)state;
+
+    bf_hashset_init(&set, &_bf_uint32_ops, NULL);
+
+    key = 42;
+    assert_false(bf_hashset_contains(&set, &key));
+    assert_ok(bf_hashset_add(&set, _make_u32(42)));
```
qdeslandes: `_make_u32(key)`

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+    for (uint32_t i = 0; i < 5; ++i)
+        assert_ok(bf_hashset_add(&set, _make_u32(i * 100)));
+
+    assert_int_equal(bf_hashset_size(&set), 5);
+
+    for (uint32_t i = 0; i < 5; ++i) {
+        uint32_t key = i * 100;
+        assert_true(bf_hashset_contains(&set, &key));
+    }
+}
```
qdeslandes, line 125: This could be a single test.

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+
+    key = 20;
+    assert_true(bf_hashset_contains(&set, &key));
+
+    key = 99;
+    assert_int_equal(bf_hashset_delete(&set, &key), -ENOENT);
+    assert_int_equal(bf_hashset_size(&set), 1);
+}
+
+static void remove_and_readd(void **state)
```
qdeslandes: `remove_and_readd` -> typo

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+    size_t count;
+
+    (void)state;
+
+    bf_hashset_init(&set, &_bf_uint32_ops, NULL);
+
+    // foreach on empty set does nothing
+    count = 0;
+    bf_hashset_foreach (&set, elem) {
+        (void)elem;
```
qdeslandes, line 217: `assert_true(0)` in the loop should be enough.

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+
+    key = 3;
+    assert_ok(bf_hashset_delete(&set, &key));
+
+    // Removed element is skipped, order preserved
+    bf_hashset_foreach (&set, elem) {
+        assert_int_equal(*(uint32_t *)elem->data, expected[idx]);
+        ++idx;
+    }
+    assert_int_equal(idx, 4);
```
qdeslandes, line 261: `bf_hashset_size()` could be tested here too.

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+        assert_ok(bf_hashset_add(&set, _make_u32(i + 1)));
+
+    count = 0;
+    bf_hashset_foreach (&set, elem) {
+        (void)elem;
+        ++count;
+        break;
+    }
+    assert_int_equal(count, 1);
+}
```
qdeslandes: Not convinced this test brings value.

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+
+    (void)state;
+
+    bf_hashset_init(&set, &_bf_uint32_ops, NULL);
+
+    assert_ok(bf_hashset_reserve(&set, 0));
+    assert_int_equal(bf_hashset_cap(&set), 0);
+
+    assert_ok(bf_hashset_reserve(&set, 100));
+    assert_true(bf_hashset_cap(&set) >= 200);
```
qdeslandes: Unit tests should not speculate about the internal behaviour of the function. If we reserve `100`, we expect to have at least `100`, not `200`.

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+
+    size_t cap_after = bf_hashset_cap(&set);
+    for (uint32_t i = 0; i < 100; ++i)
+        assert_ok(bf_hashset_add(&set, _make_u32(i)));
+    assert_int_equal(bf_hashset_cap(&set), cap_after);
+    assert_int_equal(bf_hashset_size(&set), 100);
+
+    assert_ok(bf_hashset_reserve(&set, 10));
+    assert_int_equal(bf_hashset_cap(&set), cap_after);
```
qdeslandes: Those tests assume knowledge of the hashset internal, which is not expected from unit tests.

tests/unit/libbpfilter/core/hashset.c
```diff
@@ -0,0 +1,431 @@
...
+
+static void grow(void **state)
+{
+    _clean_bf_hashset_ bf_hashset set;
+    size_t idx;
+
+    (void)state;
+
+    bf_hashset_init(&set, &_bf_uint32_ops, NULL);
+    assert_true(bf_hashset_cap(&set) <= 16);
```
qdeslandes: Same issue with `bf_hashset_cap`: it's purely internal to `bf_hashset`, it doesn't need to be public.

qdeslandes (changes requested):

tests/unit/libbpfilter/set.c
pzmarzly: clang-format disagrees:

```
/repos/bpfilter/tests/unit/libbpfilter/set.c:37:45: error: code should be clang-formatted [-Wclang-format-violations]
    _free_bf_set_ struct bf_set *set = NULL;
                                            ^
```

tests/unit/libbpfilter/set.c
pzmarzly: clang-format disagrees.

tests/unit/libbpfilter/core/hashset.c
pzmarzly: I added it after AI reviewer kept complaining. Removing.

pzmarzly:
Rebased, addressed feedback. Split `bf_hashset_delete` into `bf_hashset_delete` (removes elem and frees it) and `bf_hashset_take` (removes elem and returns it). The `bf_set_add_many` now uses `take` + `add` to move element between sets with 1 malloc instead of 3.

qdeslandes (approved):
Great work, thanks for doing it and congrats on the merged PR!

src/libbpfilter/include/bpfilter/core/vector.h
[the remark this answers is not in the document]
qdeslandes: Sure, though it would be low-ish priority :)

qdeslandes:
I can't find the remaining open comment(s) from Yaakov, so I'll assume everything has been addressed and merge this PR :)

PR #471, opened 2026-03-12
lib: chain: disable rules with incompatible same-layer matchers

qdeslandes:
A rule combining matchers on the same network layer but targeting different protocols (e.g. ip4.saddr + ip6.daddr, or tcp.sport + udp.dport) can never match a packet. Detect these during chain validation and disable the rule instead of generating dead BPF code.

### DIFF
--- a/src/libbpfilter/chain.c
+++ b/src/libbpfilter/chain.c
@@ -6,6 +6,7 @@
 #include "bpfilter/chain.h"
 
 #include <errno.h>
+#include <limits.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -14,6 +15,7 @@
 #include "bpfilter/hook.h"
 #include "bpfilter/list.h"
 #include "bpfilter/logger.h"
+#include "bpfilter/matcher.h"
 #include "bpfilter/pack.h"
 #include "bpfilter/rule.h"
 #include "bpfilter/set.h"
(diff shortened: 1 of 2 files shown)


src/libbpfilter/chain.c
```diff
@@ -70,6 +117,9 @@ int _bf_chain_check_rule(struct bf_chain *chain, struct bf_rule *rule)
 
     rule->disabled = r;
 
+    if (!rule->disabled)
+        rule->disabled = _bf_rule_has_incompatible_matchers(rule);
```
yaakov-stein: `rule->disabled = r | _bf_rule_has_incompatible_matchers(rule);`? Or probably `||` so we don't run  `_bf_rule_has_incompatible_matchers(rule)` unless we need to

tests/e2e/rules/incompatible_matchers.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
yaakov-stein: Are the e2e tests valuable here? IIUC, the prior behavior would be the same as the behavior after the change, so what are we really testing? The only difference would be the codegen should be omitted and the rule should be disabled, which the e2e tests can't really cover, right?

qdeslandes: You're right, this test is bullshit!

tests/unit/libbpfilter/chain.c
```diff
@@ -141,6 +141,137 @@ static void mixed_enabled_disabled_log_flag(void **state)
...
+        assert_ok(bf_chain_new(&chain, "test", BF_HOOK_TC_EGRESS,
+                               BF_VERDICT_ACCEPT, NULL, &rules));
+        assert_true(rule->disabled);
+        assert_int_equal(chain->flags & BF_FLAG(BF_CHAIN_LOG), 0);
+    }
+}
+
 static void get_set_by_name(void **state)
 {
     _free_bf_chain_ struct bf_chain *chain = bft_chain_dummy(true);
```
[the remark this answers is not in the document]
qdeslandes: Those tests wouldn't give use a useful signal, and we already have a lot of tests for ensure the matchers are properly generated.

src/libbpfilter/chain.c
```diff
@@ -58,6 +58,53 @@ static int _bf_rule_references_empty_set(const struct bf_chain *chain,
...
+
+    bf_list_foreach (&rule->matchers, matcher_node) {
+        struct bf_matcher *matcher = bf_list_node_get_data(matcher_node);
+        const struct bf_matcher_meta *meta;
+
+        if (bf_matcher_get_type(matcher) == BF_MATCHER_SET)
+            continue;
+
+        meta = bf_matcher_get_meta(bf_matcher_get_type(matcher));
+        if (!meta || meta->layer <= BF_MATCHER_NO_LAYER)
```
yaakov-stein: Should we return error on `!meta`?

tests/unit/libbpfilter/chain.c
```diff
@@ -141,6 +141,137 @@ static void mixed_enabled_disabled_log_flag(void **state)
...
+                                      &port, sizeof(port)));
+        assert_ok(bf_list_add_tail(&rules, rule));
+
+        assert_ok(bf_chain_new(&chain, "test", BF_HOOK_TC_EGRESS,
+                               BF_VERDICT_ACCEPT, NULL, &rules));
+        assert_false(rule->disabled);
+    }
+
+    // Disabled rule does not set BF_CHAIN_LOG.
+    {
```
yaakov-stein: This is already tested here: https://github.com/facebook/bpfilter/blob/main/tests/unit/libbpfilter/chain.c#L109

tests/unit/libbpfilter/chain.c
```diff
@@ -141,6 +141,137 @@ static void mixed_enabled_disabled_log_flag(void **state)
...
+        assert_ok(bf_rule_add_matcher(rule, BF_MATCHER_UDP_DPORT, BF_MATCHER_EQ,
+                                      &port, sizeof(port)));
+        assert_ok(bf_list_add_tail(&rules, rule));
+
+        assert_ok(bf_chain_new(&chain, "test", BF_HOOK_TC_EGRESS,
+                               BF_VERDICT_ACCEPT, NULL, &rules));
+        assert_true(rule->disabled);
+    }
+
+    // L4 conflict: ICMP + ICMPv6 matchers.
```
yaakov-stein: Do we need multiple L4 conflicts?

yaakov-stein (dismissed):
Some minor comments, overall LGTM!

src/libbpfilter/chain.c
```diff
@@ -58,6 +59,58 @@ static int _bf_rule_references_empty_set(const struct bf_chain *chain,
...
+/**
+ * @brief Check if a rule has matchers on the same layer but different
+ * protocols.
+ *
+ * A rule matching on e.g. both IPv4 source address and IPv6 destination
+ * address can never match a packet (a packet is either IPv4 or IPv6, not
+ * both). Same for layer 4 (e.g. TCP vs UDP). Such rules should be disabled.
+ *
+ * @param rule Rule to check.
+ * @return 0 if no conflict, 1 if the rule contains incompatible matchers.
```
yaakov-stein: negative on error

src/libbpfilter/chain.c
```diff
@@ -68,7 +121,7 @@ int _bf_chain_check_rule(struct bf_chain *chain, struct bf_rule *rule)
     if (r < 0)
         return r;
 
-    rule->disabled = r;
+    rule->disabled = r || _bf_rule_has_incompatible_matchers(rule);
```
yaakov-stein: I think we now need to account for the error here 🤦

PR #476, opened 2026-03-16
Fix DSCP matchers to match 6-bit field per RFC 2474

yaakov-stein:
### Summary

The DSCP matchers were matching the full 8-bit Traffic Class / TOS byte instead of the 6-bit Differentiated Services Code Point defined in https://www.rfc-editor.org/rfc/rfc2474:

The DSCP bit layout ([Section 3](https://www.rfc-editor.org/rfc/rfc2474#section-3)):
```
  0   1   2   3   4   5   6   7
+---+---+---+---+---+---+---+---+
|         DSCP          |  CU   |
+---+---+---+---+---+---+---+---+
```

The low 2 bits (CU, now ECN per [RFC 3168](https://www.rfc-editor.org/rfc/rfc3168)) were being included in the match, meaning a rule targeting DSCP 46 (EF) would fail to match packets where ECN bits were set, and could false-match packets with a different DSCP but matching ECN bits.

### Changes

**`lib: matcher: deduplicate numeric parsers with _bf_strtoul`**
- Extract a generic `_bf_strtoul` helper that parses a decimal or hexadecimal string into an unsigned integer
- Refactor `_bf_parse_int`, `_bf_parse_int_range`, `_bf_parse_u8`, `_bf_parse_l3_proto`, `_bf_parse_l4_proto`, `_bf_parse_icmp_type`, `_bf_parse_icmp_code`, and `_bf_parse_icmpv6_type` to use it, removing
duplicated `strtoul`

**`lib: matcher: accept hexadecimal values for L4 protocol matchers`**
- Uses `_bf_strtoul` to enable parsing hex. Updates tests and docs.

**`daemon: cgen: match only 6-bit DSCP field, not full traffic class`**
- IPv4: mask TOS byte with `0xfc` to isolate DSCP, compare against `dscp << 2`
- IPv6: mask first 2 header bytes with `0x0fc0` to isolate DSCP, compare against `dscp << 6`
- Using `_bf_parse_uint`, restrict parser input range from 0-255 to 0-63 (`BF_DSCP_MAX`)
- Update unit and e2e tests for the new valid range

### Test plan

- Unit + e2e tests, including updated `ip4_dscp` and `ip6_dscp` tests that validate the 0-63 range and reject values ≥ 64

### DIFF
--- a/src/libbpfilter/matcher.c
+++ b/src/libbpfilter/matcher.c
@@ -36,6 +36,7 @@
 
 #define INET4_ADDRSTRLEN 16
 #define INET6_ADDRSTRLEN 46
+#define BF_DSCP_MAX 63
 
 #define BF_PAYLOAD_OPS(type, size, parser_cb, printer_cb)                      \
     [type] = {size, parser_cb, printer_cb}
@@ -728,30 +729,30 @@ void _bf_print_icmp_code(const void *payload)
     (void)fprintf(stdout, "%" PRIu8, *(uint8_t *)payload);
 }
 
-static int _bf_parse_u8(enum bf_matcher_type type, enum bf_matcher_op op,
-                        void *payload, const char *raw_payload)
+static int _bf_parse_dscp(enum bf_matcher_type type, enum bf_matcher_op op,
+                          void *payload, const char *raw_payload)
 {
     unsigned long value;
 
     assert(payload);
     assert(raw_payload);
 
-    if (!_bf_strtoul(raw_payload, &value) && value <= UINT8_MAX) {
+    if (!_bf_strtoul(raw_payload, &value) && value <= BF_DSCP_MAX) {
         *(uint8_t *)payload = (uint8_t)value;
         return 0;
     }
 
     return bf_err_r(
         -EINVAL,
-        "\"%s %s\" expects a decimal or hexadecimal value (0-255), not '%s'",
+        "\"%s %s\" expects a DSCP value (0-63) in decimal or hexadecimal notation, not '%s'",
         bf_matcher_type_to_str(type), bf_matcher_op_to_str(op), raw_payload);
 }
 
-static void _bf_print_u8(const void *payload)
+static void _bf_print_dscp(const void *payload)
 {
     assert(payload);
 
-    (void)fprintf(stdout, "0x%02" PRIx8, *(uint8_t *)payload);
+    (void)fprintf(stdout, "%" PRIu8, *(uint8_t *)payload);
 }
 
 static int _bf_parse_icmpv6_type(enum bf_matcher_type type,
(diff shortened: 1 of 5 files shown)


src/libbpfilter/matcher.c
```diff
@@ -249,25 +241,16 @@ int _bf_parse_l4_proto(enum bf_matcher_type type, enum bf_matcher_op op,
...
-    }
 
-    bf_err(
+    return bf_err_r(
+        -EINVAL,
         "\"%s %s\" expects a transport layer protocol name (e.g. \"ICMP\", case insensitive), or a valid decimal internet protocol number, not '%s'",
         bf_matcher_type_to_str(type), bf_matcher_op_to_str(op), raw_payload);
-
-    return -EINVAL;
 }
```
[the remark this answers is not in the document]
yaakov-stein: This requires tests and docs, so this can be a follow-up change

src/libbpfilter/matcher.c
```diff
@@ -263,11 +253,10 @@ int _bf_parse_l4_proto(enum bf_matcher_type type, enum bf_matcher_op op,
         return 0;
     }
 
-    bf_err(
+    return bf_err_r(
```
yaakov-stein, line 268: Drive-by fixes/simplifications to `_bf_parse_l4_proto`

src/libbpfilter/matcher.c
```diff
@@ -64,32 +65,50 @@ struct bf_matcher
...
+        return -EINVAL;
+
+    if (max <= UINT8_MAX)
+        *(uint8_t *)payload = (uint8_t)value;
+    else if (max <= UINT16_MAX)
+        *(uint16_t *)payload = (uint16_t)value;
+    else if (max <= UINT32_MAX)
         *(uint32_t *)payload = (uint32_t)value;
+    else
+        *(uint64_t *)payload = (uint64_t)value;
```
qdeslandes: We should not infer the size of `payload` from `max`.

src/libbpfilter/matcher.c
```diff
@@ -64,32 +65,50 @@ struct bf_matcher
...
+
+    return 0;
+}
+
+int _bf_parse_int(enum bf_matcher_type type, enum bf_matcher_op op,
```
qdeslandes: `static`

src/libbpfilter/matcher.c
```diff
@@ -64,32 +65,50 @@ struct bf_matcher
     uint8_t payload[];
 };
 
-int _bf_parse_int(enum bf_matcher_type type, enum bf_matcher_op op,
-                  void *payload, const char *raw_payload)
+static int _bf_parse_uint(void *payload, const char *raw_payload,
```
qdeslandes: This function is named similarly to every other parsing callback, which is confusion as it can't be used as a parsing callback (not directly at least). Maybe we could have `_bf_parse_uX` (8, 16, 32) which calls `_bf_parse_integer` which perform the heavy lifting.

src/libbpfilter/matcher.c
```diff
@@ -64,32 +65,50 @@ struct bf_matcher
...
+    return 0;
+}
+
+int _bf_parse_int(enum bf_matcher_type type, enum bf_matcher_op op,
+                  void *payload, const char *raw_payload)
+{
+    assert(payload);
+    assert(raw_payload);
+
+    if (!_bf_parse_uint(payload, raw_payload, UINT32_MAX))
```
qdeslandes: `_bf_parse_int` should allow strings of negative values.

src/libbpfilter/matcher.c
```diff
@@ -263,11 +253,10 @@ int _bf_parse_l4_proto(enum bf_matcher_type type, enum bf_matcher_op op,
         return 0;
     }
```
qdeslandes: `_bf_parse_uint`?

src/libbpfilter/matcher.c
qdeslandes: `_bf_parse_uint` should not infer the size of `payload` from `max`.

src/libbpfilter/matcher.c
yaakov-stein: I was debating whether adding hex functionality (+ tests and docs) to everything that uses `l4_proto` was beyond the scope of the PR. But agree that it is kind of a drive by fix, so added it as its own commit.

src/libbpfilter/matcher.c
yaakov-stein: We actually don't have any parsing/printing that should allow negative values, so these are renamed to u32 for clarity.

PR #485, opened 2026-03-20
[0/2] Refactor Logging Infrastructure

yaakov-stein:
Stack:
- -> #485
- #486
- #491

### Summary

This sets the stage for adding logging support for `CGROUP_SOCK_ADDR`. The first step is renaming the logging infrastructure in a way that isn't packet-specific (or making it clear that certain parts are packet-specific and naming accordingly), and adding a `union` that allows us to add the necessary fields to `bf_log` to support this change.

No functional change (besides the `bf_log_type` enum helper methods).

### DIFF
--- a/src/libbpfilter/include/bpfilter/runtime.h
+++ b/src/libbpfilter/include/bpfilter/runtime.h
@@ -7,10 +7,23 @@
 
 #define bf_aligned(x) __attribute__((aligned(x)))
 
+/**
+ * @brief Give an anonymous union or struct a name only for Doxygen.
+ *
+ * Allows anonymous unions/structs in code while keeping Doxygen's parser happy.
+ */
+#ifdef DOXYGEN
+#define BF_ANONYMOUS_MEMBER(name) name
+#else
+#define BF_ANONYMOUS_MEMBER(name)
+#endif
+
 // _Static_assert doesn't exist in C++
 #ifndef __cplusplus
 #define static_assert _Static_assert
 #endif
+
+#include <linux/in6.h>
 
 #include <asm/types.h>
 
@@ -50,6 +63,9 @@ static_assert(BF_L3_SLICE_LEN % 8 == 0,
 static_assert(BF_L4_SLICE_LEN % 8 == 0,
               "BF_L4_SLICE_LEN should be aligned to 8 bytes");
 
+/** Size of the process name buffer, matches TASK_COMM_LEN. */
+#define BF_COMM_LEN 16
+
 /**
  * @brief Types of network packet headers.
  */
@@ -78,35 +94,28 @@ enum bf_pkthdr
 };
 
 /**
- * @brief Log structure published by a chain when the `log` action is hit.
- *
- * The structure is published into a log buffer by the chain, when a hit rule
- * has a `log` action defined.
- *
- * Except for the raw packet headers (`l2hdr`, `l3hdr`, and `l4hdr`), all the
- * values are stored in host byteorder.
+ * @brief Log entry type discriminator.
  */
-struct bf_log
+enum bf_log_type
 {
-    /** Timestamp of the packet processing. */
-    __u64 ts;
+    /** Packet-based log entry (XDP, TC, NF, cgroup_skb). */
+    BF_LOG_TYPE_PACKET,
 
+    /** Socket address log entry (cgroup_sock_addr). */
+    BF_LOG_TYPE_SOCK_ADDR,
+
+    _BF_LOG_TYPE_MAX,
+};
+
+/**
+ * @brief Packet log payload fields (XDP, TC, NF, cgroup_skb).
+ */
+struct bf_log_pkt
+{
     /** Total size of the packet, including the payload. */
     __u64 pkt_size;
 
-    /** ID of the rule triggering the log. */
-    __u32 rule_id;
-
-    /** Verdict of the rule triggering the log. */
-    __u32 verdict;
-
-    /** Layer 3 (internet) protocol identifier. */
-    __u16 l3_proto;
-
-    /** Layer 4 (transport) protocol identifier. */
-    __u8 l4_proto;
-
-    /** User-request headers, as defined in the rule. */
+    /** User-requested headers, as defined in the rule. */
     __u8 req_headers:4;
 
     /** Logged headers, as not all hooks can access all headers. */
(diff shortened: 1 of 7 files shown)


src/libbpfilter/include/bpfilter/runtime.h
```diff
@@ -106,20 +123,53 @@ struct bf_log
...
-    bf_aligned(8) __u8 l3hdr[BF_L3_SLICE_LEN];
+            /** Source address (4 bytes for IPv4, 16 for IPv6). */
+            bf_aligned(8) __u8 saddr[sizeof(struct in6_addr)];
 
-    /** Layer 4 header. */
-    bf_aligned(8) __u8 l4hdr[BF_L4_SLICE_LEN];
+            /** Destination address (4 bytes for IPv4, 16 for IPv6). */
+            bf_aligned(8) __u8 daddr[sizeof(struct in6_addr)];
+        } sock_addr;
+    } payload;
```
yaakov-stein: I'd ideally have left this unnamed as there isn't much of a benefit of naming it and it now requires `->payload.pkt/sock_addr`, but doxygen requires unions to be named :(

src/libbpfilter/include/bpfilter/runtime.h
```diff
@@ -50,31 +52,91 @@ static_assert(BF_L3_SLICE_LEN % 8 == 0,
...
+    /** Socket address log entry (cgroup_sock_addr). */
+    BF_LOG_TYPE_SOCK_ADDR,
+
+    _BF_LOG_TYPE_MAX,
+};
+
+/**
+ * @brief Packet log payload fields (XDP, TC, NF, cgroup_skb).
+ */
+struct _bf_log_pkt
```
yaakov-stein: This are only necessary internally but live in a public header file... I went with the approach of using the `_bf` prefix, but there weren't any examples to follow. Let me know if there's a better appraoch (maybe defining the structs directly inside the union?).

src/libbpfilter/include/bpfilter/runtime.h
qdeslandes: Leave it unnamed, we'll find a way to deal with the Doxygen issue differently.

src/libbpfilter/include/bpfilter/runtime.h
qdeslandes: Why is it only internal? This file can be included by clients reading from the ring buffer. I would suggest removing the `_` prefix, as it's reserved in C. It's fine to use it in our own translation unit, as we control includes and naming, but it's not for public headers.

src/libbpfilter/include/bpfilter/runtime.h
```diff
@@ -77,23 +82,81 @@ enum bf_pkthdr
...
+{
+    /** Total size of the packet, including the payload. */
+    __u64 pkt_size;
+
+    /** User-requested headers, as defined in the rule. */
+    __u8 req_headers;
+
+    /** Logged headers, as not all hooks can access all headers. */
+    __u8 headers;
```
qdeslandes: Why did you remove the `:4` bitfield?

src/libbpfilter/rule.c
```diff
@@ -19,28 +19,56 @@
...
+    return _bf_log_opt_strs[opt];
+}
+
+int bf_log_opt_from_str(const char *str, enum bf_log_opt *opt)
+{
+    assert(opt);
+
+    for (int i = 0; i < _BF_LOG_OPT_MAX; ++i) {
+        if (bf_streq_i(str, _bf_log_opt_strs[i])) {
+            *opt = (enum bf_log_opt)i;
```
qdeslandes: I don't think we need sock_addr logging to be configurable, there isn't that much data to log, and the only helper we'll use do `current->tgid << 32 | current->pid`. Hence, we can keep `bf_pkthdr` for packet logging options.

src/libbpfilter/rule.c
```diff
@@ -19,28 +19,56 @@
...
-            *hdr = (enum bf_pkthdr)i;
+    for (int i = 0; i < _BF_LOG_TYPE_MAX; ++i) {
+        if (bf_streq_i(str, _bf_log_type_strs[i])) {
+            *log_type = (enum bf_log_type)i;
+            return 0;
+        }
+    }
+
+    return -EINVAL;
+}
```
qdeslandes: We don't need this as we don't use it.

qdeslandes (changes requested):

src/libbpfilter/rule.c
yaakov-stein: Hm, currently in the final PR I'm allowing for internet, transport, pid, comm options. All enabled looks like:
Screenshot 2026-03-29 at 3 10 01 PM

src/libbpfilter/include/bpfilter/runtime.h
yaakov-stein: Related to the comment about adding the new pid, comm options, we had over four options. Even though we never need pid/comm for packet flavors, I thought it might be a good idea to expand it to avoid future bugs (can be a subtle source for bugs in the future if we do make changes), and it also is less BPF instructions (IIUC). It also, both before and after the change, require padding, so it doesn't make any difference in terms of actual space (at the moment atleast). I'll switch it back for now though.

src/libbpfilter/include/bpfilter/runtime.h
yaakov-stein: Ok, using this for now, can iterate further as well:
```
/**
 * @brief Give an anonymous union or struct a name only for Doxygen.
 *
 * Allows anonymous unions/structs in code while keeping Doxygen's parser happy.
 */
#ifdef DOXYGEN
#define ANONYMOUS_MEMBER(name) name
#else
#define ANONYMOUS_MEMBER(name)
#endif
```
This is a bit hacky but let's doxygen think there is a name while not actually having a name on the struct.

src/libbpfilter/include/bpfilter/runtime.h
```diff
@@ -7,11 +7,24 @@
...
+/**
+ * @brief Give an anonymous union or struct a name only for Doxygen.
+ *
+ * Allows anonymous unions/structs in code while keeping Doxygen's parser happy.
+ */
+#ifdef DOXYGEN
+#define ANONYMOUS_MEMBER(name) name
+#else
+#define ANONYMOUS_MEMBER(name)
+#endif
```
qdeslandes, line 19: As it's defined in a public header, this should be prefixed with `BF_`.

src/libbpfilter/rule.c
qdeslandes: I mean that there is so little to log for sock_addr programs that it's not worth allowing users to pick which field to log. We should log all the fields at once instead.

qdeslandes (changes requested):

qdeslandes (approved):

PR #486, opened 2026-03-20
[1/2] Add flavor-specific log codegen callback

yaakov-stein:
Gave `stack-pr` a try, but it doesn't allow for multi-commit PR's :(

Stack:
- #485
- -> #486
- #491

### Summary

The main goal here is to add a callback that gives flavors the ability to more finely control the logging process. This will enable us to easily use a new `CGROUP_SOCK_ADDR` elf-stub.

### Notes
- Moves `cgen/matcher/packet.{c,h}` to `cgen/packet.{c,h}`. It was a mistake to put this here in the first place as packet-based flavors can share more than matcher codegen. We correct the placement now.
- Once again, no functional change. This moves a file around and add a callback.

### DIFF
--- a/src/libbpfilter/cgen/cgroup_sock_addr.c
+++ b/src/libbpfilter/cgen/cgroup_sock_addr.c
@@ -280,9 +280,20 @@ static int _bf_cgroup_sock_addr_get_verdict(enum bf_verdict verdict,
     }
 }
 
+static int _bf_cgroup_sock_addr_gen_inline_log(struct bf_program *program,
+                                               const struct bf_rule *rule)
+{
+    (void)program;
+    (void)rule;
+
+    return bf_err_r(-ENOTSUP,
+                    "logging is not yet supported for cgroup_sock_addr");
+}
+
 const struct bf_flavor_ops bf_flavor_ops_cgroup_sock_addr = {
     .gen_inline_prologue = _bf_cgroup_sock_addr_gen_inline_prologue,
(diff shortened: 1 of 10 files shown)


src/libbpfilter/cgen/cgroup_sock_addr.c
```diff
@@ -273,9 +273,23 @@ static int _bf_cgroup_sock_addr_get_verdict(enum bf_verdict verdict)
...
 }
 
+static int _bf_cgroup_sock_addr_gen_inline_log(struct bf_program *program,
+                                               const struct bf_rule *rule)
+{
+    assert(program);
+    assert(rule);
+
+    (void)program;
+    (void)rule;
```
qdeslandes: No need for `assert(X)` and `(void)X`, even if temporary.

qdeslandes (changes requested):

src/libbpfilter/cgen/cgroup_sock_addr.c
yaakov-stein: Ya, agreed. The `(void)`'s are needed to avoid an unused warning, but we can hold off on the `assert`s

qdeslandes (dismissed):

qdeslandes (changes requested):
Could you rebase your changes on main? It will discard the commits from PR #1, then it should be good to go.

qdeslandes (approved):
LGTM! Thank you!

PR #487, opened 2026-03-24
lib: add bf_vector and use it in cgen

pzmarzly:
I initially planned to use bf_vector in bf_hashset, but I got convinced by @yaakov-stein that it's not a good idea. However, we could use bf_vector in other places, especially where we use bf_dynbuf or where we handroll vector-like behavior.

Commits:
- `lib: core: add bf_vector` - implementation. Vector doubles in size when it hits the limits. Has a helper to remove elements as well, though shrinking is not implemented.
- `cgen: use bf_vector for img` - Instead of managing `img`, `img_size`, and `img_cap` by hand, use bf_vector.


src/libbpfilter/include/bpfilter/request.h
```diff
@@ -64,8 +65,7 @@ struct bf_request;
 int bf_request_new(struct bf_request **request, enum bf_request_cmd cmd,
                    const void *data, size_t data_len);
 
-int bf_request_new_from_dynbuf(struct bf_request **request,
-                               struct bf_dynbuf *dynbuf);
+int bf_request_new_from_vector(struct bf_request **request, bf_vector *vec);
```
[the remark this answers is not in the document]
pzmarzly: > libbpfilter has external consumers

It's very unlikely they use it. Beside, we are the ones importing the change into monorepo.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,163 @@
...
+ */
+#define bf_vector_foreach(vec, elem)                                           \
+    for (void *(elem) = (vec)->data;                                           \
+         (elem) && (elem) < (vec)->data + ((vec)->size * (vec)->elem_size);    \
+         (elem) = (elem) + (vec)->elem_size)
+
+/**
+ * @brief Allocate and initialise a new vector on the heap.
+ *
+ * @param vec Pointer to the vector pointer. Must be non-NULL. On failure,
```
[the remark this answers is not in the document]
pzmarzly: Repo is already non-C17 compliant in other places, see one of the comment in https://github.com/facebook/bpfilter/pull/470 about it. Dismissing for now, in the future maybe we'll do something about it.

tests/unit/libbpfilter/request.c
```diff
@@ -39,44 +39,44 @@ static void new_request(void **state)
...
     assert_string_equal(bf_request_data(request), data);
 
-    // dynbuf should now be empty after taking ownership
-    assert_null(dynbuf.data);
+    // vector should now be empty after taking ownership
+    assert_null(vec.data);
 }
 
-static void new_from_dynbuf_invalid(void **state)
+static void new_from_vector_invalid(void **state)
```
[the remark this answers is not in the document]
pzmarzly: We assume people will use this method according to its API described in `.h`. I don't think we need to test all unhappy paths.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,164 @@
...
+ * @return A zero-initialized `bf_vector`.
+ */
+#define bf_vector_default(esz)                                                 \
+    ((bf_vector) {.data = NULL, .size = 0, .cap = 0, .elem_size = (esz)})
+
+/**
+ * @brief Iterate over every element of a `bf_vector`.
+ *
+ * `elem` is declared as a pointer to the element type and will point to each
+ * element in turn. Do not add or remove elements during iteration.
```
qdeslandes, line 56: > Do not add

This should not be an issue.

> remove elements during iteration

Same here, as we don't shrink the memory buffer.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,164 @@
...
+ * @param vec Pointer to the vector. Must be non-NULL.
+ */
+void bf_vector_clean(bf_vector *vec);
+
+/**
+ * @brief Get a pointer to the n-th element.
+ *
+ * @param vec Initialised vector. Must be non-NULL.
+ * @param index Index of the element.
+ * @return Pointer to the element, or NULL if @p index is out of bounds.
```
qdeslandes:
```
@p index -> `index`
```

As a general rule, use backticks for parameters, variables, functions...

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,164 @@
...
+ */
+void *bf_vector_get(const bf_vector *vec, size_t index);
+
+/**
+ * @brief Overwrite the n-th element.
```
qdeslandes: "Replace the n-th element" would be clearer.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,164 @@
...
+ * @param cap Minimum number of elements the vector should be able to hold.
+ * @return 0 on success, or a negative errno value on failure.
+ */
+int bf_vector_reserve(bf_vector *vec, size_t cap);
+
+/**
+ * @brief Take ownership of the backing buffer.
+ *
+ * Returns the raw data pointer and resets the vector so it will
+ * re-allocate on the next add.
```
qdeslandes: "on the next `bf_vector_add`"

src/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,201 @@
...
+}
+
+static int _bf_vector_realloc(bf_vector *vec, size_t new_cap)
+{
+    size_t alloc_size;
+    int r;
+
+    assert(vec);
+
+    if (!vec->elem_size)
```
qdeslandes: This seems off as we never check for `vec->elem_size`. Either we assume the user might modify `vec->elem_size` (because it has access to `bf_vector` structure), in which case we need to validate it in each function, or `bf_vector` should be moved into `vector.c` (I would suggest this).

src/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,201 @@
...
+        return -EINVAL;
+
+    if (__builtin_mul_overflow(new_cap, vec->elem_size, &alloc_size))
+        return -ENOMEM;
```
qdeslandes: `return -EOVERFLOW`

src/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,201 @@
...
+
+    assert(vec);
+    assert(elem);
+
+    if (vec->size == vec->cap) {
+        size_t new_cap;
+
+        new_cap = vec->cap ? vec->cap * 2 : _BF_VECTOR_INIT_CAP;
+        if (new_cap > _BF_VECTOR_MAX_CAP)
+            return -ENOMEM;
```
qdeslandes: This will be repetitive for functions adding data to the vector. It would be better to have `_bf_vector_grow(vector, n_new_elems)`. `_bf_vector_grow` would handle the size calculation, and the growth.

src/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,201 @@
...
+        return -ENOMEM;
+
+    r = bf_realloc(&vec->data, alloc_size);
+    if (r)
+        return r;
+
+    vec->cap = new_cap;
+
+    return 0;
+}
```
qdeslandes: This could be `_bf_vector_grow(vec, n_elems_to_add)`, to deduplicate the logic from `bf_vector_add` and `bf_vector_add_many`. It should:
- Calculate the total number of elements the vector should container (`vec->size + n_elems_to_add`)
- Calculate the required capacity
- Resize the vector accordingly (handle init if vector is empty, and overflows)

Also, [instead of 2, a growth factor of 1.5 would be more suited](https://github.com/facebook/folly/blob/main/folly/docs/FBVector.md).

tests/unit/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,312 @@
...
+    {
+        // Allocate and free
+        bf_vector *vec;
+
+        assert_ok(bf_vector_new(&vec, sizeof(int)));
+        assert_int_equal(vec->size, 0);
+        assert_int_equal(vec->cap, 0);
+        assert_null(vec->data);
+
+        bf_vector_free(&vec);
```
qdeslandes, line 24: Call `bf_vector_free()` twice to ensure it's safe (it should be).

tests/unit/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,312 @@
...
+
+    (void)state;
+
+    bf_vector_foreach (&vec, elem) {
+        (void)elem;
+        ++count;
+    }
+
+    assert_int_equal(count, 0);
+}
```
qdeslandes: You can test this in the `foreach` test, before filling the vector.

tests/unit/libbpfilter/core/vector.c
```diff
@@ -0,0 +1,312 @@
...
+    assert_ok(bf_vector_add_many(&vec, chunk2, sizeof(chunk2)));
+    assert_int_equal(vec.size, 5);
+
+    uint8_t *p = vec.data;
+    assert_int_equal(p[0], 0xaa);
+    assert_int_equal(p[1], 0xbb);
+    assert_int_equal(p[2], 0xcc);
+    assert_int_equal(p[3], 0xdd);
+    assert_int_equal(p[4], 0xee);
+}
```
qdeslandes: What is the benefit of this test, compared to `add_many`?

src/libbpfilter/cgen/jmp.c
```diff
@@ -18,8 +18,14 @@
...
-        struct bpf_insn *insn = &ctx->program->img[ctx->insn_idx];
-        size_t off = ctx->program->img_size - ctx->insn_idx - 1U;
+        struct bpf_insn *insn =
+            bf_vector_get(&ctx->program->img, ctx->insn_idx);
+        if (!insn) {
+            bf_abort("jump fixup references invalid instruction index %lu",
+                     ctx->insn_idx);
+        }
+
+        size_t off = ctx->program->img.size - ctx->insn_idx - 1U;
```
qdeslandes: This variable should be defined at the beginning of the scope.

src/libbpfilter/cgen/program.c
```diff
@@ -107,6 +106,7 @@ int bf_program_new(struct bf_program **program, const struct bf_chain *chain,
     _program->flavor = bf_hook_to_flavor(chain->hook);
     _program->runtime.ops = bf_flavor_ops_get(_program->flavor);
     _program->runtime.chain = chain;
+    _program->img = bf_vector_default(sizeof(struct bpf_insn));
```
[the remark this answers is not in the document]
qdeslandes: +1 with Claude, [programs are at least 200 instructions](https://bpfilter.io/external/benchmarks/index.html), so we can reserve `512` elements.

src/libbpfilter/cgen/program.c
```diff
@@ -474,19 +470,9 @@ static int _bf_program_generate_elfstubs(struct bf_program *program)
...
-        r = bf_program_grow_img(program);
-        if (r)
-            return r;
-    }
-
-    program->img[program->img_size++] = insn;
-
-    return 0;
+    return bf_vector_add(&program->img, &insn);
 }
```
qdeslandes: `static inline` in `program.h`.

src/libbpfilter/cgen/program.c
```diff
@@ -535,8 +519,8 @@ int bf_program_emit_fixup(struct bf_program *program, enum bf_fixup_type type,
 
     /* This call could fail and return an error, in which case it is not
      * properly handled. However, this shouldn't be an issue as we previously
-     * test whether enough room is available in cgen.img, which is currently
-     * the only reason for EMIT() to fail. */
+     * reserved enough room in program->img, which is currently the only
+     * reason for EMIT() to fail. */
     EMIT(program, insn);
```
qdeslandes: The original code is weird in order to ensure the state of `bf_program` is always valid (i.e. we don't insert a fixup for an instruction that doesn't exist).

We don't need to maintain a valid `bf_program` state anymore, as it doesn't survive past the generation step. Hence, you can push the fixup to the list, then use `bf_program_emit()`, and return an error on failure. It will be properly taken care of by bpfilter.

src/libbpfilter/cgen/program.c
qdeslandes, line 550: Same as `bf_program_emit_fixup`.

qdeslandes (changes requested):

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,164 @@
...
+ * @param vec Pointer to the vector. Must be non-NULL.
+ * @param elem Name of the iteration variable. Will be declared as a
+ *        `void *` and cast by the caller.
+ */
+#define bf_vector_foreach(vec, elem)                                           \
+    for (void *(elem) = (vec)->data;                                           \
+         (elem) && (elem) < (vec)->data + ((vec)->size * (vec)->elem_size);    \
+         (elem) = (elem) + (vec)->elem_size)
+
+/**
```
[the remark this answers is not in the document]
pzmarzly: We could do that, but the codebase is already not C17 compliant in different places as well.

src/libbpfilter/include/bpfilter/core/vector.h
pzmarzly: > Do not add

If that addition causes growth/reallocation, `void *(elem) = (vec)->data;` will point to the old memory address. I could turn the iterator into index `int i = 0;` and use double-`for` trick to have `void *(elem) = (vec)->data + i * (vec)->elem_size;`, but I think it's an overkill?

> remove elements during iteration

But we shift the next elements left, i.e. 5th element becomes 4th, 4th becomes 3rd, etc. The foreach will skip an element because of `(elem) = (elem) + (vec)->elem_size`.

tests/unit/libbpfilter/core/vector.c
pzmarzly: I wanted to test that vec.data is a regular void* array, with no padding etc. But `take` test also exercises that.

src/libbpfilter/cgen/program.c
```diff
@@ -539,10 +516,6 @@ int bf_program_emit_fixup(struct bf_program *program, enum bf_fixup_type type,
 
     TAKE_PTR(fixup);
 
-    /* This call could fail and return an error, in which case it is not
-     * properly handled. However, this shouldn't be an issue as we previously
-     * test whether enough room is available in cgen.img, which is currently
-     * the only reason for EMIT() to fail. */
     EMIT(program, insn);
```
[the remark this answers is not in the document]
pzmarzly: `EMIT` before `bf_fixup_new` is pretty smart.

src/libbpfilter/cgen/program.c
```diff
@@ -523,13 +506,9 @@ int bf_program_emit_fixup(struct bf_program *program, enum bf_fixup_type type,
 
     assert(program);
 
-    if (program->img_size == program->img_cap) {
-        r = bf_program_grow_img(program);
-        if (r)
-            return r;
-    }
+    EMIT(program, insn);
```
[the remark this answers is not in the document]
pzmarzly: In daemonless version, as you said, the program is abandoned anyway on failure.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @param vec Pointer to the vector. Must be non-NULL.
+ * @param elem Name of the iteration variable. Will be declared as a
+ *        `void *` and cast by the caller.
+ */
+#define bf_vector_foreach(vec, elem)                                           \
+    for (void *(elem) = (vec)->data,                                           \
+              *__end = (vec)->data ?                                           \
+                           (vec)->data + ((vec)->size * (vec)->elem_size) :    \
+                           NULL;                                               \
```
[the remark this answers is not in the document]
pzmarzly: Duplicate of previous comment.

src/libbpfilter/include/bpfilter/core/vector.h
```diff
@@ -0,0 +1,174 @@
...
+ * `bf_vector_default`) or discarded.
+ *
+ * @param vec Pointer to the vector. Must be non-NULL.
+ */
+void bf_vector_clean(bf_vector *vec);
+
+/**
+ * @brief Get a pointer to the n-th element.
+ *
+ * @param vec Initialised vector. Must be non-NULL.
```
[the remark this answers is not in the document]
pzmarzly: They should call `init` / `default` nonetheless. We can be permissive in case they forgot, but then we shouldn't document it.

src/libbpfilter/include/bpfilter/core/vector.h
qdeslandes: Good catch. I don't think there's a big use case to modify the vector while iterating, so not worth the effort.

src/libbpfilter/include/bpfilter/core/vector.h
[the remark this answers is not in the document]
qdeslandes: The public headers are C17 compliant (as tested by the integration tests). We can fix this in a follow-up change to prevent delays. #509

qdeslandes (approved):
LGTM! Thanks!

PR #493, opened 2026-03-31
Add `CGROUP_SOCK_ADDR` Set Support

yaakov-stein:
## Summary

Before this change, sets (hash maps and LPM tries used for bulk matching) were only supported on packet-based hooks.  This meant users who wanted to filter against a list of IPs or ports on a `CONNECT/SENDMSG` hook had to create one rule per value, which scales poorly.

This change adds set support for all four `CGROUP_SOCK_ADDR` hooks, reusing the same BPF map infrastructure (hash maps for exact match, LPM tries for prefix match) that packet hooks already use.

### Refactoring: make set codegen composable/generic

Set codegen previously lived entirely in `set.c` and was tightly coupled to packet-based header loading. To reuse it for `CGROUP_SOCK_ADDR` hooks, we extracted the generic pieces (copy field to scratch, map/trie lookup) into shared helpers and moved the packet-specific logic into `packet.c`. Protocol checks and layer conflict detection were also hoisted out of set codegen into `program.c`'s existing dedup loop, so they now cover set components the same way they cover individual matchers.

As a drive-by, set components are now validated individually against the hook's supported matchers.

### `CGROUP_SOCK_ADDR` set codegen

With the shared helpers in place, `CGROUP_SOCK_ADDR` set codegen maps each key component to its `bpf_sock_addr` context field offset, copies the field to the scratch buffer, and calls the shared map lookup. No header parsing needed — `r6` already points to the context.

## Why can't we use the current set codegen?

Packet-based hooks receive a pointer to raw packet memory. The BPF program parses headers layer by layer (L2 -> L3 -> L4), and `r6` points to the current header being inspected.

`CGROUP_SOCK_ADDR` hooks receive a `bpf_sock_addr` context struct instead. The kernel pre-extracts socket metadata into typed fields:

```
struct bpf_sock_addr {
    __u32 user_ip4;        // destination IPv4
    __u32 user_ip6[4];     // destination IPv6
    __u32 user_port;       // destination port (network order)
    __u32 msg_src_ip4;     // source IPv4  (sendmsg only)
    __u32 msg_src_ip6[4];  // source IPv6  (sendmsg only)
    ...
};
```

This difference has two consequences for set codegen:

1. **No header parsing needed.** `r6` already points to the context. Instead of loading a header pointer and extracting fields from parsed packet data, we read directly from context field offsets (`r6 + offsetof(bpf_sock_addr, field)`).

2. **BPF verifier constraints on access width.** Packet memory can be read at any width. Context struct fields must be read at widths that respect field alignment. The verifier rejects misaligned reads from context pointers. The new `bf_stub_load` function handles this by picking the largest safe access width based on both source and destination alignment.

The diagram below shows the two paths:

```
Packet hooks (XDP/TC/NF/cgroup_skb):
  ┌──────────────┐     ┌──────────────┐     ┌──────────────┐
  │ Parse L2/L3  │────>│ Load header  │────>│ Copy field   │───> map lookup
  │ (dynptr)     │     │ ptr into r6  │     │ to scratch   │
  └──────────────┘     └──────────────┘     └──────────────┘

CGROUP_SOCK_ADDR:
  ┌──────────────────┐     ┌──────────────┐
  │ r6 = ctx (set    │────>│ Copy field   │───> map lookup
  │ once in prologue)│     │ to scratch   │
  └──────────────────┘     └──────────────┘
```

## Test plan

- **Unit tests**: validation of set component hook compatibility and cross-layer conflict detection
- **E2e tests**: each `CGROUP_SOCK_ADDR` hook has dry-run parsing tests and behavioral tests covering hash sets, trie sets, and multi-component sets

### DIFF
--- a/doc/usage/bfcli.rst
+++ b/doc/usage/bfcli.rst
@@ -360,7 +360,7 @@ With:
 
   .. note::
 
-      ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks operate on socket metadata rather than packet data. Supported matchers: ``meta.l3_proto``, ``meta.l4_proto``, ``meta.probability``, ``meta.dport``, ``ip4.daddr``, ``ip4.dnet``, ``ip4.proto``, ``ip6.daddr``, ``ip6.dnet``, ``udp.dport``. Connect hooks additionally support ``tcp.dport``. Sendmsg hooks additionally support ``ip4.saddr``, ``ip4.snet``, ``ip6.saddr``, and ``ip6.snet``.
+      ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks operate on socket metadata rather than packet data. Supported matchers: ``meta.l3_proto``, ``meta.l4_proto``, ``meta.probability``, ``meta.dport``, ``ip4.daddr``, ``ip4.dnet``, ``ip4.proto``, ``ip6.daddr``, ``ip6.dnet``, ``udp.dport``. Connect hooks additionally support ``tcp.dport``. Sendmsg hooks additionally support ``ip4.saddr``, ``ip4.snet``, ``ip6.saddr``, and ``ip6.snet``. Sets are supported with the same per-component restrictions: each set key component must be a matcher supported by the chain's hook. Note that ``meta.dport`` and ``meta.sport`` are not supported in sets.
 
   - ``$POLICY``: action taken if no rule matches the packet:
 
@@ -667,16 +667,21 @@ IPv6
       - Operator
       - Payload
       - Notes
-    * - :rspan:`1` Source address
-      - :rspan:`1` ``ip6.saddr``
+    * - :rspan:`2` Source address
+      - :rspan:`2` ``ip6.saddr``
       - ``eq``
-      - :rspan:`3` ``$ADDR``
-      - :rspan:`3` ``$ADDR`` must be an IPv6 address composed of 8 hexadecimal numbers (abbreviations are supported). To filter on an IPv6 network (using an IPv6 address and a subnet mask), see ``ip6.snet`` or ``ip6.dnet``.
+      - :rspan:`1` ``$ADDR``
+      - :rspan:`5` ``$ADDR`` must be an IPv6 address composed of 8 hexadecimal numbers (abbreviations are supported). To filter on an IPv6 network (using an IPv6 address and a subnet mask), see ``ip6.snet`` or ``ip6.dnet``.
     * - ``not``
-    * - :rspan:`1` Destination address
-      - :rspan:`1` ``ip6.daddr``
+    * - ``in``
+      - ``{$ADDR[,...]}``
+    * - :rspan:`2` Destination address
+      - :rspan:`2` ``ip6.daddr``
       - ``eq``
+      - :rspan:`1` ``$ADDR``
     * - ``not``
+    * - ``in``
+      - ``{$ADDR[,...]}``
     * - :rspan:`2` Source network
       - :rspan:`2` ``ip6.snet``
       - ``eq``
--- a/src/libbpfilter/cgen/cgroup_sock_addr.c
+++ b/src/libbpfilter/cgen/cgroup_sock_addr.c
@@ -14,15 +14,19 @@
 #include <stdint.h>
 #include <sys/socket.h>
 
+#include <bpfilter/chain.h>
 #include <bpfilter/flavor.h>
 #include <bpfilter/logger.h>
 #include <bpfilter/matcher.h>
 #include <bpfilter/runtime.h>
+#include <bpfilter/set.h>
 #include <bpfilter/verdict.h>
 
 #include "cgen/matcher/cmp.h"
 #include "cgen/matcher/meta.h"
+#include "cgen/matcher/set.h"
 #include "cgen/program.h"
+#include "cgen/stub.h"
 #include "cgen/swich.h"
 #include "filter.h"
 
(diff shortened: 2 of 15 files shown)


doc/usage/bfcli.rst
```diff
@@ -647,16 +647,21 @@ IPv6
       - Operator
       - Payload
       - Notes
-    * - :rspan:`1` Source address
-      - :rspan:`1` ``ip6.saddr``
+    * - :rspan:`2` Source address
```
yaakov-stein, line 670: drive-by fix (document `ip6.d/saddr` works in sets)

src/libbpfilter/chain.c
```diff
@@ -60,6 +60,41 @@ static int _bf_rule_references_empty_set(const struct bf_chain *chain,
...
+ */
+static int _bf_rule_check_layer(enum bf_matcher_type type,
+                                uint32_t layer_hdr_id[_BF_MATCHER_LAYER_MAX])
+{
+    const struct bf_matcher_meta *meta;
+
+    assert(layer_hdr_id);
+
+    meta = bf_matcher_get_meta(type);
+
```
qdeslandes: No empty line please.

src/libbpfilter/cgen/cgroup_sock_addr.c
```diff
@@ -204,6 +208,99 @@ static int _bf_cgroup_sock_addr_generate_port(struct bf_program *program,
...
+    default:
+        return SIZE_MAX;
+    }
+}
+
+static int _bf_cgroup_sock_addr_generate_set(struct bf_program *program,
+                                             const struct bf_matcher *matcher)
+{
+    assert(program);
+    assert(matcher);
```
qdeslandes: Assert after variables declaration.

src/libbpfilter/cgen/cgroup_sock_addr.c
```diff
@@ -204,6 +208,99 @@ static int _bf_cgroup_sock_addr_generate_port(struct bf_program *program,
                         bf_matcher_payload(matcher), 2, BPF_REG_1);
 }
 
+static size_t _bf_cgroup_sock_addr_ctx_offset(enum bf_matcher_type type)
```
qdeslandes: `ssize_t` would be more suited, so we can return `<0` on error.

tests/e2e/hooks/cgroup_sock_addr_connect4.sh
```diff
@@ -112,3 +122,19 @@ ${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_CONNE
...
+${FROM_NS} ${BFCLI} ruleset flush
+
+# ip4.dnet trie set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_CONNECT4{cgpath=${CGROUP_PATH}} ACCEPT rule (ip4.dnet) in { 10.0.0.0/8 } DROP"
+(! udp4_connect ${HOST_IP_ADDR} 9990)
+${FROM_NS} ${BFCLI} ruleset flush
+
+# (ip4.daddr, udp.dport) multi-component hash set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_CONNECT4{cgpath=${CGROUP_PATH}} ACCEPT rule (ip4.daddr, udp.dport) in { ${HOST_IP_ADDR}, 9990 } DROP"
+(! udp4_connect ${HOST_IP_ADDR} 9990)
```
qdeslandes, line 148: Please check counters values too

tests/e2e/hooks/cgroup_sock_addr_connect6.sh
```diff
@@ -125,3 +134,19 @@ ${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_CONNE
...
+
+# ip6.daddr hash set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_CONNECT6{cgpath=${CGROUP_PATH}} ACCEPT rule (ip6.daddr) in { ${HOST_IP6_ADDR} } DROP"
+(! udp6_connect ${HOST_IP6_ADDR} 9990)
+${FROM_NS} ${BFCLI} ruleset flush
+
+# ip6.dnet trie set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_CONNECT6{cgpath=${CGROUP_PATH}} ACCEPT rule (ip6.dnet) in { fd00::/64 } DROP"
+(! udp6_connect ${HOST_IP6_ADDR} 9990)
+${FROM_NS} ${BFCLI} ruleset flush
```
qdeslandes: Same as above, check counters values to ensure we hit the rule as expected.

tests/e2e/hooks/cgroup_sock_addr_sendmsg4.sh
```diff
@@ -114,3 +125,19 @@ ${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_SENDM
...
+# ip4.dnet trie set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_SENDMSG4{cgpath=${CGROUP_PATH}} ACCEPT rule (ip4.dnet) in { 10.0.0.0/8 } DROP"
+(! udp4_sendmsg ${HOST_IP_ADDR} 9990)
+${FROM_NS} ${BFCLI} ruleset flush
+
+# (ip4.saddr, udp.dport) multi-component hash set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_SENDMSG4{cgpath=${CGROUP_PATH}} ACCEPT rule (ip4.saddr, udp.dport) in { ${NS_IP_ADDR}, 9990 } DROP"
+(! udp4_sendmsg ${HOST_IP_ADDR} 9990)
+udp4_sendmsg ${HOST_IP_ADDR} 9991
+${FROM_NS} ${BFCLI} ruleset flush
```
qdeslandes: Check counters values.

tests/e2e/hooks/cgroup_sock_addr_sendmsg6.sh
```diff
@@ -115,3 +124,19 @@ ${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_SENDM
...
+# ip6.dnet trie set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_SENDMSG6{cgpath=${CGROUP_PATH}} ACCEPT rule (ip6.dnet) in { fd00::/64 } DROP"
+(! udp6_sendmsg ${HOST_IP6_ADDR} 9990)
+${FROM_NS} ${BFCLI} ruleset flush
+
+# (ip6.saddr, udp.dport) multi-component hash set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_SENDMSG6{cgpath=${CGROUP_PATH}} ACCEPT rule (ip6.saddr, udp.dport) in { ${NS_IP6_ADDR}, 9990 } DROP"
+(! udp6_sendmsg ${HOST_IP6_ADDR} 9990)
+udp6_sendmsg ${HOST_IP6_ADDR} 9991
+${FROM_NS} ${BFCLI} ruleset flush
```
qdeslandes: Check counters values.

tests/unit/libbpfilter/chain.c
```diff
@@ -244,6 +276,54 @@ static void policy_validation(void **state)
                             BF_VERDICT_REDIRECT, NULL, NULL));
 }
 
+static void set_component_unsupported_hook(void **state)
+{
+    (void)state;
```
qdeslandes: This should be after variables declaration.

doc/usage/bfcli.rst
```diff
@@ -360,7 +360,7 @@ With:
 
   .. note::
 
-      ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks operate on socket metadata rather than packet data. Supported matchers: ``meta.l3_proto``, ``meta.l4_proto``, ``meta.probability``, ``meta.dport``, ``ip4.daddr``, ``ip4.dnet``, ``ip4.proto``, ``ip6.daddr``, ``ip6.dnet``, ``udp.dport``. Connect hooks additionally support ``tcp.dport``. Sendmsg hooks additionally support ``ip4.saddr``, ``ip4.snet``, ``ip6.saddr``, and ``ip6.snet``.
+      ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks operate on socket metadata rather than packet data. Supported matchers: ``meta.l3_proto``, ``meta.l4_proto``, ``meta.probability``, ``meta.dport``, ``ip4.daddr``, ``ip4.dnet``, ``ip4.proto``, ``ip6.daddr``, ``ip6.dnet``, ``udp.dport``. Connect hooks additionally support ``tcp.dport``. Sendmsg hooks additionally support ``ip4.saddr``, ``ip4.snet``, ``ip6.saddr``, and ``ip6.snet``. Sets are supported with the same per-component restrictions: each set key component must be a matcher supported by the chain's hook.
```
qdeslandes: `meta.dport` and `meta.sport` are not supported in sets (yet).

qdeslandes (changes requested):
Good PR! Some minors nits.

qdeslandes (changes requested):
Sending the PR back to the author.

tests/e2e/hooks/cgroup_sock_addr_connect4.sh
yaakov-stein: As a drive-by, also added for the non-set cases

yaakov-stein:
@claude

src/libbpfilter/cgen/cgroup_sock_addr.c
```diff
@@ -204,6 +208,99 @@ static int _bf_cgroup_sock_addr_generate_port(struct bf_program *program,
...
+            return bf_err_r(
+                (int)ctx_off,
+                "set component '%s' not supported for cgroup_sock_addr",
+                bf_matcher_type_to_str(type));
+        }
+
+        /* For ports, `hdr_payload_size` is 2 but `user_port` is a
+         * 4-byte `__u32`. The BPF verifier rewrites this 2-byte narrow
+         * ctx read to extract the correct NBO port value. */
+        r = bf_stub_load(program, (size_t)ctx_off, meta->hdr_payload_size,
```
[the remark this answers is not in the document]
yaakov-stein: Nice catch, confirmed it in the kernel code and https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_CGROUP_SOCK_ADDR/

yaakov-stein:
@claude review

tests/unit/libbpfilter/chain.c
```diff
@@ -244,6 +276,54 @@ static void policy_validation(void **state)
...
+
+    uint32_t set_index = 0;
+
+    (void)state;
+
+    // Set component unsupported for hook: ip4.saddr on CONNECT4.
+    assert_ok(bf_set_new(&set, "s", key, ARRAY_SIZE(key)));
+    assert_ok(bf_set_add_elem(set, (uint8_t[4]) {10, 0, 0, 1}));
+    assert_ok(bf_list_add_tail(&sets, set));
+    set = NULL;
```
qdeslandes: `bf_list_push()`

tests/e2e/hooks/cgroup_sock_addr_connect4.sh
```diff
@@ -50,65 +60,106 @@ udp4_connect() {
...
 udp4_connect ${HOST_IP_ADDR} 9990
 (! udp4_connect ${HOST_IP_ADDR} 9991)
+test "$(get_counter c 0)" = "1"
+${FROM_NS} ${BFCLI} ruleset flush
+
+# ip4.daddr hash set
+${FROM_NS} ${BFCLI} chain set --from-str "chain c BF_HOOK_CGROUP_SOCK_ADDR_CONNECT4{cgpath=${CGROUP_PATH}} ACCEPT rule (ip4.daddr) in { ${HOST_IP_ADDR} } counter DROP"
+(! udp4_connect ${HOST_IP_ADDR} 9990)
+test "$(get_counter c 0)" = "1"
+${FROM_NS} ${BFCLI} ruleset flush
```
qdeslandes: FYI: flush is not required, as you use `chain set`, which replaces any existing chain with the same name. You can leave it as-is as the whole file does it this way, so it's consistent.

src/libbpfilter/cgen/program.c
```diff
@@ -279,12 +279,39 @@ static int _bf_program_fixup(struct bf_program *program,
...
+                                   enum bf_matcher_type type,
+                                   uint32_t *checked_layers)
+{
+    assert(program);
+    assert(checked_layers);
+
+    const struct bf_matcher_meta *meta = bf_matcher_get_meta(type);
+
+    if (!meta)
+        return bf_err_r(-EINVAL, "missing meta for matcher type %d", type);
```
qdeslandes, line 293:
```
const struct bf_matcher *meta;

assert(program);
assert(checked_layers);

meta = bf_matcher_get_meta(type);
if (!meta)
 ...
```

qdeslandes (changes requested):

tests/e2e/hooks/cgroup_sock_addr_connect4.sh
yaakov-stein: Good point, they do add some latency so I'll just remove them in the test commit as a drive-by fix.

yaakov-stein: I also noticed that using `bpftool` for gettting the counter makes the tests significantly slower. Don't think it's too important right now, but maybe something I'll look into finding a better solution at some point.

yaakov-stein:
@claude review

qdeslandes (approved):
LGTM! Thank you :)

PR #496, opened 2026-04-03
tests: remove C++ <format> include

qdeslandes:
Remove <format> include from test.hpp to prevent compilation failure on platform for which it's not yet available.


yaakov-stein:
Thanks for this and #497! Both were causing problems for me and was planning on looking into it, so nice to see it's already been solved :)

PR #499, opened 2026-04-07
lib,cli,tests,doc: Add negation operator: "not"

jordalgo:
This supports the ability to negate a matcher by simply adding a "not" in
front of the matcher op, e.g. `tcp.sport not eq 80`.

This works for all matcher operations.

This also replaces/removes the "not" matcher operation but maintains
backward compatibility as `tcp.sport not 80` is also supported syntax and
translates to `not eq`.

The primary use-case is so we can create matchers for flags where we want
to match against one particular flag AND not match against another for the
same rule, e.g. has SYN and does not have ACK.

Made two decisions in the design:
1. Went with "not" instead of "!" because matchers are written in English
   sentences and not code so it's more natural to use "not".
2. The placement of "not" is right before the matcher op so it doesn't read
   as the whole matcher is negated (this is relevant for protocol guards
   which are not negated)

### DIFF
--- a/doc/usage/bfcli.rst
+++ b/doc/usage/bfcli.rst
@@ -509,14 +509,14 @@ Matchers are defined such as:
 
 .. code:: shell
 
-    $TYPE [$OP] $PAYLOAD
+    $TYPE [not] [$OP] $PAYLOAD
 
 With:
   - ``$TYPE``: type of the matcher, defined which part of the processed network packet need to be compared against. All the exact matcher types are defined below.
+  - ``not``: optional comparison negation. When used, it prefixes an operator, e.g. ``tcp.sport not eq 80``, ``(ip4.saddr) not in foo``, or, with no operator, ``tcp.sport not 80`` (equivalent to ``tcp.sport not eq 80``).
   - ``$OP``: comparison operation, not all ``$TYPE`` of matchers support all the existing comparison operators:
 
-    - ``eq``: exact equality.
-    - ``not``: inequality.
+    - ``eq``: exact equality. This is the default if not ``$OP`` is provided.
     - ``any``: match the packet against a set of data defined as the payload. If any of the member of the payload set is found in the packet, the matcher is positive. For example, if you want to match all the ``icmp`` and ``udp`` packets: ``ip4.proto any icmp,udp``.
     - ``all``: match the packet against a set of data defined as the payload. If all the member of the payload set are found in the packet, the matcher is positive, even if the packet contains more than only the members defined in the payload. For example, to match all the packets containing *at least* the ``ACK`` TCP flag: ``tcp.flags all ACK``.
     - ``in``: matches the packet against a hashed set of reference values. Using the ``in`` operator is useful when the packet's data needs to be compared against a large set of different values. Let's say you want to filter 1000 different IPv4 addresses, you can either define 1000 ``ip4.saddr eq $IP`` matcher, in which case ``bpfilter`` will compare the packet against every IP one after the other. Or you can use ``(ip4.saddr) in {$IP0,IP1,...}`` in which case ``bpfilter`` will compare the packet's data against the hashed set as a whole in 1 operation.
@@ -553,21 +553,19 @@ Meta
       - ``eq``
       - ``$PROTOCOL``
       - ``$PROTOCOL`` must be a transport layer protocol name (e.g. "ICMP", case insensitive), or a valid decimal or hexadecimal `internet protocol number`_.
-    * - :rspan:`2` Source port
-      - :rspan:`2` ``meta.sport``
+    * - :rspan:`1` Source port
+      - :rspan:`1` ``meta.sport``
       - ``eq``
-      - :rspan:`1` ``$PORT``
-      - :rspan:`1` ``$PORT`` must be a valid decimal port number.
-    * - ``not``
+      - ``$PORT``
+      - ``$PORT`` must be a valid decimal port number.
     * - ``range``
       - ``$START-$END``
       - ``$START`` and ``$END`` are valid port values, as decimal integers.
-    * - :rspan:`2` Destination port
-      - :rspan:`2` ``meta.dport``
+    * - :rspan:`1` Destination port
+      - :rspan:`1` ``meta.dport``
       - ``eq``
-      - :rspan:`1` ``$PORT``
-      - :rspan:`1` ``$PORT`` must be a valid decimal port number.
-    * - ``not``
+      - ``$PORT``
+      - ``$PORT`` must be a valid decimal port number.
     * - ``range``
       - ``$START-$END``
       - ``$START`` and ``$END`` are valid port values, as decimal integers.
@@ -576,18 +574,16 @@ Meta
       - ``eq``
       - ``$PROBABILITY``
       - ``$PROBABILITY`` is a floating-point percentage value (i.e., within [0%, 100%], e.g., "50%" or "33.33%"). For ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks, probability applies per socket operation: each ``connect()`` or ``sendmsg()`` call is independently accepted or dropped.
-    * - :rspan:`1` Mark
-      - :rspan:`1` ``meta.mark``
+    * - Mark
+      - ``meta.mark``
       - ``eq``
-      - :rspan:`1` ``$MARK``
-      - :rspan:`1` ``$MARK`` must be a valid decimal or hexadecimal 32-bits value. Incompatible with ``BF_HOOK_XDP`` and ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks.
-    * - ``not``
-    * - :rspan:`2` Flow hash
-      - :rspan:`2` ``meta.flow_hash``
+      - ``$MARK``
+      - ``$MARK`` must be a valid decimal or hexadecimal 32-bits value. Incompatible with ``BF_HOOK_XDP`` and ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks.
+    * - :rspan:`1` Flow hash
+      - :rspan:`1` ``meta.flow_hash``
       - ``eq``
-      - :rspan:`1` ``$HASH``
-      - :rspan:`1` ``$HASH`` must be a decimal or hexadecimal 32-bits integer. Only compatible with ``BF_HOOK_TC_INGRESS`` and ``BF_HOOK_TC_EGRESS`` hooks.
-    * - ``not``
+      - ``$HASH``
+      - ``$HASH`` must be a decimal or hexadecimal 32-bits integer. Only compatible with ``BF_HOOK_TC_INGRESS`` and ``BF_HOOK_TC_EGRESS`` hooks.
     * - ``range``
       - ``$START-$END``
       - ``$START`` and ``$END`` must be decimal or hexadecimal 32-bits integers, with ``$START <= $END``. Only compatible with ``BF_HOOK_TC_INGRESS`` and ``BF_HOOK_TC_EGRESS`` hooks.
@@ -610,48 +606,42 @@ IPv4
       - Operator
       - Payload
       - Notes
-    * - :rspan:`2` Source address
-      - :rspan:`2` ``ip4.saddr``
+    * - :rspan:`1` Source address
+      - :rspan:`1` ``ip4.saddr``
       - ``eq``
-      - :rspan:`1` ``$ADDR``
-      - :rspan:`5` ``$ADDR`` is an IPv4 address in dotted-decimal format, "ddd.ddd.ddd.ddd", where ddd is a decimal number of up to three digits in the range 0 to 255. To filter on an IPv4 network (using an IPv4 address and a subnet mask), see ``ip4.snet`` or ``ip4.dnet``.
-    * - ``not``
+      - ``$ADDR``
+      - :rspan:`3` ``$ADDR`` is an IPv4 address in dotted-decimal format, "ddd.ddd.ddd.ddd", where ddd is a decimal number of up to three digits in the range 0 to 255. To filter on an IPv4 network (using an IPv4 address and a subnet mask), see ``ip4.snet`` or ``ip4.dnet``.
     * - ``in``
       - ``{$ADDR[,...]}``
-    * - :rspan:`2` Destination address
-      - :rspan:`2` ``ip4.daddr``
+    * - :rspan:`1` Destination address
+      - :rspan:`1` ``ip4.daddr``
       - ``eq``
-      - :rspan:`1` ``$ADDR``
-    * - ``not``
+      - ``$ADDR``
     * - ``in``
       - ``{$ADDR[,...]}``
-    * - :rspan:`2` Source network
-      - :rspan:`2` ``ip4.snet``
+    * - :rspan:`1` Source network
+      - :rspan:`1` ``ip4.snet``
       - ``eq``
-      - :rspan:`1` ``$ADDR/$MASK``
-      - :rspan:`5` ``$ADDR`` is an IPv4 network address in dotted-decimal format, \"ddd.ddd.ddd.ddd\", where ddd is a decimal number of up to three digits in the range 0 to 255, ``$MASK`` is a subnet mask in the range 0 to 32.
-    * - ``not``
+      - ``$ADDR/$MASK``
+      - :rspan:`3` ``$ADDR`` is an IPv4 network address in dotted-decimal format, \"ddd.ddd.ddd.ddd\", where ddd is a decimal number of up to three digits in the range 0 to 255, ``$MASK`` is a subnet mask in the range 0 to 32.
     * - ``in``
       - ``{$ADDR/$MASK[,...]}``
-    * - :rspan:`2` Destination network
-      - :rspan:`2` ``ip4.dnet``
+    * - :rspan:`1` Destination network
+      - :rspan:`1` ``ip4.dnet``
       - ``eq``
-      - :rspan:`1` ``$ADDR/$MASK``
-    * - ``not``
+      - ``$ADDR/$MASK``
     * - ``in``
       - ``{$ADDR/$MASK[,...]}``
-    * - :rspan:`1` Protocol
-      - :rspan:`1` ``ip4.proto``
+    * - Protocol
+      - ``ip4.proto``
       - ``eq``
-      - :rspan:`1` ``$PROTOCOL``
-      - :rspan:`1` ``$PROTOCOL`` must be a transport layer protocol name (e.g. "ICMP", case insensitive), or a valid decimal or hexadecimal `internet protocol number`_.
-    * - ``not``
-    * - :rspan:`1` DSCP
-      - :rspan:`1` ``ip4.dscp``
+      - ``$PROTOCOL``
+      - ``$PROTOCOL`` must be a transport layer protocol name (e.g. "ICMP", case insensitive), or a valid decimal or hexadecimal `internet protocol number`_.
+    * - DSCP
+      - ``ip4.dscp``
       - ``eq``
-      - :rspan:`1` ``$DSCP``
-      - :rspan:`1` ``$DSCP`` is a `DSCP class name`_ (e.g. "ef", "cs1", "af21", case insensitive) or a numeric value (0-63, decimal or hexadecimal). "be" is accepted as an alias for "cs0".
-    * - ``not``
+      - ``$DSCP``
+      - ``$DSCP`` is a `DSCP class name`_ (e.g. "ef", "cs1", "af21", case insensitive) or a numeric value (0-63, decimal or hexadecimal). "be" is accepted as an alias for "cs0".
 
 
 IPv6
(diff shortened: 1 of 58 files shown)


src/libbpfilter/include/bpfilter/rule.h
```diff
@@ -137,11 +137,12 @@ void bf_rule_dump(const struct bf_rule *rule, prefix_t *prefix);
  *        type . Can be NULL but only if @p payload_len is 0, in which case
  *        there is no payload.
  * @param payload_len Length of the payload.
+ * @param negate If true, negate the matcher's comparison.
  * @return 0 on success, or negative errno value on failure.
  */
 int bf_rule_add_matcher(struct bf_rule *rule, enum bf_matcher_type type,
                         enum bf_matcher_op op, const void *payload,
-                        size_t payload_len);
+                        size_t payload_len, bool negate);
```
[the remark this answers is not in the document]
jordalgo: @qdeslandes Do we have to worry about public API breakages like this? Do we know of other code linking against a packaged version of lbpfilter?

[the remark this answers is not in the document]
yaakov-stein: I'm pretty sure we don't need to worry about this at the moment. Given the recent breaking changes with daemonless and given we're < release 1.0, I think Quentin mentioned this is fine as of now (but will let him confirm).

src/libbpfilter/include/bpfilter/matcher.h
```diff
@@ -329,6 +330,8 @@ enum bf_matcher_op bf_matcher_get_op(const struct bf_matcher *matcher);
 const void *bf_matcher_payload(const struct bf_matcher *matcher);
 size_t bf_matcher_payload_len(const struct bf_matcher *matcher);
 size_t bf_matcher_len(const struct bf_matcher *matcher);
+void bf_matcher_set_negate(struct bf_matcher *matcher, bool negate);
```
yaakov-stein: Why do we need `negate` to be mutable? Can't we just pass it in `bf_matcher_new` like `op` and the rest of the parameters? Do we need this method?

src/libbpfilter/cgen/matcher/cmp.c
```diff
@@ -17,6 +17,33 @@
...
+    case BF_MATCHER_ALL:
+        continue_on_equal = true;
+        break;
+    case BF_MATCHER_NE:
+    case BF_MATCHER_ANY:
+    case BF_MATCHER_IN:
+        continue_on_equal = false;
+        break;
+    case BF_MATCHER_RANGE:
+    case _BF_MATCHER_OP_MAX:
```
yaakov-stein: Don't want to get involved in the heated argument here... but for the sake of humans and AI agents not getting confused by a one-off usage, can we switch this to a default like the rest of the codebase?

src/libbpfilter/cgen/matcher/cmp.h
```diff
@@ -5,12 +5,29 @@
...
 struct bf_program;
 
+/**
+ * @brief Get the BPF jump opcode for a matcher, accounting for negation.
+ *
+ * Returns the BPF opcode to use in a jump-to-next-rule instruction. The
+ * returned opcode encodes the "skip if not matching" semantics, with the
+ * matcher's @ref bf_matcher_op and negation flag already folded in.
+ *
+ * Not valid for @ref BF_MATCHER_RANGE, which requires two separate jump
```
yaakov-stein: nit: backticks instead of `@ref`. Also below on L27 and above on L21

src/libbpfilter/cgen/matcher/cmp.h
```diff
@@ -5,12 +5,29 @@
...
+ * instructions with different opcodes.
+ *
+ * @param matcher Matcher to query. Can't be NULL.
+ * @return BPF jump opcode (e.g. @c BPF_JNE or @c BPF_JEQ).
+ */
+uint8_t bf_cmp_get_jmp_ins(const struct bf_matcher *matcher);
```
yaakov-stein, line 26: Hm, I find it slightly strange that we're putting this in `cmp.h/c` given that it isn't really doing any comparison, but I'm not quite sure if there is a better place.

src/libbpfilter/cgen/matcher/cmp.c
```diff
@@ -70,15 +97,12 @@ static void _bf_prefix_to_mask(unsigned int prefixlen, uint8_t *mask,
...
 
-int bf_cmp_value(struct bf_program *program, enum bf_matcher_op op,
-                 const void *ref, unsigned int size, int reg)
+int bf_cmp_value(struct bf_program *program, uint8_t jmp_op, const void *ref,
+                 unsigned int size, int reg)
 {
     assert(program);
     assert(ref);
 
-    if (op != BF_MATCHER_EQ && op != BF_MATCHER_NE)
```
yaakov-stein, line 79: Let me know if I'm missing something here, but it seems like we're going from callee verifying that only specific op's work -> every caller needing to verify (and that is why we first need to call `bf_cmp_get_jmp_ins` every time in the caller). Can't we keep it how it was, especially given `bf_cmp_value` should only be EQ/NE, `bf_cmp_range` should be RANGE etc. (i.e. it makes sense to check internally), and just pass the matcher and call `bf_cmp_get_jmp_ins` in the callee? This would allow us to not need to add the checks at every callsite.

This also, imo, keeps the public methods a bit simpler (as of now, we sometimes add `negate`, sometimes `jmp_op`, and sometimes `op`, while we can just pass matcher and handle details internally).

Thoughts?

src/libbpfilter/cgen/matcher/cmp.c
```diff
@@ -17,6 +17,33 @@
...
+        continue_on_equal = true;
+        break;
+    case BF_MATCHER_NE:
+    case BF_MATCHER_ANY:
+    case BF_MATCHER_IN:
+        continue_on_equal = false;
+        break;
+    case BF_MATCHER_RANGE:
+    case _BF_MATCHER_OP_MAX:
+        bf_abort("invalid matcher op to get jmp instruction %d",
```
yaakov-stein, line 38: Wondering - is there a reason we are aborting instead of propagating the error? Seems like we use `bf_abort` pretty sparingly as of now.

src/libbpfilter/cgen/matcher/cmp.h
```diff
@@ -36,14 +53,14 @@ int bf_cmp_value(struct bf_program *program, enum bf_matcher_op op,
  * size 4 and `BPF_REG_3`/`BPF_REG_4` for size 16.
  *
  * @param program Program to emit into. Can't be NULL.
- * @param op `BF_MATCHER_EQ` or `BF_MATCHER_NE`.
+ * @param jmp_op BPF jump opcode from @ref bf_cmp_get_jmp_ins.
```
yaakov-stein: nit: backticks (will stop commenting on this to avoid comment bloat, but would be great if you can just check the rest of the doxygen comments)

src/libbpfilter/cgen/matcher/meta.c
```diff
@@ -45,13 +51,25 @@ static int
...
-        program,
-        BPF_JMP_IMM(BPF_JGT, BPF_REG_0,
-                    (uint32_t)((double)UINT32_MAX * (proba / 100.0)), 0));
+
+    if (bf_matcher_get_negate(matcher)) {
+        _clean_bf_jmpctx_ struct bf_jmpctx j0 = bf_jmpctx_default();
+
+        j0 = bf_jmpctx_get(program,
+                           BPF_JMP_IMM(BPF_JGT, BPF_REG_0, threshold, 0));
+        EMIT_FIXUP_JMP_NEXT_RULE(program, BPF_JMP_A(0));
```
yaakov-stein: Doesn't `EMIT_FIXUP_JMP_NEXT_RULE(program, BPF_JMP_IMM(BPF_JLE, BPF_REG_0, threshold, 0));` work here instead of using the `jmpctx`?

src/libbpfilter/cgen/matcher/meta.c
```diff
@@ -140,10 +168,16 @@ _bf_matcher_generate_meta_flow_probability(struct bf_program *program,
...
-    EMIT_FIXUP_JMP_NEXT_RULE(
-        program,
-        BPF_JMP32_IMM(BPF_JGT, BPF_REG_0,
-                      (uint32_t)((double)UINT32_MAX * (proba / 100.0)), 0));
+    if (bf_matcher_get_negate(matcher)) {
+        _clean_bf_jmpctx_ struct bf_jmpctx j0 = bf_jmpctx_default();
+
+        j0 = bf_jmpctx_get(program,
+                           BPF_JMP32_IMM(BPF_JGT, BPF_REG_0, threshold, 0));
+        EMIT_FIXUP_JMP_NEXT_RULE(program, BPF_JMP_A(0));
```
yaakov-stein: Same as above (regarding using `BPF_JLE` instead)

src/libbpfilter/cgen/matcher/set.c
```diff
@@ -84,6 +85,10 @@ int bf_matcher_generate_set(struct bf_program *program,
     assert(program);
     assert(matcher);
 
+    if (bf_matcher_get_op(matcher) != BF_MATCHER_IN)
```
yaakov-stein: nit: braces around multiple lines after if statement

doc/usage/bfcli.rst
```diff
@@ -506,9 +506,10 @@ Matchers are defined such as:
 
 .. code:: shell
 
-    $TYPE [$OP] $PAYLOAD
+    $TYPE [!] [$OP] $PAYLOAD
 
 With:
+  - ``!``: optional comparison negation. When used, it prefixes an operator, e.g. ``tcp.sport ! eq 80``, ``(ip4.saddr) ! in foo``, or, with no operator, ``tcp.sport ! 80`` (equivalent to ``tcp.sport not 80``).
```
yaakov-stein: Is it just me that finds the `meta.probability ! eq 33.3` confusing? Are we sure we even want to support that? If so, should we maybe add some documentation that explain what it does?

yaakov-stein (dismissed):
Cool how this unlocks/makes it easier to do a ton of stuff! Have [one main comment ](https://github.com/facebook/bpfilter/pull/499/changes/47ee4568332ec79a54f4dd14a77105c493b487b0#r3083420120) that I'd be interested to hear your thoughts on, rest are largely nits or more minor questions

src/libbpfilter/cgen/matcher/cmp.c
jordalgo: This is a good call out and it was kinda bothering me too. I'm just going to pass in the entire matcher as suggested and then we can keep the checks as they were.

src/libbpfilter/cgen/matcher/cmp.c
jordalgo: > Don't want to get involved in the heated argument here

A heated argument over switch statements? Sign me up! Seriously, I don't feel too strongly about this one but I prefer the exhaustive check as you can bake in (not sure if we have it now) compile time warnings/errors if a new enum is added and not supported in this switch. 

Also just because something is repeated in a code base doesn't make it right 😉

src/libbpfilter/cgen/matcher/cmp.c
jordalgo: Because this is a developer error and not a user error. It should be noisy and explode - kinda like an assert.

src/libbpfilter/cgen/matcher/cmp.h
jordalgo: Now this is something we can have a script to both enforce and, potentially, fix.

src/libbpfilter/cgen/matcher/cmp.h
jordalgo: I couldn't find a better place either.

src/libbpfilter/cgen/matcher/set.c
jordalgo: Allow me to get on my soap box for a moment: this is one of the reasons why omitting braces for ANY if statement is LAME - aside from the whole it's really easy to shoot yourself in the foot.

doc/usage/bfcli.rst
jordalgo: What's confusing about just negating meta.probability eq? I can't imagine it will be used much but why restrict that power to users?

src/libbpfilter/include/bpfilter/matcher.h
jordalgo: > Why do we need negate to be mutable?

No it doesn't need to be mutable. I just didn't want to add yet another required param to `bf_matcher_new` functions - if this was c++ I wouldn't care and we could have a default value.

src/libbpfilter/cgen/matcher/cmp.c
yaakov-stein: 😆 I'll let Quentin take it from here as I don't really have an opinion on this.

> Also just because something is repeated in a code base doesn't make it right 😉

Yes, but mixing the two can get confusing. If we do switch to this style/convention, we probably should do it in a different PR and do it for all instances in the codebase so we don't have clashing styles/conventions.

src/libbpfilter/cgen/matcher/cmp.h
yaakov-stein: Good callout - I'll add an issue for this (and in general to do more to enforce the style doc)

src/libbpfilter/cgen/matcher/set.c
yaakov-stein: You'll have to now argue with both Quentin ([doc](https://bpfilter.io/developers/style.html#formatting)) AND Linus ([doc](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces)). Sounds fun :)

doc/usage/bfcli.rst
yaakov-stein: Ya, I guess it's more in general that `meta.probability eq X` is just slightly different than the rest of the matchers. For example, the rest of the `eq` test equality, as one would imagine. In `meta.probability`, the `eq` actually tests if a random number is _less than_ the provided number (while negating will check if it is greater than). But I think your changes don't add any confusion that isn't already there, so will resolve.

src/libbpfilter/cgen/matcher/cmp.h
yaakov-stein: #510

src/libbpfilter/cgen/matcher/meta.c
jordalgo: Good call - will fix.

src/libbpfilter/cgen/matcher/meta.c
jordalgo: Same as above - will fix.

src/libbpfilter/cgen/matcher/meta.c
```diff
@@ -155,12 +178,14 @@ int bf_matcher_generate_meta(struct bf_program *program,
     case BF_MATCHER_META_IFACE:
         return _bf_matcher_generate_meta_iface(program, matcher);
     case BF_MATCHER_META_L3_PROTO: {
-        uint16_t be_val = htobe16(*(uint16_t *)bf_matcher_payload(matcher));
-        return bf_cmp_value(program, BF_MATCHER_EQ, &be_val, 2, BPF_REG_7);
+        uint16_t be_val;
```
yaakov-stein: nit: Any reason we are seperating declaration and initialization here?

yaakov-stein (commented):
LGTM, sending over to @qdeslandes

doc/usage/bfcli.rst
qdeslandes: > Is it just me that finds the meta.probability ! eq 33.3 confusing?

I think `meta.x ! eq XXX` reads naturally. For `meta.probability`, the rationale was to try to prevent introducing a new keyword only for this use case: `meta.probability eq 60%` should be read as "if the probability is within 60%". Even having a dedicated keyword such as `less than 60%` it could be confusing. For this matcher specifically, `meta.probability 60%` is the same and is clearer.

> ``tcp.sport ! 80`` (equivalent to ``tcp.sport not 80``)

Do we still need the `not` keyword?

src/bfcli/print.c
```diff
@@ -215,10 +217,12 @@ void bfc_chain_dump(struct bf_chain *chain, struct bf_hookopts *hookopts,
                         (void)fprintf(stdout, ", ");
                 }
 
+                const char *in_str = negate ? "! in" : "in";
+
                 if (set->name) {
-                    (void)fprintf(stdout, ") in %s", set->name);
+                    (void)fprintf(stdout, ") %s %s", in_str, set->name);
```
qdeslandes, line 221: `(void)fprintf(stdout, ") %sin %s", negate ? "! " : "", set->name);`

src/bfcli/print.c
```diff
@@ -246,6 +250,8 @@ void bfc_chain_dump(struct bf_chain *chain, struct bf_hookopts *hookopts,
                 (void)fprintf(
                     stdout, "        %s",
                     bf_matcher_type_to_str(bf_matcher_get_type(matcher)));
+                if (negate)
+                    (void)fprintf(stdout, " !");
```
qdeslandes: Same as https://github.com/facebook/bpfilter/pull/499/changes#r3123293400.

src/libbpfilter/cgen/matcher/cmp.c
qdeslandes: > Also just because something is repeated in a code base doesn't make it right 😉

Many people being wrong don't make them right, is that it? ;) 

> can we switch this to a default like the rest of the codebase?

I would agree here, at least for consistency's sake. Jordan's argument makes perfect sense, but it should be applied codebase-wide to fix it once and for all.

src/libbpfilter/cgen/matcher/meta.c
```diff
@@ -90,20 +106,24 @@ static int _bf_matcher_generate_meta_port(struct bf_program *program,
     // If r1 == 0: no TCP nor UDP header found, jump to the next rule
     EMIT_FIXUP_JMP_NEXT_RULE(program, BPF_JMP_IMM(BPF_JEQ, BPF_REG_1, 0, 0));
 
-    if (bf_matcher_get_op(matcher) == BF_MATCHER_RANGE) {
+    if (op == BF_MATCHER_RANGE) {
```
qdeslandes: `bf_matcher_get_op(matcher) == BF_MATCHER_RANGE` as `op` is only used here.

src/libbpfilter/cgen/packet.c
```diff
@@ -273,9 +283,13 @@ static int _bf_matcher_pkt_generate_ip6_dscp(struct bf_program *program,
                                              const struct bf_matcher *matcher,
                                              const struct bf_matcher_meta *meta)
 {
+    enum bf_matcher_op op = bf_matcher_get_op(matcher);
     uint8_t dscp;
     int r;
 
+    if (op != BF_MATCHER_EQ && op != BF_MATCHER_NE)
+        return bf_err_r(-EINVAL, "unsupported operator %d", op);
```
qdeslandes: This kind of checks should already be handled during chain validation.

src/libbpfilter/cgen/packet.c
```diff
@@ -247,9 +255,13 @@ static int _bf_matcher_pkt_generate_ip4_dscp(struct bf_program *program,
...
```
qdeslandes: This kind of checks should already be handled during chain validation. E.g. https://github.com/facebook/bpfilter/blob/main/src/libbpfilter/chain.c#L167C8-L172C10

src/libbpfilter/include/bpfilter/matcher.h
qdeslandes: I don't mind long function definition with many parameters (up to a point). The main reason for the `bf_matcher_new()` (and constructors) function to have that many matchers is to avoid creating a incomplete matcher (e.g. type set but no op, or no payload). Once `bf_matcher_new()` returns successfully, the matcher is valid.

Regarding `negate`, I don't have any strong opinion. It has a valid default value, although it's essentially part of the operator, so it would make sense to have it as an argument to `bf_matcher_new()`.

src/libbpfilter/cgen/packet.c
```diff
@@ -206,7 +207,14 @@ _bf_matcher_pkt_generate_ip6_nexthdr(struct bf_program *program,
                                      const struct bf_matcher *matcher)
 {
     const uint8_t ehdr = *(uint8_t *)bf_matcher_payload(matcher);
+    enum bf_matcher_op op = bf_matcher_get_op(matcher);
     uint8_t eh_mask;
+    uint8_t jmp_op;
+
+    if (op != BF_MATCHER_EQ && op != BF_MATCHER_NE)
+        return bf_err_r(-EINVAL, "unsupported operator %d", op);
```
qdeslandes: This kind of checks should already be handled during chain validation.

src/libbpfilter/include/bpfilter/rule.h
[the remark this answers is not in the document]
qdeslandes: Yep, we can break for now :)

qdeslandes (changes requested):

src/bfcli/print.c
jordalgo: Both branches check negate so I added this line above which includes `in`:
```
const char *in_str = negate ? "not in" : "in";
```

src/libbpfilter/cgen/matcher/meta.c
jordalgo: What - you don't like as many stack variables as humanly possible. I mean, c'mon LLVM probably just optimizes away the `op` variable anyway. Trust your compiler! Ok, I'm done, I'll change it.

src/libbpfilter/cgen/matcher/cmp.c
jordalgo: Fine fine. Updated to use `default`

src/libbpfilter/include/bpfilter/matcher.h
jordalgo: Ok, updated `bf_matcher_new`

jordalgo:
Updated (after offline conversation) to replace the NE matcher with a "not" operator that can be used on all other operators.

src/bfcli/print.c
qdeslandes: Then variables should be defined at the beginning of the scope :p This ain't C++ here sir!

default.profraw
qdeslandes, line 1: This doesn't need to be upstreamed ;)

qdeslandes (changes requested):
Few nits, otherwise LGTM!

default.profraw
jordalgo: oops. I should add it to gitignore

qdeslandes (approved):
Alright, thanks for doing this, it's a really elegant implementation!

qdeslandes (approved):

jordalgo:
Thanks @yaakov-stein and @qdeslandes for the thorough review!

PR #500, opened 2026-04-07
tests: check: use find_program for iwyu_tool

jordalgo:
Use `find_program` instead of the absolute path for iwyu. This fixes an issue where this binary can't be found in a nix shell.

Tested locally and now the iwyu tests pass and the binary can be located.

### DIFF
--- a/tests/check/CMakeLists.txt
+++ b/tests/check/CMakeLists.txt
@@ -27,13 +27,6 @@ add_test(NAME "check.lint"
             ${bf_srcs}
 )
 
-add_test(NAME "check.iwyu"
-    COMMAND
-        /usr/bin/iwyu_tool.py
-            -p ${CMAKE_BINARY_DIR}
-            ${CMAKE_SOURCE_DIR}/src/libbpfilter/cgen/program.c
-)
-
 set_tests_properties("check.lint" PROPERTIES
     LABELS "check"
 )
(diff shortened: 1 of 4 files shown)


tests/check/CMakeLists.txt
```diff
@@ -27,17 +28,25 @@ add_test(NAME "check.lint"
...
+                ${CMAKE_SOURCE_DIR}/src/libbpfilter/cgen/program.c
+    )
+
+    set_tests_properties("check.iwyu" PROPERTIES
+        LABELS "check"
+    )
+else()
+    message(WARNING "iwyu_tool not found, check.iwyu test will be skipped")
+endif()
+
```
qdeslandes: I would remove the `if()`: if iwyu is not found, it should error out. Otherwise, we could miss the warning and proceed without it.

qdeslandes (changes requested):
Quick nit, otherwise LGTM!

tests/check/CMakeLists.txt
jordalgo: I had it without the `if` in the first iteration but I think it was causing a build failure in one or two of the CI jobs

jordalgo: Let me try again

qdeslandes:
Closing this PR for inactivity, it can be re-opened if needed.

PR #503, opened 2026-04-10
doc: add bfcli matcher hook compatibility table

### LINKED ISSUE #481: Update bfcli Documentation with Hook <> Matcher Compatibility
The bfcli documentation lists matchers but doesn't comprehensively indicate which hooks each matcher supports. This information is already tracked via the `unsupported_hooks` field in `struct bf_matcher_meta` ([src/libbpfilter/matcher.c](https://github.com/facebook/bpfilter/blob/main/src/libbpfilter/matcher.c)).

We should refactor the docs so this information can be presented clearly.

LiuHuaize:
## Summary
- add a generated matcher-to-hook compatibility section to `doc/usage/bfcli.rst`
- derive the compatibility matrix from `src/libbpfilter/matcher.c` so the docs follow `unsupported_hooks`
- wire the generator into `doc/CMakeLists.txt` and commit the generated include

Fixes #481

## Testing
- python3 -m py_compile doc/generate_matcher_hook_compat.py
- python3 doc/generate_matcher_hook_compat.py --matcher-c src/libbpfilter/matcher.c --matcher-h src/libbpfilter/include/bpfilter/matcher.h --hook-h src/libbpfilter/include/bpfilter/hook.h --output doc/usage/_generated/bfcli_matcher_hook_compatibility.rst --check

## Notes
- Full Sphinx/Doxygen doc build was not run locally because this environment does not have `cmake`, `ninja`, `sphinx-build`, or `doxygen` installed.


yaakov-stein (changes requested):
Hi, thanks for working on this and welcome to the repo! This will be a really helpful addition! I have a few high-level comments/questions before getting into the details:

1) I'm thinking that the proper approach here is to add the code that generates the file and then generate the file as a part of `make doc` as opposed to having the generated code live in the repo. This is similar to how it's already done for other generated code in the repo and should simplify things (don't need to check the diff at CI time).

2) As opposed to writing a parser and manually parsing the relevant source files (seems prone to error), we should be able to use `bf_matcher_get_meta` (public) and we can rely on the exact same code that is used directly in the repo. This should make the generation simpler and less fragile.

Let me know what you think of those points!

qdeslandes (changes requested):
Sending the PR back to the contributor for them to notice @yaakov-stein's comment.

qdeslandes:
@LiuHuaize it's been 2 weeks, do you intent to update this PR? :)

qdeslandes:
Closing this PR for inactivity, it can be re-opened if needed.

PR #506, opened 2026-04-14
build: add iwyu to docker files

jordalgo:
This will be needed when we switch to requiring iwyu as a Cmake dependency.


jordalgo:
I need to land this before being able to do this: https://github.com/facebook/bpfilter/pull/500

qdeslandes (changes requested):
Could you add it to other Dockerfiles too? We often use the Dockerfile as source of truth for the dependencies required on a specific system.

jordalgo:
Done.

qdeslandes (changes requested):
On Ubuntu, the package is `iwyu`.

jordalgo:
> On Ubuntu, the package is iwyu.

I gotta stop trusting Gemini.

PR #517, opened 2026-04-17
[1.5/2] Add Bare `log` Option

yaakov-stein:
Stack:
- #485
- #486
- -> #517 (didn't want to renumber all of them)
- #491

In order to add logging support for `CGROUP_SOCK_ADDR` hooks, we'd like to be able to log all of the data we have without needing to pass parameters to log. To do that, I was able to come up with three approaches:

1) Have this be purely on the parsing side and no "backend" changes. When the parser sees `log`, it just translates it to `log link,internet,transport`. However, this doesn't work too well for the new `CGROUP_SOCK_ADDR` logging, as it recognizes it as `log link,internet,transport`, so we'd need to accept that syntax as well, which isn't ideal.
2) Add some sentinel bit value in `rule->log` that signals "we want to log everything". This would work, but is a bit hacky and doesn't really encode the intent of what we're doing. It'd come along with a nice size comment that explains that the sentinel value means `ALL` but I could see this easily being a source of bugs in the future.
3) Add some way to distinguish between log options and the type of logging you want (none, all, or some). This requires a bit more work, but by pulling apart the options from logging "mode", we give ourselves alot more flexibility both now and in the future (we can add new modes that use the options for something else etc.).

Option 3 is a bit more code but seemed like the right choice due to the reasons included above.

### Summary
(Bulk of the changes are renaming in the tests.) Adds `enum bf_rule_log_mode` to `bf_rule`, separating the user's logging intent (`NONE`, `DEFAULT`, `PKT_HEADERS`) from the packet header bitmask (renamed `log` to `log_opts`). It also adds bare `log` keyword support in the parser and lexer.

### Tests
- Unit + e2e tests
- Manual Testing

### DIFF
--- a/src/bfcli/lexer.l
+++ b/src/bfcli/lexer.l
@@ -21,7 +21,6 @@
 %option nounput
 
 %s STATE_HOOK_OPTS
-%s STATE_LOG_OPTS
 %s STATE_MARK_OPTS
 %s STATE_REDIRECT_IFACE
 %s STATE_REDIRECT_DIR
@@ -96,14 +95,7 @@ mark            { BEGIN(STATE_MARK_OPTS); return MARK; }
 }
 
     /* Logs */
-log             { BEGIN(STATE_LOG_OPTS); return LOG; }
-<STATE_LOG_OPTS>{
-    [0-9a-zA-Z]+(,[0-9a-zA-Z]+)* {
-        BEGIN(INITIAL);
-        yylval.sval = strdup(yytext);
-        return LOG_HEADERS;
-    }
-}
+log             { return LOG; }
 
     /* Sets */
 \([ \t\n\r\f\v]*([a-zA-Z0-9_]+\.[a-zA-Z0-9_]+)([ \t\n\r\f\v]*,[ \t\n\r\f\v]*([a-zA-Z0-9_]+\.[a-zA-Z0-9_]+))*[ \t\n\r\f\v]*\) {
--- a/tests/harness/Rule.hpp
+++ b/tests/harness/Rule.hpp
@@ -5,7 +5,6 @@
 
 #pragma once
 
-#include <bitset>
 #include <memory>
 #include <stdexcept>
 #include <vector>
@@ -21,8 +20,6 @@ extern "C" {
 namespace bf
 {
 
-using RuleLogBitset = std::bitset<_BF_LOG_OPT_MAX>;
-
 class Rule
 {
 private:
@@ -36,11 +33,11 @@ private:
 
     bf_verdict _verdict;
     bool _counters;
-    RuleLogBitset _log;
+    uint8_t _log;
     std::vector<Matcher> _matchers;
 
 public:
-    Rule(bf_verdict verdict, bool counters = false, RuleLogBitset log = {},
+    Rule(bf_verdict verdict, bool counters = false, uint8_t log = 0,
          std::vector<Matcher> matchers = {}):
         _verdict {verdict},
         _counters {counters},
@@ -63,7 +60,7 @@ public:
         if (r != 0)
             throw std::runtime_error("failed to create bf_rule");
 
-        rule->log = static_cast<uint8_t>(_log.to_ulong());
+        rule->log = _log;
         rule->counters = _counters;
         rule->verdict = _verdict;
 
--- a/tests/harness/Rule.hpp
+++ b/tests/harness/Rule.hpp
@@ -21,7 +21,7 @@ extern "C" {
 namespace bf
 {
 
-using RuleLogBitset = std::bitset<_BF_PKTHDR_MAX>;
+using RuleLogBitset = std::bitset<_BF_LOG_OPT_MAX>;
 
 class Rule
 {
--- a/doc/usage/bfcli.rst
+++ b/doc/usage/bfcli.rst
@@ -409,14 +409,17 @@ Rules are defined such as:
     rule
         [$MATCHER...]
         [$SET...]
-        [log link,internet,transport]
+        [log [$HEADERS]]
         [counter]
         [mark $MARK]
         $VERDICT
 
 With:
   - ``$MATCHER``: zero or more matchers. Matchers are defined later.
-  - ``log``: optional. If set, log the requested protocol headers. ``link`` will log the link (layer 2) header, ``internet`` with log the internet (layer 3) header, and ``transport`` will log the transport (layer 4) header. At least one header type is required. ``log`` is **not** supported by ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks.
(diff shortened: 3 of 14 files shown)


qdeslandes:
@claude review

qdeslandes (changes requested):
I understand the reasoning here, but I'm not convinced with the approach picked.

> Have this be purely on the parsing side and no "backend" changes. When the parser sees log, it just translates it to log link,internet,transport. However, this doesn't work too well for the new CGROUP_SOCK_ADDR logging, as it recognizes it as log link,internet,transport, so we'd need to accept that syntax as well, which isn't ideal.

Agreed.

> Add some sentinel bit value in rule->log that signals "we want to log everything". This would work, but is a bit hacky and doesn't really encode the intent of what we're doing. It'd come along with a nice size comment that explains that the sentinel value means ALL but I could see this easily being a source of bugs in the future.

This has the advantage of not creating a new enum type, which complexify the code and might cause discrepancies we need to gate for (if `log_opt` is set, but `log_mode` is 0). 

I think there is room for a middle ground, something like:

```
enum bf_log_opt {
    BF_LOG_NONE,
    BF_LOG_PKT_LINK,
    BF_LOG_PKT_INTERNET,
    BF_LOG_PKT_TRANSPORT,
    BF_LOG_DEFAULT = ~0u,
    _BF_PKTHDR_MAX,
};
```

yaakov-stein:
> This has the advantage of not creating a new enum type, which complexify the code and might cause discrepancies we need to gate for (if `log_opt` is set, but `log_mode` is 0).
> 
> I think there is room for a middle ground, something like:
> 
> ```
> enum bf_log_opt {
>     BF_LOG_NONE,
>     BF_LOG_PKT_LINK,
>     BF_LOG_PKT_INTERNET,
>     BF_LOG_PKT_TRANSPORT,
>     BF_LOG_DEFAULT = ~0u,
>     _BF_PKTHDR_MAX,
> };
> ```
Yep, that makes sense and is a good middle ground. We're slightly conflating the idea of what type of logging the user wants (none, headers, default), where the values are used directly, and the headers bitfields, which are used in `BF_FLAG(BF_LOG_PKT_LINK)`, which can be a bit confusing (atleast for me), but I think that is fine for now. I'll make the change

yaakov-stein:
@claude review

yaakov-stein:
@claude review

src/bfcli/lexer.l
```diff
@@ -98,11 +98,18 @@ mark            { BEGIN(STATE_MARK_OPTS); return MARK; }
     /* Logs */
 log             { BEGIN(STATE_LOG_OPTS); return LOG; }
 <STATE_LOG_OPTS>{
+    [ \t\n]     ;
+    "#".*       ;
+    (counter|mark|rule|chain|set|ACCEPT|DROP|CONTINUE|NEXT|REDIRECT) {
```
qdeslandes: Is there no way to do it without listing all the keywords that can follow a log instruction? E.g.:
```
log {return LOG;}
(link|transport|internet) {return LOG_OPT;}
```
Then have the parser deal with the fact options are optional:
```
%union {
    // [...]
    enum bf_log_option options;
    // [...]
}

rule_option     : LOG
                {
                    $$ = (struct bf_rule_options){
                        .log = BF_LOG_DEFAULT,
                        .flags = BF_RULE_OPTION_LOG,
                    };
                }
                // vvv this already exists
                | LOG LOG_HEADERS
                {
                    _cleanup_free_ char *in = $2;
                    char *tmp = in;
                    char *saveptr;
                    char *token;
                    uint8_t log = 0;

                    while ((token = strtok_r(tmp, ",", &saveptr))) {
                        enum bf_pkthdr header;

                        if (bf_pkthdr_from_str(token, &header) < 0)
                            bf_parse_err("unknown packet header '%s'", token);

                        log |= BF_FLAG(header);

                        tmp = NULL;
                    }

                    $$ = (struct bf_rule_options){
                        .log = log,
                        .flags = BF_RULE_OPTION_LOG,
                    };
               }
```

qdeslandes: Alright, I should have checked the next file.

src/libbpfilter/cgen/packet.c
```diff
@@ -362,13 +362,18 @@ int bf_packet_gen_inline_matcher(struct bf_program *program,
...
     assert(rule);
 
+    log = rule->log == BF_LOG_OPT_DEFAULT ? BF_FLAGS_MASK(_BF_LOG_OPT_MAX) :
+                                            rule->log;
+
     EMIT(program, BPF_MOV64_REG(BPF_REG_1, BPF_REG_10));
     EMIT(program, BPF_ALU64_IMM(BPF_ADD, BPF_REG_1, BF_PROG_CTX_OFF(arg)));
     EMIT(program, BPF_MOV64_IMM(BPF_REG_2, rule->index));
-    EMIT(program, BPF_MOV64_IMM(BPF_REG_3, rule->log));
+    EMIT(program, BPF_MOV64_IMM(BPF_REG_3, log));
```
qdeslandes: Why not keeping `BF_LOG_OPT_DEFAULT` value? It's `0xff` and covers all the options already.

tests/harness/Rule.hpp
```diff
@@ -36,11 +33,11 @@ class Rule
 
     bf_verdict _verdict;
     bool _counters;
-    RuleLogBitset _log;
+    uint8_t _log;
```
qdeslandes, line 36: Why not a `bitset` anymore?

yaakov-stein: Since we now have `BF_LOG_OPT_DEFAULT` which isn't a bitmask and can't be represented with the previous bitset. The previous `std::bitset<_BF_LOG_OPT_MAX>` is only 3 bits wide and wouldn't hold `BF_LOG_OPT_DEFAULT`. We can theoretically do `std::bitset<8>` and then have a representation for `BF_LOG_OPT_DEFAULT` but that is pretty much just a more complicated version `uint_8`. Can switch it over to `std::bitset<8>` if you think that is better.

src/bfcli/lexer.l
yaakov-stein: Agreed on the callout. I spent some time trying to find a cleaner way to do this and I think I was able to come up with something that works nicely and outputs clear errors. 

I've now removed the `STATE_LOG_OPTS` from the lexer and we only tokenize `LOG`. The log options then get tokenized as `STRING ',' STRING ...` and the actual parsing logic goes in `parser.y`. There, we check `bf_log_opt_from_str` as the source of truth instead needing any hardcoding (as it was before). If someone inputs an erroneous string, they get a clear message. The only side effect is we now allow spaces between the log options, but IMO that's more of a feature than a problem. Let me know what you think!

yaakov-stein:
@claude review

qdeslandes (approved):
LGTM, thanks!

PR #525, opened 2026-05-03
cli: add --no-set-content flag to chain/ruleset get

qdeslandes:
Printing a chain or ruleset that contains large sets produces output too long to be useful: a set with thousands of elements drowns out the rules around it. Add a `--no-set-content` flag to "bfcli chain get" and "bfcli ruleset get" that elides set elements and prints only the element count, while keeping the set name and key visible so the set remains identifiable.

Without `--no-set-content`:
```
$ bfcli chain get --name test_xdp
chain test_xdp BF_HOOK_XDP{ifindex=2} ACCEPT
    counters policy 274 packets 31542 bytes; error 0 packets 0 bytes
    set blocked_ips (ip4.saddr) in {
        10.0.0.1
        10.0.0.2
        10.0.0.3
    }
    rule
        (ip4.saddr) in blocked_ips
        DROP
    rule
        (ip4.saddr) in {
            192.168.1.1
            192.168.1.2
        }
        DROP
```

With `--no-set-content`:
```
$ bfcli chain get --name test_xdp --no-set-content chain test_xdp BF_HOOK_XDP{ifindex=2} ACCEPT
    counters policy 417 packets 47594 bytes; error 0 packets 0 bytes
    set blocked_ips (ip4.saddr) in { /* 3 elements, content elided */ }
    rule
        (ip4.saddr) in blocked_ips
        DROP
    rule
        (ip4.saddr) in { /* 2 elements, content elided */ }
        DROP
```

### DIFF
--- a/src/bfcli/print.c
+++ b/src/bfcli/print.c
@@ -34,6 +34,8 @@
 #include <bpfilter/set.h>
 #include <bpfilter/verdict.h>
 
+#include "bpfilter/core/hashset.h"
+
 struct bfc_chain_opts;
 
 #define INET6_ADDRSTRLEN 46
@@ -81,7 +83,7 @@ static void bf_dump_hex_local(const void *data, size_t len)
 }
 
 void bfc_chain_dump(struct bf_chain *chain, struct bf_hookopts *hookopts,
-                    bf_list *counters)
+                    bf_list *counters, bool no_set_content)
 {
     struct bf_counter *counter;
     bf_list_node *counter_node, *policy_counter_node, *err_counter_node;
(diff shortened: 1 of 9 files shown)


src/bfcli/print.c
```diff
@@ -169,6 +171,14 @@ void bfc_chain_dump(struct bf_chain *chain, struct bf_hookopts *hookopts,
             if (i != set->n_comps - 1)
                 (void)fprintf(stdout, ", ");
         }
+
+        if (no_set_content) {
+            (void)fprintf(stdout,
+                          ") in { /* %zu elements, content elided */ }\n",
+                          bf_hashset_size(&set->elems));
```
yaakov-stein, line 178: Wonder if we should add a `bf_set_size` so we don't need to leak internals of `bf_set`

src/bfcli/ruleset.c
```diff
@@ -51,13 +51,11 @@ int bfc_ruleset_get(const struct bfc_opts *opts)
     _clean_bf_list_ bf_list counters = bf_list_default(bf_list_free, NULL);
     int r;
 
-    (void)opts;
```
yaakov-stein, line 54: nit: `assert` instead?

yaakov-stein (dismissed):
LGTM, two minor comments (hashset one probably can be addressed separately)

src/bfcli/print.c
qdeslandes: Added to my list of bootcamp/beginner tasks :)

PR #526, opened 2026-05-05
Refactor ruleset discovery and locking

qdeslandes:
`bf_ctx_setup()` currently initializes a global context with a file descriptor to the bpfilter pin directory, locks it (exclusive lock), and restore all the existing chains.

This behaviour is sub-optimal, as it prevents any concurrent `bfcli` call (or library call) to read the ruleset or modify it. This PR aims to improve this behaviour by:
- Restoring only the necessary BPF objects: `bf_ctx_setup()` won't perform any discovery. Each `bf_chain*` call will restore the require chain, and `bf_ruleset*` calls will restore the ruleset when needed.
- Improve locking granularity by allowing the ruleset and/or specific chains to be locked in read or write mode.

Full details are available in the commits and documentation.


qdeslandes:
@claude review

qdeslandes:
@claude review

PR #527, opened 2026-05-08
Embed counter snapshots in `bf_rule` and `bf_chain`

qdeslandes:
Counter values were previously delivered alongside chains as a parallel `bf_list` parameter on `bf_chain_get` / `bf_ruleset_get`, requiring every consumer to walk two lists in lockstep and to re-derive which slot belonged to which rule. This series moves the snapshots into the data model and refreshes them on the read path.


qdeslandes:
@claude review

qdeslandes:
@claude review

PR #532, opened 2026-05-12
build: fix building benchmark

pzmarzly:
`benchmark_bin` currently cannot be built:

```
$ ninja -C build benchmark_bin
ninja: Entering directory `build'
[1/2] Building CXX object tools/benchmarks/CMakeFiles/benchmark_bin.dir/main.cpp.o
FAILED: [code=1] tools/benchmarks/CMakeFiles/benchmark_bin.dir/main.cpp.o
/usr/bin/c++  -I/root/repos/bpfilter-merge-bpf-maps/tools/benchmarks -I/root/repos/bpfilter-merge-bpf-maps/tests/harness -I/root/repos/bpfilter-merge-bpf-maps/src/libbpfilter/include -I/root/repos/bpfilter-merge-bpf-maps/build/src/libbpfilter/include -g -std=c++20 -DWITH_GZFILEOP -MD -MT tools/benchmarks/CMakeFiles/benchmark_bin.dir/main.cpp.o -MF tools/benchmarks/CMakeFiles/benchmark_bin.dir/main.cpp.o.d -o tools/benchmarks/CMakeFiles/benchmark_bin.dir/main.cpp.o -c /root/repos/bpfilter-merge-bpf-maps/tools/benchmarks/main.cpp
/root/repos/bpfilter-merge-bpf-maps/tools/benchmarks/main.cpp: In function ‘void {anonymous}::single_rule__ip4_saddr(benchmark::State&)’:
/root/repos/bpfilter-merge-bpf-maps/tools/benchmarks/main.cpp:105:20: error: no matching function for call to ‘bf::Rule::Rule(bf_verdict, bool, <brace-enclosed initializer list>, std::vector<bf::Matcher>)’
  105 |                   });
      |                    ^
  • there are 3 candidates
In file included from /root/repos/bpfilter-merge-bpf-maps/tests/harness/Chain.hpp:17,
                 from /root/repos/bpfilter-merge-bpf-maps/tools/benchmarks/main.cpp:31:
    • candidate 1: ‘bf::Rule::Rule(bf_verdict, std::optional<bf_counter>, uint8_t, std::vector<bf::Matcher>)’
      /root/repos/bpfilter-merge-bpf-maps/tests/harness/Rule.hpp:42:5:
         42 |     Rule(bf_verdict verdict,
            |     ^~~~
      • no known conversion for argument 2 from ‘bool’ to ‘std::optional<bf_counter>’
[...]

PR #535, opened 2026-05-12
tools: add benchmark for loading sets

pzmarzly:
Measure time it takes to load a chain that contains a large set. This will be a common operation (especially since sets can update).

Requires https://github.com/facebook/bpfilter/pull/532.

```
$ rm -rf build && cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=release -DNO_DOCS=1 -DNO_CHECKS=1
$ ninja -C build benchmark_bin bfcli
$ sudo build/output/sbin/benchmark_bin \
    --cli build/output/sbin/bfcli \
    --srcdir . \
    --outfile /tmp/load.json \
    --filter chain_load

...
Run on (4 X 2396.4 MHz CPU s)
...
chain_load__ip4_saddr__x_elem_set/8/manual_time           7.20 ms         7.08 ms          106 load chain, ip4.saddr, 8 elements set
chain_load__ip4_saddr__x_elem_set/128/manual_time         7.69 ms         7.61 ms          107 load chain, ip4.saddr, 128 elements set
chain_load__ip4_saddr__x_elem_set/32768/manual_time       55.3 ms         61.5 ms           12 load chain, ip4.saddr, 32768 elements set
```

### DIFF
--- a/tools/benchmarks/main.cpp
+++ b/tools/benchmarks/main.cpp
@@ -207,6 +207,50 @@ BENCHMARK(single_rule__ip4_saddr__x_elem_set)
     ->Arg(1 << 7)
     ->Arg(1 << 15);
 
+void chain_set__ip4_saddr__x_elem_set(::benchmark::State &state)
+{
+    const std::string chain_name = "bf_benchmark";
+    Chain chain(chain_name, BF_HOOK_XDP, BF_VERDICT_ACCEPT);
+
+    Set s = Set({BF_MATCHER_IP4_SADDR});
+    uint32_t nelems = state.range(0);
+    for (uint32_t i = 0; i < nelems; ++i)
+        s << uint32ToIp4(i);
+    chain << s;
+
+    chain << Rule(BF_VERDICT_DROP, std::nullopt, 0,
+                  std::vector<Matcher> {
+                      Matcher(BF_MATCHER_SET, BF_MATCHER_IN, {0, 0, 0, 0}),
+                  });
+
+    auto chainp = chain.get();
+
+    for (auto _: state) {
+        int ret = bf_chain_set(chainp.get(), nullptr);
+        if (ret < 0) {
+            state.SkipWithError("failed to load chain");
+            break;
+        }
+
+        state.PauseTiming();
+        ret = bf_chain_flush(chain_name.c_str());
+        if (ret < 0) {
+            state.SkipWithError("failed to flush chain");
+            break;
+        }
+        state.ResumeTiming();
+    }
+
+    state.SetLabel(
+        std::format("load chain, ip4.saddr, {} elements set", nelems));
+}
+
+BENCHMARK(chain_set__ip4_saddr__x_elem_set)
+    ->Arg(1 << 3)
+    ->Arg(1 << 16)
+    ->Arg(1 << 20)
+    ->Unit(::benchmark::kMillisecond);
+
 void single_rule__ip4_saddr_c(::benchmark::State &state)
 {
     Chain chain("bf_benchmark", BF_HOOK_XDP, BF_VERDICT_ACCEPT);


pzmarzly:
@claude review

tools/benchmarks/main.cpp
```diff
@@ -207,10 +208,60 @@ BENCHMARK(single_rule__ip4_saddr__x_elem_set)
     ->Arg(1 << 7)
     ->Arg(1 << 15);
 
+void chain_load__ip4_saddr__x_elem_set(::benchmark::State &state)
+{
+    const std::string chainName = "bf_benchmark";
+    uint32_t nelems = state.range(0);
+
+    for (auto _: state) {
```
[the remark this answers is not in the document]
pzmarzly: > manual timing of a heavy kernel operation that cannot be batched

They probably can be batched though. The change from `while` to `for` was done for me by local Claude Code and I didn't catch it, but now it is a good moment to ask which one is more correct. Quentin what do you think? I'm leaning towards regular KeepRunningBatch, but I don't understand GBenchmark well.

tools/benchmarks/main.cpp
```diff
@@ -207,6 +208,56 @@ BENCHMARK(single_rule__ip4_saddr__x_elem_set)
...
+        for (uint32_t i = 0; i < nelems; ++i)
+            s << uint32ToIp4(i);
+
+        chain << Rule(BF_VERDICT_DROP, std::nullopt, 0,
+                      std::vector<Matcher> {
+                          Matcher(BF_MATCHER_SET, BF_MATCHER_IN, {0, 0, 0, 0}),
+                      });
+        chain << s;
+
+        auto chainp = chain.get();
```
qdeslandes: This can be done outside of the benchmarking loop.

tools/benchmarks/main.cpp
```diff
@@ -207,6 +208,56 @@ BENCHMARK(single_rule__ip4_saddr__x_elem_set)
...
+        }
+
+        state.SetIterationTime(
+            std::chrono::duration<double>(end - start).count());
+
+        ret = bf_chain_flush(chainName.c_str());
+        if (ret < 0) {
+            state.SkipWithError("failed to flush chain");
+            break;
+        }
```
qdeslandes: Use 
```
{
    benchmark::ScopedPauseTiming pause(state); // Pauses timing
    ret = bf_chain_flush(chainName.c_str());
    if (ret < 0) {
        state.SkipWithError("failed to flush chain");
        break;
    }
}
```

So you don't have to manually set the iteration time.

tools/benchmarks/main.cpp
```diff
@@ -207,6 +208,56 @@ BENCHMARK(single_rule__ip4_saddr__x_elem_set)
     ->Arg(1 << 7)
     ->Arg(1 << 15);
 
+void chain_load__ip4_saddr__x_elem_set(::benchmark::State &state)
```
qdeslandes: `chain_set__ip4_saddr__x_elem_set`

tools/benchmarks/main.cpp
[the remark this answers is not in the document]
qdeslandes: `for (auto _: state)` is correct, so let's keep it. This lets Google Benchmark define the number of iterations to run the benchmark for.

We don't use it to benchmark BPF programs, because `BPF_PROG_TEST_RUN` also accept an `iterations` argument, so we use it instead and use `KeepRunningBatch()` to check if we need more iterations. That ensures benchmarks are run with a warm cache, which is more reflective of real-world use cases.

tools/benchmarks/main.cpp
```diff
@@ -207,6 +208,56 @@ BENCHMARK(single_rule__ip4_saddr__x_elem_set)
...
+    }
+
+    state.SetLabel(
+        std::format("load chain, ip4.saddr, {} elements set", nelems));
+}
+
+BENCHMARK(chain_load__ip4_saddr__x_elem_set)
+    ->Arg(1 << 3)
+    ->Arg(1 << 7)
+    ->Arg(1 << 15)
```
qdeslandes: Let's use different values to have a clear overview of the perf here:
- 1 << 3 -> 8
- 1 << 16 -> 65k
- 1 << 22 -> ~4M

qdeslandes (changes requested):

tools/benchmarks/main.cpp
pzmarzly: ScopedPauseTiming was only merged a month ago, and is not available yet in gbenchmark shipped by my distro. Fun! https://github.com/google/benchmark/pull/2157

I'll use PauseTiming and ResumeTiming functions.

tools/benchmarks/main.cpp
pzmarzly: Good idea, but I'll make the last be 1<<20, so it doesn't take too long.

```
-----------------------------------------------------------------------------------
Benchmark                                         Time             CPU   Iterations
-----------------------------------------------------------------------------------
chain_set__ip4_saddr__x_elem_set/8             6.90 ms         6.70 ms          121 load chain, ip4.saddr, 8 elements set
chain_set__ip4_saddr__x_elem_set/65536          104 ms          103 ms            7 load chain, ip4.saddr, 65536 elements set
chain_set__ip4_saddr__x_elem_set/1048576       1793 ms         1780 ms            1 load chain, ip4.saddr, 1048576 elements set
```

pzmarzly:
@claude review

qdeslandes (approved):
LGTM! Thanks!

PR #544, opened 2026-05-15
core: make vector foreach standard C compliant

### LINKED ISSUE #509: Make `bf_vector_foreach` compliant with the C standard
`bf_vector_foreach` performs pointer arithmetic on `void * ((elem) + (vec)->elem_size and (vec)->data + ...)`, which is a GCC extension not valid in standard C.

Since `vector.h` is a public header under `include/bpfilter/core/`, any downstream consumer using this macro with `-pedantic -std=c17` will get a compile error. The pedantic integration test (`tests/integration/pedantic_c.sh`) includes all public headers with `-pedantic-errors -std=c17 -Werror` — the macro definition alone won't trigger an error (unexpanded macros aren't compiled), but any call site under those flags will fail.

MukundaKatta:
## Summary
- avoid GNU void-pointer arithmetic in bf_vector_foreach
- use char-pointer arithmetic internally so the public macro compiles under standard C

Fixes #509.

## Testing
- git diff --check
- cc -std=c17 -pedantic-errors -Werror -I src/libbpfilter/include -c /tmp/bpfilter-vector-XXXXXX.c

Note: full CMake configuration is blocked locally because libbpf is not installed.

### DIFF
--- a/src/libbpfilter/include/bpfilter/core/vector.h
+++ b/src/libbpfilter/include/bpfilter/core/vector.h
@@ -62,9 +62,11 @@ void bf_vector_init(bf_vector *vec, size_t elem_size);
 #define bf_vector_foreach(vec, elem)                                           \
     for (void *(elem) = (vec)->data,                                           \
               *__end = (vec)->data ?                                           \
-                           (vec)->data + ((vec)->size * (vec)->elem_size) :    \
+                           (char *)(vec)->data +                               \
+                               ((vec)->size * (vec)->elem_size) :              \
                            NULL;                                               \
-         (elem) && (elem) < __end; (elem) = (elem) + (vec)->elem_size)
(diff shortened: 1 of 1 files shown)


qdeslandes:
> cc -std=c17 -pedantic-errors -Werror -I src/libbpfilter/include -c /tmp/bpfilter-vector-XXXXXX.c

That's not how we build the project.

> Note: full CMake configuration is blocked locally because libbpf is not installed.

"I made a change to prevent a build error but didn't build the project" is not an acceptable testing plan.

It's the second time you're opening a PR on this project, and it's the second time you disregard the style guide (#519). It's either a lack of human involvement or a lack of basic decency when it comes to testing the code you submit. Next time I suspect another AI-generated comment or PR from you, I'll block you.

I spent more time writing this comment than you spent opening 2 PRs on this project.

PR #547, opened 2026-05-23
lib: cgen: replace counters map with per-CPU array

qdeslandes:
Switch BF_MAP_TYPE_COUNTERS from BPF_MAP_TYPE_ARRAY to
BPF_MAP_TYPE_PERCPU_ARRAY. Each CPU updates its own counter slot
without contention, which eliminates the need for atomic operations
in the BPF fast path (bpf_map_lookup_elem on a per-CPU map returns
a direct pointer to the current CPU's value).

## Changes

**lib:**
- Add `BF_BPF_MAP_TYPE_PERCPU_ARRAY` to `bf_bpf_map_type`
- Switch the counters map to `BF_BPF_MAP_TYPE_PERCPU_ARRAY`
- `bf_handle_get_counter()`: allocate a per-CPU buffer, do one lookup,
  then sum all CPU slots into the returned `bf_counter`
- Add `bf_handle_set_counter()`: builds a per-CPU buffer with the value
  in CPU 0's slot and zeroes elsewhere, then calls
  `bf_bpf_map_update_elem()` with the full buffer
- `_bf_cgen_transfer_counters()`: replace `bf_map_set_elem()` with
  `bf_handle_set_counter()` — the old call passed a 16-byte buffer to
  a syscall that reads `num_cpus × 16` bytes, writing garbage from the
  stack into the new map

**tests:**
- Update `bpftool map dump` jq queries in 8 e2e tests: per-CPU dumps
  use `values` (array per CPU) instead of `value`, so counter reads
  change from `.[N].value.count` to
  `.[N].values | map(.value.count) | add`

### DIFF
--- a/src/libbpfilter/cgen/cgen.c
+++ b/src/libbpfilter/cgen/cgen.c
@@ -455,7 +455,7 @@ static int _bf_cgen_transfer_counters(const struct bf_handle *old_handle,
         if (!counter.count && !counter.size)
             continue;
 
-        r = bf_map_set_elem(new_handle->cmap, &i, &counter);
+        r = bf_handle_set_counter(new_handle, i, &counter);
         if (r)
             return bf_err_r(r, "failed to write counter %u", i);
     }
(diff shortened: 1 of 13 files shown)


qdeslandes:
@claude review

PR #553, opened 2026-05-27
lib: cgen: add per-rule state map for log rate limiting

qdeslandes:
Rules with a log action currently emit one entry per matching packet with no throttling. To support rate-limited logging, BPF programs need mutable per-rule state that persists across invocations.

Introduce bf_smap, a single-entry BPF_MAP_TYPE_ARRAY map per chain whose value is a flat array of bf_rule_state entries (one per rule). The map is looked up once in the program prologue and the base pointer stored in bf_runtime.state_map; per-rule access is then a constant offset, avoiding one bpf_map_lookup_elem call per rule at packet time.

For each rule with log_rate_ns set, the generated code reads last_log_ts, calls bpf_ktime_get_ns(), and skips the log elfstub if the elapsed time falls short of the configured interval. R9 (callee-saved) holds the state entry pointer across the ktime call. Writes to last_log_ts are best-effort: no CAS, but a naturally aligned 8-byte store is atomic on x86-64, so no torn writes occur.

Expose the rate in the DSL as 'log [<headers>] every <N>{ns,us,ms,s}' and wire it to the new bf_rule.log_rate_ns field serialized alongside the existing rule fields.

### DIFF
--- a/doc/usage/bfcli.rst
+++ b/doc/usage/bfcli.rst
@@ -425,7 +425,7 @@ Rules are defined such as:
     rule
         [$MATCHER...]
         [$SET...]
-        [log [$HEADERS]]
+        [log [$HEADERS] [every $FREQUENCY]]
         [counter]
         [mark $MARK]
         $VERDICT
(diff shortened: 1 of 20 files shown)


qdeslandes:
@claude review

Issue #562, opened 2026-07-16
Port range in sets


qdeslandes:
Allow users to specify a port range in sets:

```
[...]
rule
    (tcp.dport) in { 11, 22, 33-44 }
    [...]
```

This rule would be converted to a set containing `{ 11, 22, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43, 44 }`. 

That being said, what should be printed by `bfcli ruleset get` when printing such a set? Should we try to factorize the elements to print the range (keeping in mind the sets are not sorted)?

PR #570, opened 2026-07-18
lib: set: support port ranges in set elements

ChatchawanIllyes:
## Summary

Implements port ranges in set elements (#562).

`bf_set_add_elem_raw()` now recognises range syntax (`START-END`) in a set
element's components and expands it into individual set elements. For example:

```
(tcp.dport) in { 11; 22; 33-44 }
```

produces a set containing `{ 11, 22, 33, 34, …, 44 }` (14 elements).

Range detection reuses the `BF_MATCHER_RANGE` parsing that already backs the
standalone `range` matcher operator — a component is treated as a range only
when the token contains a `-` and the component's matcher type registers
`BF_MATCHER_RANGE` ops. This scopes the feature to the matcher types that
already support ranges (ports today) without hardcoding anything port-specific,
and non-range / multi-component parsing is unchanged.

## Behaviour

- Each value in `[start, end]` is expanded into its own element and inserted
  into the set's hashset, deduplicated like any other element.
- Expanded values are re-parsed through the existing single-value
  (`BF_MATCHER_IN`) path, so their in-memory representation is byte-identical to
  an element written as a plain value.
- Range bounds are validated by the existing `_bf_parse_l4_port_range()`:
  both bounds must be valid decimal ports `<= 65535`, and `start <= end`.
  `44-33`, `33-`, and `-44` are rejected.

## Scope / limitations

A couple of things are intentionally left out to keep this a minimal, focused
change — flagging them for maintainer input:

- **Printing is unchanged.** Expanded ranges are printed element-by-element,
  not re-factored back into a `START-END` range. This is the open question
  raised in #562; happy to add range factorization on `ruleset get` in a
  follow-up if that's the preferred direction.
- **One ranged component per element.** A set element with more than one ranged
  component (e.g. `(tcp.sport, tcp.dport) in { 1-2, 3-4 }`) returns `-EINVAL`
  rather than performing a Cartesian expansion. This avoids surprising blow-up
  in element count; can revisit if Cartesian expansion is wanted.

## A note on element separators

The issue's example is written as `{ 11, 22, 33-44 }`, but bpfilter separates
set **elements** with `;` and **components within an element** with `,`
(e.g. `{ 80; 443 }` vs `(ip4.daddr, tcp.sport) in { 1.2.3.4, 80 }`). So the
range form is `{ 11; 22; 33-44 }`. This PR does not change the grammar, so the
comma-separated form from the issue's example still parses as multiple
components. If treating `,` as an element separator for single-component sets
is desired, that's a separate grammar change and I'm glad to take it on
separately.

## Tests

Added unit tests in `tests/unit/libbpfilter/set.c`:

- single-component range expansion (`{11; 22; 33-44}` → 14 elements)
- range boundaries (`0-1`, degenerate `65535-65535`)
- invalid ranges (`44-33`, `33-`, `-44`)
- a range in one component of a two-component key (`{1.2.3.4, 80-82}` → 3)
- ranges in both components of an element (rejected)

All `unit.libbpfilter.set` tests pass (28/28).

### DIFF
--- a/src/libbpfilter/set.c
+++ b/src/libbpfilter/set.c
@@ -6,7 +6,9 @@
 #include "bpfilter/set.h"
 
 #include <errno.h>
+#include <inttypes.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
@@ -161,9 +163,15 @@ int bf_set_add_elem_raw(struct bf_set *set, const char *raw_elem)
 {
     _cleanup_free_ void *elem = NULL;
     _cleanup_free_ char *_raw_elem = NULL;
+    const struct bf_matcher_ops *range_in_ops = NULL;
     char *tmp, *saveptr, *token;
     size_t elem_offset = 0;
     size_t comp_idx = 0;
+    size_t range_offset = 0;
+    enum bf_matcher_type range_type = 0;
+    uint16_t range_start = 0;
+    uint16_t range_end = 0;
+    bool has_range = false;
     int r;
 
     assert(set);
@@ -183,6 +191,7 @@ int bf_set_add_elem_raw(struct bf_set *set, const char *raw_elem)
     tmp = _raw_elem;
     while ((token = strtok_r(tmp, ",", &saveptr))) {
         const struct bf_matcher_ops *ops;
+        const struct bf_matcher_ops *range_ops;
 
         if (comp_idx >= set->n_comps) {
             return bf_err_r(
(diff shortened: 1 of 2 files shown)


qdeslandes:
Hi, could you please disclose if/how much AI has been used for this work? Thank you.

ChatchawanIllyes:
> Hi, could you please disclose if/how much AI has been used for this work? Thank you.

Hey there, only used AI for two things: drafting the unit tests in tests/unit/libbpfilter/set.c, and writing up the PR description. 

I reviewed both, and ran the test suite locally (unit.libbpfilter.set, 28/28) to confirm the cases and checked the tests actually cover what they say they do.

qdeslandes:
> I reviewed both, and ran the test suite locally (unit.libbpfilter.set, 28/28) to confirm the cases and checked the tests actually cover what they say they do.

Why didn't you run all the tests?

ChatchawanIllyes:
I didn't. I ran `ctest -R unit.libbpfilter.set` and left it there. The change only touches `bf_set_add_elem_raw()`, so I figured the set tests were the ones at risk, which meant `check.checkstyle` never ran. That's what's failing.

One thing I hadn't realized until I went looking: my container was built on Ubuntu with clang-format 18, so even a full local run might have passed while CI still failed. I've rebuilt from `.github/fedora-44.Dockerfile` so I'm checking against the same version you are.

Branch is rebased on main now (single commit, no merge), formatting fixed, and I ran everything this time. 123/123: 23 unit, 93 e2e, 1 fuzzing, 2 integration, 4 check. CI on the new head needs approval when you get a chance.

qdeslandes (changes requested):
I have a very strong feeling that I'm talking with an AI bot here, not a human. 

That being said:
- Why running a single test suite make `make test_bin test` will run  all the tests for you?
- One ranged component per element: no, we should allow as many ranges as the user wants to use.
- For now, ranges should not be allowed in multi-components keys
- Add details to your commit message

PR #578, opened 2026-08-11
lib: log: support packet 5-tuple logging

### LINKED ISSUE #568: Support 5-tuple logging
By default, a `log` action will log all the known header of a given packet, which can be large depending on the L2/3/4 headers in the packet. Potentially, users only want specific information, such as the 5-tuple (addresses, ports, and protocol), which would greatly reduce the amount of data sent back to userspace.

Implement 5-tuple logging with `log 5-tuple` and limit the amount of data logged to the ring buffer.

blackdragoon26:
## Summary

Add `log 5-tuple` as an exclusive packet logging mode.

The implementation introduces a dedicated ELF stub that emits compact records
containing source and destination addresses, source and destination ports, and
the transport protocol. It supports complete IPv4/IPv6 TCP/UDP tuples while
preserving the existing fixed-size `struct bf_log` reservation and ring-buffer
capacity.

Unsupported packets do not emit a tuple record or fall back to raw-header
logging. Their counters, marks, verdicts, and normal rule execution remain
unchanged.

## Related issue

Fixes #568

## Testing

Added coverage for:

- valid `log 5-tuple` syntax, with and without `every`
- rejection of combinations with packet-layer options
- rejection on `cgroup_sock_addr`
- option conversion and rule serialization
- IPv4 and IPv6 with TCP and UDP
- tuple addresses, ports, protocols, rule ID, and verdict
- no tuple record for ICMP and non-IP packets
- counters and verdicts when tuple logging is skipped
- unsupported packets not consuming the rate-limit interval
- unchanged raw-header logging behaviour

Debug build with sanitizers and coverage enabled:

```text
100% tests passed, 0 tests failed out of 122

Label Time Summary:
check          = 16.28 sec*proc (2 tests)
e2e            = 99.33 sec*proc (94 tests)
integration    =  0.14 sec*proc (2 tests)
unit           =  0.65 sec*proc (23 tests)
```

Release build:
```
100% tests passed, 0 tests failed out of 122

Label Time Summary:
check          = 15.02 sec*proc (2 tests)
e2e            = 57.04 sec*proc (94 tests)
integration    =  0.14 sec*proc (2 tests)
unit           =  0.22 sec*proc (23 tests)
```

Additional `passed` verification:
- `make -C build-issue568` fixstyle
- `make -C build-issue568 test_bin` test
- `make -C build-issue568 doc`
- debug ASan/UBSan execution 
- coverage generation — 84.4% lines and 95.5% functions
- final lint, IWYU, checkstyle, and diff whitespace checks

## Notes for the reviewer

`5-tuple` is mutually exclusive with `link`, `internet`, and `transport`.
Tuple records are emitted only for IPv4/IPv6 packets using TCP or UDP, following
the eligibility behaviour discussed in #568. The eligibility check happens
before rate-limit bookkeeping, so an unsupported packet does not consume the
rule's `every` interval.
`cgroup_sock_addr` is intentionally unsupported for this mode. Existing raw
packet logging and socket-address logging remain unchanged.
The new log type and ELF-stub identifiers are appended so existing identifier
values remain stable.

## AI disclosure
I used AI for codebase exploration, understanding the existing logging and BPF
code-generation paths, design iteration, parts of the implementation and tests,
verification, review, and wording of this PR.
I reviewed the resulting design and test evidence and remain responsible for
the submitted changes.

## Checklist

- [x] `make -C $BUILD test_bin test` passes and the code follows the style guide

### DIFF
--- a/doc/usage/bfcli.rst
+++ b/doc/usage/bfcli.rst
@@ -148,7 +148,8 @@ bfcli will print log entries as they are published by the chain. Hit ``Ctrl+C`` 
 
 Every log entry begins with a shared header: the receive timestamp, the matching rule's index, and the applied verdict. The remaining fields depend on the hook type:
 
-- For packet-based hooks, the header also includes the matched packet size. It is followed by each requested layer's protocol headers (see the ``log`` action below). If a requested layer could not be processed by the chain, the corresponding output will be truncated.
+- For packet-based hooks using packet-layer logging, the header also includes the matched packet size. It is followed by each requested layer's protocol headers (see the ``log`` action below). If a requested layer could not be processed by the chain, the corresponding output will be truncated.
+- For packet-based hooks using 5-tuple logging, the source and destination addresses and ports are printed on a single line with the transport protocol.
 - For ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks, the entry includes destination address, destination port, process ID, and process name. Sendmsg hooks additionally include the source address.
 
 **Options**
@@ -432,12 +433,13 @@ Rules are defined such as:
 
 With:
   - ``$MATCHER``: zero or more matchers. Matchers are defined later.
-  - ``log``: optional. Two forms are supported:
+  - ``log``: optional. Three forms are supported:
 
     - ``log $HEADERS``: log specific packet headers. ``$HEADERS`` is a comma-separated list of ``link`` (layer 2), ``internet`` (layer 3), and/or ``transport`` (layer 4). Only supported by packet-based hooks (XDP, TC, NF, cgroup_skb).
+    - ``log 5-tuple``: log source and destination addresses and ports, and the transport protocol. This mode is only supported by packet-based hooks and only emits entries for IPv4/IPv6 packets using TCP or UDP. It is mutually exclusive with the packet-layer options; unsupported packets are not logged.
     - ``log``: log all available data for the hook type. For packet-based hooks, this is equivalent to ``log link,internet,transport``. For ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks, this records the process ID, process name, destination address, and destination port. Sendmsg hooks additionally include the source address.
 
-    Either form accepts an optional ``every $FREQUENCY`` suffix to rate-limit log events. ``$FREQUENCY`` is a positive number (integer or decimal) followed by a unit: ``ns``, ``us``, ``ms``, or ``s`` (e.g. ``every 1s``, ``every 500ms``, ``every 1.5s``). At most one log entry is emitted per ``$FREQUENCY`` interval per rule. Without ``every``, every match is logged.
+    Each form accepts an optional ``every $FREQUENCY`` suffix to rate-limit log events. ``$FREQUENCY`` is a positive number (integer or decimal) followed by a unit: ``ns``, ``us``, ``ms``, or ``s`` (e.g. ``every 1s``, ``every 500ms``, ``every 1.5s``). At most one log entry is emitted per ``$FREQUENCY`` interval per rule. Without ``every``, every match is logged.
   - ``counter``: optional literal. If set, the filter will count the number of events matched by the rule. For packet-based hooks, this includes both the number of packets and the total bytes. For ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks, this counts the number of socket operations (``connect()`` or ``sendmsg()`` calls).
   - ``mark``: optional, ``$MARK`` must be a valid decimal or hexadecimal 32-bits value. If set, write the packet's marker value. This marker can be used later on in a rule (see ``meta.mark``) or with a TC filter.
   - ``$VERDICT``: action taken by the rule if the packet is matched against **all** the criteria: either ``ACCEPT``, ``DROP``, ``CONTINUE``, ``NEXT``, or ``REDIRECT``.
--- a/src/bfcli/lexer.l
+++ b/src/bfcli/lexer.l
@@ -290,7 +290,7 @@ tcp\.flags      { BEGIN(STATE_MATCHER_TCP_FLAGS); yylval.sval = strdup(yytext); 
     }
 }
 
-[a-zA-Z0-9_]+   { yylval.sval = strdup(yytext); return STRING; }
+[a-zA-Z0-9_-]+  { yylval.sval = strdup(yytext); return STRING; }
 
 .               { return *yytext; }
 
--- a/src/bfcli/print.c
+++ b/src/bfcli/print.c
@@ -677,6 +677,55 @@ static void _bf_chain_log_sock_addr(const struct bf_log *log)
                   bf_logger_get_color(BF_COLOR_RESET, BF_STYLE_RESET));
 }
 
+static void _bf_chain_log_5_tuple(const struct bf_log *log)
+{
+    char src_addr[INET6_ADDRSTRLEN];
+    char dst_addr[INET6_ADDRSTRLEN];
+    const char *protocol;
+    int family;
+
+    assert(log);
+
+    if (log->l3_proto == ETH_P_IP) {
+        family = AF_INET;
+    } else if (log->l3_proto == ETH_P_IPV6) {
+        family = AF_INET6;
+    } else {
+        (void)fprintf(stdout, "  5-tuple   : <unknown protocol 0x%04x>\n",
+                      log->l3_proto);
+        return;
+    }
+
+    inet_ntop(family, log->pkt_5_tuple.saddr, src_addr, sizeof(src_addr));
+    inet_ntop(family, log->pkt_5_tuple.daddr, dst_addr, sizeof(dst_addr));
+    protocol = bf_ipproto_to_str(log->l4_proto);
+    /* Tuple logging only emits TCP or UDP records, both of which are known to
+     * bf_ipproto_to_str(). */
+
+    (void)fprintf(stdout, "  5-tuple   : %s%s%s ",
+                  bf_logger_get_color(BF_COLOR_LIGHT_MAGENTA, BF_STYLE_BOLD),
+                  protocol ?: "unknown",
+                  bf_logger_get_color(BF_COLOR_RESET, BF_STYLE_RESET));
+
+    if (family == AF_INET6) {
+        (void)fprintf(stdout, "%s[%s]:%u%s → %s[%s]:%u%s\n",
+                      bf_logger_get_color(BF_COLOR_LIGHT_CYAN, BF_STYLE_BOLD),
+                      src_addr, log->pkt_5_tuple.sport,
+                      bf_logger_get_color(BF_COLOR_RESET, BF_STYLE_RESET),
+                      bf_logger_get_color(BF_COLOR_LIGHT_CYAN, BF_STYLE_BOLD),
+                      dst_addr, log->pkt_5_tuple.dport,
+                      bf_logger_get_color(BF_COLOR_RESET, BF_STYLE_RESET));
+    } else {
+        (void)fprintf(stdout, "%s%s:%u%s → %s%s:%u%s\n",
+                      bf_logger_get_color(BF_COLOR_CYAN, BF_STYLE_BOLD),
+                      src_addr, log->pkt_5_tuple.sport,
+                      bf_logger_get_color(BF_COLOR_RESET, BF_STYLE_RESET),
+                      bf_logger_get_color(BF_COLOR_CYAN, BF_STYLE_BOLD),
+                      dst_addr, log->pkt_5_tuple.dport,
+                      bf_logger_get_color(BF_COLOR_RESET, BF_STYLE_RESET));
+    }
+}
+
 void bfc_print_log(const struct bf_log *log)
 {
     assert(log);
@@ -694,6 +743,9 @@ void bfc_print_log(const struct bf_log *log)
         break;
     case BF_LOG_TYPE_SOCK_ADDR:
         _bf_chain_log_sock_addr(log);
+        break;
+    case BF_LOG_TYPE_PACKET_5_TUPLE:
+        _bf_chain_log_5_tuple(log);
         break;
     default:
         break;
--- a/src/libbpfilter/cgen/packet.c
+++ b/src/libbpfilter/cgen/packet.c
@@ -427,15 +427,29 @@ int bf_packet_gen_inline_log(struct bf_program *program,
     EMIT(program, BPF_MOV64_REG(BPF_REG_1, BPF_REG_10));
     EMIT(program, BPF_ALU64_IMM(BPF_ADD, BPF_REG_1, BF_PROG_CTX_OFF(arg)));
     EMIT(program, BPF_MOV64_IMM(BPF_REG_2, rule->index));
-    EMIT(program, BPF_MOV64_IMM(BPF_REG_3, rule->log));
-    EMIT(program, BPF_MOV64_IMM(BPF_REG_4, rule->verdict));
 
-    // Pack l3_proto and l4_proto
-    EMIT(program, BPF_MOV64_REG(BPF_REG_5, BPF_REG_7));
-    EMIT(program, BPF_ALU64_IMM(BPF_LSH, BPF_REG_5, 16));
-    EMIT(program, BPF_ALU64_REG(BPF_OR, BPF_REG_5, BPF_REG_8));
+    if (rule->log == BF_FLAG(BF_LOG_OPT_5_TUPLE)) {
+        EMIT(program, BPF_MOV64_IMM(BPF_REG_3, rule->verdict));
 
-    EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_PKT_LOG);
+        // Pack l3_proto and l4_proto
+        EMIT(program, BPF_MOV64_REG(BPF_REG_4, BPF_REG_7));
+        EMIT(program, BPF_ALU64_IMM(BPF_LSH, BPF_REG_4, 16));
+        EMIT(program, BPF_ALU64_REG(BPF_OR, BPF_REG_4, BPF_REG_8));
+
+        EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_PKT_5_TUPLE_LOG);
+    } else {
+        EMIT(program, BPF_MOV64_IMM(BPF_REG_3, rule->log == BF_LOG_OPT_DEFAULT ?
+                                                   BF_LOG_PACKET_HEADERS :
+                                                   rule->log));
+        EMIT(program, BPF_MOV64_IMM(BPF_REG_4, rule->verdict));
+
+        // Pack l3_proto and l4_proto
+        EMIT(program, BPF_MOV64_REG(BPF_REG_5, BPF_REG_7));
+        EMIT(program, BPF_ALU64_IMM(BPF_LSH, BPF_REG_5, 16));
+        EMIT(program, BPF_ALU64_REG(BPF_OR, BPF_REG_5, BPF_REG_8));
+
+        EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_PKT_LOG);
+    }
 
     return 0;
 }
--- a/src/libbpfilter/cgen/program.c
+++ b/src/libbpfilter/cgen/program.c
@@ -7,8 +7,11 @@
 
 #include <linux/bpf.h>
 #include <linux/bpf_common.h>
+#include <linux/if_ether.h>
+#include <linux/in.h> // NOLINT
 #include <linux/limits.h>
 
+#include <endian.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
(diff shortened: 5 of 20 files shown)


doc/usage/bfcli.rst
```diff
@@ -432,12 +433,13 @@ Rules are defined such as:
 
 With:
   - ``$MATCHER``: zero or more matchers. Matchers are defined later.
-  - ``log``: optional. Two forms are supported:
+  - ``log``: optional. Three forms are supported:
 
     - ``log $HEADERS``: log specific packet headers. ``$HEADERS`` is a comma-separated list of ``link`` (layer 2), ``internet`` (layer 3), and/or ``transport`` (layer 4). Only supported by packet-based hooks (XDP, TC, NF, cgroup_skb).
+    - ``log 5-tuple``: log source and destination addresses and ports, and the transport protocol. This mode is only supported by packet-based hooks and only emits entries for IPv4/IPv6 packets using TCP or UDP. It is mutually exclusive with the packet-layer options; unsupported packets do not fall back to packet-layer logging.
     - ``log``: log all available data for the hook type. For packet-based hooks, this is equivalent to ``log link,internet,transport``. For ``BF_HOOK_CGROUP_SOCK_ADDR_*`` hooks, this records the process ID, process name, destination address, and destination port. Sendmsg hooks additionally include the source address.
```
qdeslandes, line 440: > "unsupported packets do not fall back to packet-layer logging"

-> "unsupported packets are not logged."

src/bfcli/lexer.l
```diff
@@ -290,6 +290,7 @@ tcp\.flags      { BEGIN(STATE_MATCHER_TCP_FLAGS); yylval.sval = strdup(yytext);
     }
 }
 
+5-tuple         { yylval.sval = strdup(yytext); return STRING; }
```
qdeslandes: Don't create a specific token for the `5-tuple` keyword, update the definition of `STRING` below to support `-` in strings instead.

src/bfcli/print.c
```diff
@@ -677,6 +677,53 @@ static void _bf_chain_log_sock_addr(const struct bf_log *log)
...
+        family = AF_INET6;
+    } else {
+        (void)fprintf(stdout, "  5-tuple   : <unknown protocol 0x%04x>\n",
+                      log->l3_proto);
+        return;
+    }
+
+    inet_ntop(family, log->pkt_5_tuple.saddr, src_addr, sizeof(src_addr));
+    inet_ntop(family, log->pkt_5_tuple.daddr, dst_addr, sizeof(dst_addr));
+    protocol = bf_ipproto_to_str(log->l4_proto);
```
qdeslandes, line 701: Add a comment that this only works because `l4_proto` is either TCP or UDP, and those are defined in `bf_ipproto_to_str`.

tests/fuzz/keywords.dict
```diff
@@ -146,6 +147,7 @@
 "link,internet"
 "internet,transport"
 "link,internet,transport"
+"log 5-tuple"
```
qdeslandes: This can be removed.

src/libbpfilter/chain.c
```diff
@@ -170,6 +171,13 @@ static int _bf_chain_check_rule(struct bf_chain *chain, struct bf_rule *rule)
         rule->disabled = r;
     }
 
+    if (rule->log != BF_LOG_OPT_DEFAULT && (rule->log & tuple) &&
```
qdeslandes, line 174: `BF_LOG_DEFAULT` should represent `link,internet,transport` logs.

src/libbpfilter/chain.c
```diff
@@ -170,6 +171,13 @@ static int _bf_chain_check_rule(struct bf_chain *chain, struct bf_rule *rule)
...
+        rule->log != tuple) {
```
qdeslandes, line 175: Can be simplified to `(rule->log & BF_FLAG(BF_LOG_OPT_5_TUPLE)) != rule->log`.

tests/unit/libbpfilter/rule.c
```diff
@@ -49,6 +49,7 @@ static void pack_and_unpack(void **state)
 
     // Pack the source rule
     assert_non_null(source = bft_rule_dummy(8));
+    source->log = BF_FLAG(BF_LOG_OPT_5_TUPLE);
```
qdeslandes: `bft_rule_dummy()` sets `rule->log = BF_FLAGS(BF_LOG_OPT_INTERNET, BF_LOG_OPT_TRANSPORT);`. Overwriting `source->log` here removes a test to introduce another. As `source->log` serialization is already tested, no need for a new test specifically for `BF_LOG_OPT_5_TUPLE`.

tests/harness/Rule.hpp
```diff
@@ -55,6 +56,13 @@ class Rule
         return *this;
     }
 
+    Rule &logEvery(uint64_t rateNs)
+    {
+        _logRateNs = rateNs;
+
+        return *this;
+    }
```
qdeslandes: Follow the existing pattern: add `std::optional<uint64_t> logRateNs` to the constructor.

tests/e2e/actions/log_5_tuple.cpp
qdeslandes, line 1: This would deserve to be moved in `tests/e2e/actions/log_5_tuple.cpp`.

tests/e2e/rules/log_5_tuple.cpp
```diff
@@ -0,0 +1,245 @@
...
+namespace
+{
+
+constexpr uint8_t kTupleLog = BF_FLAG(BF_LOG_OPT_5_TUPLE);
+
+static std::vector<bf::Matcher> alwaysMatch()
+{
+    return {bf::Matcher(BF_MATCHER_META_PROBABILITY, BF_MATCHER_EQ,
+                        bft_float_payload(100.0f))};
+}
```
qdeslandes: Convert to a static method in `Matcher` class.

tests/e2e/rules/log_5_tuple.cpp
```diff
@@ -0,0 +1,245 @@
...
+    assert_int_equal(BF_LOG_TYPE_PACKET_5_TUPLE, log.log_type);
+    assert_int_equal(l3Proto, log.l3_proto);
+    assert_int_equal(l4Proto, log.l4_proto);
+    assert_int_equal(0, log.rule_id);
+    assert_int_equal(BF_VERDICT_DROP, log.verdict);
+    assertAddress(log.pkt_5_tuple.saddr, family, saddr);
+    assertAddress(log.pkt_5_tuple.daddr, family, daddr);
+    assert_int_equal(sport, log.pkt_5_tuple.sport);
+    assert_int_equal(dport, log.pkt_5_tuple.dport);
+}
```
qdeslandes: Those are generic functions that should be part of the testing harness instead.

src/libbpfilter/cgen/program.c
```diff
@@ -567,55 +634,29 @@ static int _bf_program_generate_rule(struct bf_program *program,
...
-        // Rate-limited log: check last_log_ts in the state map before logging.
-        //
-        // R9 (callee-saved) holds the pointer to this rule's state entry
-        // across the bpf_ktime_get_ns() call.
-        EMIT(program, BPF_LDX_MEM(BPF_DW, BPF_REG_9, BPF_REG_10,
-                                  BF_PROG_CTX_OFF(state_map)));
+    if (rule->log == BF_FLAG(BF_LOG_OPT_5_TUPLE)) {
+        // A 5-tuple is only complete for IPv4/IPv6 packets using TCP/UDP.
+        // Skip only the log action for other packets, leaving the rule's
+        // remaining actions and verdict unchanged.
```
qdeslandes: Use `/* */` for multiline comments.

src/libbpfilter/cgen/program.c
```diff
@@ -504,6 +507,70 @@ static int _bf_program_check_proto(struct bf_program *program,
...
+    int r;
+
+    assert(program);
+    assert(rule);
+
+    if (rule->log && rule->log_rate_ns) {
+        // Rate-limited log: check last_log_ts in the state map before logging.
+        //
+        // R9 (callee-saved) holds the pointer to this rule's state entry
+        // across the bpf_ktime_get_ns() call.
```
qdeslandes: Use `/* */` for multiline comments.

src/libbpfilter/cgen/program.c
```diff
@@ -504,6 +507,70 @@ static int _bf_program_check_proto(struct bf_program *program,
...
+        EMIT(program, BPF_LDX_MEM(BPF_DW, BPF_REG_9, BPF_REG_10,
+                                  BF_PROG_CTX_OFF(state_map)));
+        {
+            // Outer skip: state_map is NULL (shouldn't happen at runtime,
+            // but the verifier requires the NULL check).
```
qdeslandes: Use `/* */` for multiline comments.

src/libbpfilter/cgen/program.c
```diff
@@ -567,55 +634,29 @@ static int _bf_program_generate_rule(struct bf_program *program,
...
                     return r;
             }
         }
-    } else if (rule->log) {
-        r = program->runtime.ops->gen_inline_log(program, rule);
+    } else {
+        r = _bf_program_generate_log(program, rule);
         if (r)
             return r;
     }
```
qdeslandes: This could be simplified:

```
if log:
    if log 5-tuple:
        setup 5-tuple logging
    bf_program_generate_log();
```

In `bf_program_generate_log`:
```
if log_rate_ns:
    rate-limit setup

gen_inline_log();
```

qdeslandes (changes requested):

blackdragoon26:
Thanks a lot for your detailed review. @qdeslandes 
I will be needing bit of time to replying and working on all of em, by the upcoming week.
Regards
Sankalp

doc/usage/bfcli.rst
blackdragoon26: alright, will simplify the wording

src/bfcli/lexer.l
blackdragoon26: OH Okay, will do that

src/bfcli/print.c
blackdragoon26: Gotcha, will add comment for improved documentation and understanding.

tests/fuzz/keywords.dict
blackdragoon26: Okayy

src/libbpfilter/chain.c
blackdragoon26: Will, correct this masking situation

src/libbpfilter/chain.c
blackdragoon26: wow Yeah, thats way better, i dont know why i overcomplicated it like that

tests/unit/libbpfilter/rule.c
blackdragoon26: Valid point, um I will remove that tuple-specific overwrite.

tests/harness/Rule.hpp
blackdragoon26: Got it!

tests/e2e/actions/log_5_tuple.cpp
blackdragoon26: Sure thing

tests/e2e/rules/log_5_tuple.cpp
blackdragoon26: alrightty

tests/e2e/rules/log_5_tuple.cpp
blackdragoon26: Hmmm makes sense will do that

src/libbpfilter/cgen/program.c
blackdragoon26: Understood

src/libbpfilter/cgen/program.c
blackdragoon26: Okay

src/libbpfilter/cgen/program.c
blackdragoon26: Cool

src/libbpfilter/cgen/program.c
blackdragoon26: Agreed. I will keep tuple eligibility setup separate and then call that common logging generator once.

blackdragoon26:
@qdeslandes , can you rerun this CI again.
Thanks

blackdragoon26:
@qdeslandes Could you review this pr.

yaakov-stein:
@claude review

yaakov-stein:
I'll go over claude's comments (to check their accuracy) and give this a review sometime in the next few days

blackdragoon26:
> I'll go over claude's comments (to check their accuracy) and give this a review sometime in the next few days

Alrighty!

src/libbpfilter/include/bpfilter/runtime.h
```diff
@@ -90,12 +90,20 @@ enum bf_log_opt
...
 
-    _BF_LOG_OPT_MAX,
+    /**
+     * Packet 5-tuple: source and destination addresses and ports, and the
+     * transport protocol.
+     */
+    BF_LOG_OPT_5_TUPLE,
 
-    /** Log all available data for the hook type. */
-    BF_LOG_OPT_DEFAULT = 0xFF,
```
yaakov-stein, line 96: At the moment, I don't believe we can remove this. While the other options (`BF_LOG_OPT_(LINK|INTERNET|TRANSPORT)`) are used for their specific bit (and thus `_BF_LOG_OPT_MAX` < 8 as it is a `uint8_t`), `BF_LOG_OPT_DEFAULT` is a sentinel meant to signify "log everything" in both the packet and `CGROUP_SOCK_ADDR` context (and anything else we may add in the future). This value was chosen because it is by definition both different than any other value (it's not a single bit, it's all the bits) and also will match every bit of the other values when used as a bit mask.

Therefore, if we remove this, we can no longer differentiate between `log link,internet,transport` and `log`, and thus 1) `log link,internet,transport` becomes valid syntax for `CGROUP_SOCK_ADDR`, which doesn't make much sense, and 2) not as much of a problem, but we can't print what the user gave as input for packet logging.

Check out https://github.com/facebook/bpfilter/pull/517 for a bit more background if you're interested.

src/bfcli/lexer.l
```diff
@@ -290,7 +290,7 @@ tcp\.flags      { BEGIN(STATE_MATCHER_TCP_FLAGS); yylval.sval = strdup(yytext);
     }
 }
 
-[a-zA-Z0-9_]+   { yylval.sval = strdup(yytext); return STRING; }
+[a-zA-Z0-9_-]+  { yylval.sval = strdup(yytext); return STRING; }
```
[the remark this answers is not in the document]
yaakov-stein: I think this is probably fine as is. The weird part is that `--` and `-` become valid set and chain names, but `__` and `_` already were.

src/libbpfilter/cgen/program.c
```diff
@@ -504,6 +507,63 @@ static int _bf_program_check_proto(struct bf_program *program,
     return 0;
 }
 
+static int _bf_program_generate_log(struct bf_program *program,
```
yaakov-stein, line 510: I think once we are making a `_bf_program_generate_log` method, let's put all of the logging logic there and can keep it simple doing something like this:
```
  static int _bf_program_generate_log(struct bf_program *program,
                                      const struct bf_rule *rule)
  {
      // Declare and assert

      if (!rule->log)
          return 0;

      if (rule->log == BF_FLAG(BF_LOG_OPT_5_TUPLE)) {
          ...
      }

      if (rule->log_rate_ns) {
          ...
      }

      return program->runtime.ops->gen_inline_log(program, rule);
  }
```

src/libbpfilter/chain.c
```diff
@@ -170,7 +171,13 @@ static int _bf_chain_check_rule(struct bf_chain *chain, struct bf_rule *rule)
         rule->disabled = r;
     }
 
-    if (rule->log && rule->log != BF_LOG_OPT_DEFAULT &&
+    if ((rule->log & tuple) && (rule->log & tuple) != rule->log) {
```
yaakov-stein: Per [this comment](https://github.com/facebook/bpfilter/pull/578/changes#r3867870017), we'll need to switch this to:
```
if (rule->log != BF_LOG_OPT_DEFAULT &&
      (rule->log & tuple) && rule->log != tuple) {
```

yaakov-stein (dismissed):
Overall looks pretty good, left a few comments and please squash the follow-up fix commit into the original commit so the change lands as one self-contained commit.

src/libbpfilter/include/bpfilter/runtime.h
blackdragoon26: Oh I see, will be checking in that pr, I really appreciate your detailed reasoning.
I missed that `BF_LOG_OPT_DEFAULT` is intentionally a semantic sentinel for bare `log`, rather than only a packet-header mask.
I will restore the sentinel, retain `BF_LOG_DEFAULT` for the explicit `link,internet,transport` mask, and preserve that distinction through validation, CLI printing, and packet codegen.

src/libbpfilter/cgen/program.c
blackdragoon26: Alrighty, I will move the no-log early return, 5-tuple eligibility guards, rate-limit handling, and final `gen_inline_log()` call into `_bf_program_generate_log()`.
so, `_bf_program_generate_rule()` will then call this helper once, while unsupported tuple packets will still continue through counters, marks, and verdict generation.

src/libbpfilter/chain.c
blackdragoon26: Oh Cool, I will use the suggested condition.

blackdragoon26:
Thanks for the review @yaakov-stein. I’ve addressed the requested changes.

src/libbpfilter/include/bpfilter/runtime.h
yaakov-stein: I think instead of keeping `BF_LOG_DEFAULT` (which is confusing with `BF_LOG_OPT_DEFAULT`), let's rename it to something like `BF_LOG_PACKET_HEADERS` or something like that.

blackdragoon26: Okayy, will do that.

src/libbpfilter/cgen/packet.c
```diff
@@ -421,21 +421,40 @@ int bf_packet_gen_inline_matcher(struct bf_program *program,
...
     EMIT(program, BPF_ALU64_IMM(BPF_ADD, BPF_REG_1, BF_PROG_CTX_OFF(arg)));
     EMIT(program, BPF_MOV64_IMM(BPF_REG_2, rule->index));
-    EMIT(program, BPF_MOV64_IMM(BPF_REG_3, rule->log));
-    EMIT(program, BPF_MOV64_IMM(BPF_REG_4, rule->verdict));
 
-    // Pack l3_proto and l4_proto
-    EMIT(program, BPF_MOV64_REG(BPF_REG_5, BPF_REG_7));
-    EMIT(program, BPF_ALU64_IMM(BPF_LSH, BPF_REG_5, 16));
-    EMIT(program, BPF_ALU64_REG(BPF_OR, BPF_REG_5, BPF_REG_8));
+    if (is_5_tuple) {
```
qdeslandes: `if (rule->log == BF_FLAG(BF_LOG_OPT_5_TUPLE)` would be enough here, no need for a variable that's only used once.

src/libbpfilter/cgen/packet.c
```diff
@@ -421,21 +421,40 @@ int bf_packet_gen_inline_matcher(struct bf_program *program,
...
+
+        // Pack l3_proto and l4_proto
+        EMIT(program, BPF_MOV64_REG(BPF_REG_4, BPF_REG_7));
+        EMIT(program, BPF_ALU64_IMM(BPF_LSH, BPF_REG_4, 16));
+        EMIT(program, BPF_ALU64_REG(BPF_OR, BPF_REG_4, BPF_REG_8));
 
-    EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_PKT_LOG);
+        EMIT_FIXUP_ELFSTUB(program, BF_ELFSTUB_PKT_5_TUPLE_LOG);
+    } else {
+        EMIT(program, BPF_MOV64_IMM(BPF_REG_3, headers));
```
qdeslandes: Passing `BF_LOG_PACKET_HEADERS` here instead of `BF_LOG_DEFAULT` doesn't bring any value here.

src/libbpfilter/include/bpfilter/runtime.h
```diff
@@ -90,12 +90,23 @@ enum bf_log_opt
...
+    BF_LOG_OPT_5_TUPLE,
+
     _BF_LOG_OPT_MAX,
 
     /** Log all available data for the hook type. */
     BF_LOG_OPT_DEFAULT = 0xFF,
 };
 
+/** Log all available packet headers. */
+#define BF_LOG_PACKET_HEADERS                                                  \
```
qdeslandes: Overall, this doesn't bring any value. The elfstub properly handle `BF_LOG_DEFAULT`, there's no need for a dedicated macro.

src/libbpfilter/include/bpfilter/runtime.h
qdeslandes: I don't think there is any value (as of now) to introduce `BF_LOG_PACKET_HEADERS`. We use `BF_LOG_OPT_DEFAULT` for now, which is valid for both packet and sock_addr programs. For sock_addr programs, it's only a flag (log or don't log) as we can't specify which data to log. For packet programs, `BF_LOG_OPT_DEFAULT` complicates the code: instead of dealing with `BF_LOG_OPT_DEFAULT` we need to deal with `BF_LOG_OPT_DEFAULT` AND `BF_LOCK_PACKET_HEADERS`.

qdeslandes (changes requested):

src/libbpfilter/cgen/packet.c
qdeslandes: Same as above, pass `rule->log == BF_LOG_OPT_DEFAULT ? BF_LOG_PACKET_HEADERS : rule->log` directly.

qdeslandes (changes requested):

src/libbpfilter/cgen/packet.c
blackdragoon26: Okay, I will inline it!

src/libbpfilter/cgen/packet.c
blackdragoon26: Alrighty

qdeslandes (approved):
LGTM, thanks for completing this :)

PR #582, opened 2026-09-14
Add `META_D/SPORT` set support

yaakov-stein:
## Summary

Previously `META_D/SPORT` did not support `BF_MATCHER_IN` (being used as a set key). Thus, if you wanted to have a set with either UDP/TCP ports, you'd need to make two separate rules. Here, we add the ability to use `META_D/SPORT` as a set key for both packet and sock hooks.

`_bf_matcher_generate_meta_port` was located in `meta.c`, even though it was packet specific (the `cgroup_sock_addr` path preempted it). Let's move that to the right location (`packet.c`), simplify it by using the fact the UDP/TCP header payload offsets are equal for source and destination port, and then add set support for `META_D/SPORT`.

## Testing

- Unit + e2e tests
- Manual tests - Played around using `META_D/SPORT` in sets as single keys and in combination with other set keys. Tested both packet and `CGROUP_SOCK_ADDR` path.

## AI disclosure

Worked with AI, mostly for coding.

## Checklist

- [x] `make -C $BUILD test_bin test` passes and the code follows the style guide


yaakov-stein:
@claude review

yaakov-stein:
@claude review

qdeslandes (changes requested):
Mostly good to me except for a few nits:
- Add tests to confirm `meta.sport` is rejected for sock_addr programs
- Add tests to validate `meta.(d|s)port` in sets for pkt programs (`e2e/parsing`)

qdeslandes (approved):
LGTM, thanks!

PR #586, opened 2026-09-16
lib: set: support port ranges in set elements

### LINKED ISSUE #562: Port range in sets

Jeevankumar-s:
## Summary

Add support for port ranges in single-component sets.

Port ranges such as 27-30 are parsed using the existing BF_MATCHER_RANGE matcher and expanded into individual set elements before being inserted into the set.

For example:

(tcp.dport) in { 11; 22; 27-30 }

is stored as the equivalent of:

{ 11; 22; 27; 28; 29; 30 }

Ranges in multi-component set keys are intentionally rejected.

## Related issue

Fixes #562

Fixes #

## Testing

Added unit tests for:

valid port ranges
boundary values
invalid/reversed ranges
multi-component range rejection

Ran:

make -C build test_bin test

Result: 121/122 tests passed.

All unit, integration, and e2e tests passed. The only failure is check.lint, caused by existing clang-tidy warnings in unchanged files. The modified set.c passes clang-tidy.

## Notes for the reviewer

Kept the change limited to single-component sets and reused the existing range parser instead of adding separate parsing logic.

## AI disclosure

I used AI to help understand the codebase and review the implementation. I reproduced the issue, reviewed the changes, and ran the tests locally.

## Checklist

- [ ] `make -C $BUILD test_bin test` passes and the code follows the style guide


src/libbpfilter/set.c
```diff
@@ -199,6 +203,55 @@ int bf_set_add_elem_raw(struct bf_set *set, const char *raw_elem)
...
+
+            if (set->n_comps != 1) {
+                return bf_err_r(
+                    -EINVAL,
+                    "ranges are not supported in multi-component set keys");
+            }
+
+            r = range_ops->parse(set->key[comp_idx], BF_MATCHER_RANGE, bounds,
+                                 token);
+
```
qdeslandes: No empty line here.

qdeslandes (changes requested):
Thanks for working on this!

Your implementation forces the specific behaviour of an edge-case (port range in set) over all the sets being processed. One example of this is `range_ops = bf_matcher_get_ops()` for each component of the value: we don't need to fetch the ops for every component as, by definition, a set we 2+ components won't be eligible for sets expansion.

Instead, I would suggest:
- Before the while loop over the value components, define `range_ops`, and call `bf_matcher_get_ops()` if: the key is a single component, this component is a port.
- Within the loop: if `range_ops` AND `strchr(token, '-')`, process the range. This will ensure we only run `strchr()` if we _might_ have a range, not for all cases (which is what is done here, because `strchr` runs before `range_ops` is checked.
- The logic used to expand the range should be moved to a static function.

Additionally:
- You need to update the documentation to reflect this change
- Your commit lacks a body describing its purpose
- You need to test edge cases: range at the beginning of the set, and in the middle. You only test when the range is the last element

tests/unit/libbpfilter/set.c
```diff
@@ -281,6 +281,60 @@ static void new_from_raw_multiple_keys(void **state)
...
+{
+    _free_bf_set_ struct bf_set *set = NULL;
+
+    (void)state;
+
+    assert_err(bf_set_new_from_raw(&set, "test", "(tcp.dport)", "{27-17}"));
+
+    assert_err(bf_set_new_from_raw(&set, "test", "(tcp.dport)", "{27-}"));
+
+    assert_err(bf_set_new_from_raw(&set, "test", "(tcp.dport)", "{-17}"));
```
qdeslandes, line 337: No need for newlines.

src/libbpfilter/set.c
```diff
@@ -199,6 +203,55 @@ int bf_set_add_elem_raw(struct bf_set *set, const char *raw_elem)
...
+                if (!range_elem) {
+                    return bf_err_r(-ENOMEM,
+                                    "failed to allocate a new set element");
+                }
+
+                (void)snprintf(value_str, sizeof(value_str), "%u", value);
+
+                r = ops->parse(set->key[comp_idx], BF_MATCHER_IN, range_elem,
+                               value_str);
+
```
qdeslandes: No empty line here.

qdeslandes (changes requested):

Jeevankumar-s:
> Thanks for working on this!
> 
> Your implementation forces the specific behaviour of an edge-case (port range in set) over all the sets being processed. One example of this is `range_ops = bf_matcher_get_ops()` for each component of the value: we don't need to fetch the ops for every component as, by definition, a set we 2+ components won't be eligible for sets expansion.
> 
> Instead, I would suggest:
> 
> * Before the while loop over the value components, define `range_ops`, and call `bf_matcher_get_ops()` if: the key is a single component, this component is a port.
> * Within the loop: if `range_ops` AND `strchr(token, '-')`, process the range. This will ensure we only run `strchr()` if we _might_ have a range, not for all cases (which is what is done here, because `strchr` runs before `range_ops` is checked.
> * The logic used to expand the range should be moved to a static function.
> 
> Additionally:
> 
> * You need to update the documentation to reflect this change
> * Your commit lacks a body describing its purpose
> * You need to test edge cases: range at the beginning of the set, and in the middle. You only test when the range is the last element

Thanks for the review! I've updated the implementation based on your feedback, moved the range handling out of the component loop, extracted the expansion logic into a helper, added the missing edge-case tests and documentation, and updated the commit message. I've also added a test for rejecting extra components after a range. Ready for another look when you have time.

qdeslandes:
@claude review

doc/usage/bfcli.rst
```diff
@@ -519,6 +519,27 @@ Here is an example:
...
+
+    set ports (tcp.dport) in { 11; 22; 27-30 }
+
+This is equivalent to:
+
+.. code:: shell
+
+    set ports (tcp.dport) in { 11; 22; 27; 28; 29; 30 }
+
+Each range is expanded into individual set elements. Duplicate ports, including those introduced by overlapping ranges, are stored only once. Ranges are supported in both named and anonymous sets.
```
qdeslandes: This line can be dropped.

src/libbpfilter/set.c
```diff
@@ -157,10 +160,53 @@ static int _bf_set_parse_key(const char *raw_key, enum bf_matcher_type *key,
...
+        if (r) {
+            return bf_err_r(r, "failed to parse expanded range value '%s'",
+                            value_str);
+        }
+
+        r = bf_hashset_add(&set->elems, &range_elem);
+        if (r == -EEXIST)
+            continue;
+        if (r)
+            return bf_err_r(r, "failed to insert element into set");
```
qdeslandes, line 183: What you do here:
1. Parse the range
2. Loop over the values in the range
3. Convert each value back to a string
4. Parse the value to insert it in the set

Don't you think some of those steps can be skipped? We should not convert the values to string to the parse them again.

qdeslandes (changes requested):
Few nits, otherwise LGTM :)

src/libbpfilter/set.c
```diff
@@ -180,6 +226,13 @@ int bf_set_add_elem_raw(struct bf_set *set, const char *raw_elem)
     if (!elem)
         return bf_err_r(-ENOMEM, "failed to allocate a new set element");
```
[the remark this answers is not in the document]
qdeslandes: `BF_MATCHER_META_SPORT` and `BF_MATCHER_META_DPORT` now support sets, you can allow them here.

qdeslandes (changes requested):

Jeevankumar-s:
Thanks for the review @qdeslandes! Updated the range expansion to insert parsed values directly without converting them back to strings. I also added meta.sport and meta.dport support after syncing with the latest main, and cleaned up the docs as suggested. Relevant unit and matcher tests are passing.

